PR #3, opened 2015-05-19
SONARCSANA-84 Parameter names differs in partial method declaration

tvsonar:


src/Tests/SonarQube.CSharp.CodeAnalysis.UnitTest/RuleFinderTest.cs
```diff
@@ -43,7 +43,7 @@ public void GetExtraRuleAssembly()
         [TestMethod]
         public void GetParameterlessAnalyzerTypes()
         {
-            new RuleFinder().GetParameterlessAnalyzerTypes().Should().HaveCount(46);
+            new RuleFinder().GetParameterlessAnalyzerTypes().Should().HaveCount(47);
```
dbolkensteyn: I'm not sure about the value of this test - WDYT?

src/Tests/SonarQube.CSharp.CodeAnalysis.RulingTest/Expected/S927.json
```diff
@@ -0,0 +1,13 @@
+{
+  "Version": "0.1",
+  "ToolInfo": {
+    "ToolName": "SonarQube.CSharp.CodeAnalysis.RulingTest",
+    "FileVersion": "1.0.0"
+  },
+  "Issues": [
+    {
+      "RuleId": "S927",
+      "Locations": []
```
dbolkensteyn: @tvsonar Shouldn't there be at least one issue in the ruling IT to test for non-regressions?

src/CSharp.CodeAnalysis/Rules.Description/S927.html
```diff
@@ -0,0 +1,46 @@
...
+{
+  partial void OnChange(int x, int y);
+}
+
+partial class Point
+{
+  int x = 0;
+  int y = 0;
+
+  partial void OnChange(int yk, int x)  // Noncompliant
```
dbolkensteyn: @tvsonar looks like there is a typo here `yk` instead of `y`

src/CSharp.CodeAnalysis/Rules/ParameterNamesInPartialMethod.cs
```diff
@@ -0,0 +1,87 @@
...
+                    var implementationParameters = methodSyntax.ParameterList.Parameters;
+                    var definitionParameters = methodSymbol.PartialDefinitionPart.Parameters;
+
+                    for (var i = 0; i < implementationParameters.Count; i++)
+                    {
+                        var implementationParameter = implementationParameters[i];
+                        if (definitionParameters.Length > i)
+                        {
+                            var definitionParameter = definitionParameters[i];
+                            if ((string) implementationParameter.Identifier.Value != definitionParameter.Name)
```
dbolkensteyn: You can use `ValueText` here instead of the cast to string

src/CSharp.CodeAnalysis/Rules/ParameterNamesInPartialMethod.cs
```diff
@@ -0,0 +1,87 @@
...
+                        return;
+                    }
+
+                    var implementationParameters = methodSyntax.ParameterList.Parameters;
+                    var definitionParameters = methodSymbol.PartialDefinitionPart.Parameters;
+
+                    for (var i = 0; i < implementationParameters.Count; i++)
+                    {
+                        var implementationParameter = implementationParameters[i];
+                        if (definitionParameters.Length > i)
```
dbolkensteyn: I'd rewrite into `if (i < definitionParamters.Length)`

src/CSharp.CodeAnalysis/Rules.Description/S927.html
tvsonar: @dbolkensteyn the typo is the reason for the noncompliance. But you are right, `x` and `yk` should be swapped, and also `yk` should be used in the body instead of `y`.

src/Tests/SonarQube.CSharp.CodeAnalysis.UnitTest/RuleFinderTest.cs
tvsonar: @dbolkensteyn I'll rewrite it to check for `Count > 0`

src/Tests/SonarQube.CSharp.CodeAnalysis.RulingTest/Expected/S927.json
tvsonar: @dbolkensteyn As discussed I'll leave it like this for now.

src/CSharp.CodeAnalysis/Rules/ParameterNamesInPartialMethod.cs
```diff
@@ -0,0 +1,87 @@
...
+                    for (var i = 0; i < implementationParameters.Count; i++)
+                    {
+                        var implementationParameter = implementationParameters[i];
+                        if (i < definitionParameters.Length)
+                        {
+                            var definitionParameter = definitionParameters[i];
+                            if (implementationParameter.Identifier.ValueText != definitionParameter.Name)
+                            {
+                                c.ReportDiagnostic(Diagnostic.Create(Rule, implementationParameter.Identifier.GetLocation(),
+                                    implementationParameter.Identifier.Value, definitionParameter.Name));
```
dbolkensteyn: @tvsonar you'll want to use `ValueText` here as well, so it might be better to extract into a variable

dbolkensteyn:
@tvsonar I've added one more comment - no need to reassign to me afterwards :+1:

PR #7, opened 2015-05-21
SONARCSANA-69 Fields that are only assigned in the constructor should…

tvsonar:
… be "readonly"

### DIFF
--- a/src/CSharp.CodeAnalysis/Rules/FieldShouldBeReadonly.cs
+++ b/src/CSharp.CodeAnalysis/Rules/FieldShouldBeReadonly.cs
@@ -1,0 +1,201 @@
+﻿/*
+ * SonarQube C# Code Analysis
+ * Copyright (C) 2015 SonarSource
+ * dev@sonar.codehaus.org
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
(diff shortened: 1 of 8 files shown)


src/CSharp.CodeAnalysis/Rules/FieldShouldBeReadonly.cs
```diff
@@ -0,0 +1,202 @@
...
+        private static void ProcessExpressionChange(ExpressionSyntax expression, SemanticModel semanticModel,
+            ref ImmutableHashSet<IFieldSymbol> nonCandidateFields, ref ImmutableHashSet<IFieldSymbol> assignedAsReadonly)
+        {
+            var fieldSymbol = semanticModel.GetSymbolInfo(expression).Symbol as IFieldSymbol;
+            if (!FieldIsRelevant(fieldSymbol))
+            {
+                return;
+            }
+
+            var anonymousExpression = expression.FirstAncestorOrSelf<AnonymousFunctionExpressionSyntax>();
```
tvsonar: It would probably be better to get the ```SemanticModel.GetEnclosingSymbol(position)``` and if that is not a constructor, then it is non-candidate.

src/CSharp.CodeAnalysis/Rules/FieldShouldBeReadonly.cs
```diff
@@ -0,0 +1,202 @@
...
+                        candidateFields = candidateFields.Add(field.Symbol);
+
+                        if (field.Syntax.Initializer != null)
+                        {
+                            assignedAsReadonly = assignedAsReadonly.Add(field.Symbol);
+                        }
+                    }
+                }, SyntaxKind.FieldDeclaration);
+
+                var nonCandidateFields = ImmutableHashSet<IFieldSymbol>.Empty;
```
dbolkensteyn: @tvsonar I'd move all state up to the beginning of the compilation start method

src/CSharp.CodeAnalysis/Rules/FieldShouldBeReadonly.cs
```diff
@@ -0,0 +1,202 @@
...
+                    },
+                    SyntaxKind.PostDecrementExpression,
+                    SyntaxKind.PostIncrementExpression);
+
+                analysisContext.RegisterSyntaxNodeAction(c =>
+                {
+                    var argument = (ArgumentSyntax) c.Node;
+
+                    if (argument.RefOrOutKeyword.IsKind(SyntaxKind.None) ||
+                        !(argument.Expression is IdentifierNameSyntax))
```
dbolkensteyn: @tvsonar What happens with something like `foo(ref this.MyField)` ?

src/CSharp.CodeAnalysis/Rules/FieldShouldBeReadonly.cs
dbolkensteyn: Agreed

PR #8, opened 2015-05-21
SONARCSANA-66 Properties that don't access instance data should be st…

tvsonar:
…atic


src/CSharp.CodeAnalysis/Rules.Description/S2325.html
```diff
@@ -18,7 +39,28 @@
 <pre>
 public class Utilities
 {
-  public static int Sum(int a, int b)  // Compliant
+  public int MagicNum
```
dbolkensteyn: @tvsonar isn't a `static` missing here?

src/CSharp.CodeAnalysis/Rules/MethodOrPropertyWithoutInstanceData.cs
```diff
@@ -0,0 +1,211 @@
...
+                   symbol.IsOverride ||
+                   symbol.IsAbstract;
+        }
+
+        private static bool IsInterfaceImplementation(ISymbol symbol)
+        {
+            var containingType = symbol.ContainingType;
+
+            var interfaces = new Queue<INamedTypeSymbol>(containingType.Interfaces);
+            var allInterfaces = new List<INamedTypeSymbol>();
```
dbolkensteyn: @tvsonar Why not simply `allInterfaces = containingType.AllInterfaces` ?

src/CSharp.CodeAnalysis/Rules/MethodOrPropertyWithoutInstanceData.cs
```diff
@@ -0,0 +1,211 @@
...
+                        if (methodOrPropertySymbol == null ||
+                            HasAllowedModifier(methodOrPropertySymbol) ||
+                            IsInterfaceImplementation(methodOrPropertySymbol))
+                        {
+                            return;
+                        }
+
+                        var inheritanceChain = GetInheritanceChain(methodOrPropertySymbol.ContainingType).ToImmutableHashSet();
+
+                        var reportShouldBeStatic = true;
```
dbolkensteyn: @tvsonar I'd move all state to the beginning of the code block start method

src/CSharp.CodeAnalysis/Rules/MethodOrPropertyWithoutInstanceData.cs
```diff
@@ -0,0 +1,211 @@
...
+                        cbc.RegisterCodeBlockEndAction(c =>
+                        {
+                            if (!reportShouldBeStatic)
+                            {
+                                return;
+                            }
+
+                            if (methodDeclaration != null)
+                            {
+                                methodsCouldBeStatic = methodsCouldBeStatic.Add(methodDeclaration);
```
dbolkensteyn: @tvsonar it seems that you could already create the issue at this point instead of buffering the issue till the end of the compilation?

src/CSharp.CodeAnalysis/Rules/MethodOrPropertyWithoutInstanceData.cs
```diff
@@ -0,0 +1,211 @@
...
+                    {
+                        c.ReportDiagnostic(Diagnostic.Create(Rule, methodDeclarationSyntax.Identifier.GetLocation(),
+                            methodDeclarationSyntax.Identifier.ValueText, "method"));
+                    }
+
+                    foreach (var property in propertyAccessorsByProperty)
+                    {
+                        var accessors = property.Value;
+
+                        if (accessors.All(accessor => accessorsCouldBeStatic.Contains(accessor)))
```
dbolkensteyn: @tvsonar aie - if I understand properly, the compilation state is required here only to check that both accessors do not use non-static fields?

tvsonar: @dbolkensteyn yes, that's right.

PR #9, opened 2015-05-21
Fixing tests to only use lower-case letters in the file paths

tvsonar:


dbolkensteyn:
@tvsonar As discussed - let's first check if this one is strictly required and understand why

PR #10, opened 2015-05-21
SONARCSANA-91 Improve the performance of rule S2326

tvsonar:


src/CSharp.CodeAnalysis/Helpers/SemanticsHelper.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace SonarQube.CSharp.CodeAnalysis.Helpers
+{
+    internal static class SemanticsHelper
+    {
+        internal static bool IsInterfaceImplementation(this IMethodSymbol methodSymbol)
+        {
+            var containingType = methodSymbol.ContainingType;
+
+            var interfaces = new Queue<INamedTypeSymbol>(containingType.Interfaces);
```
dbolkensteyn: @tvsonar This could use simply `AllInterfaces` right?

src/CSharp.CodeAnalysis/Helpers/SemanticsHelper.cs
```diff
@@ -0,0 +1,48 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+
+namespace SonarQube.CSharp.CodeAnalysis.Helpers
+{
+    internal static class SemanticsHelper
```
dbolkensteyn: @tvsonar Could you schedule a 10 minutes discussion regarding extension methods?

dbolkensteyn:
@tvsonar How does this change affect the performances of the rule?

tvsonar:
@dbolkensteyn That, we don't know yet, but the analyzer should be implemented this way. It was suggested by the Roslyn team.

dbolkensteyn:
@tvsonar why is this pull request still open?

tvsonar:
@dbolkensteyn Because it is still a valid refactoring. We have pushed the proper performance testing to a later sprint, so we'll come back to this PR later.

dbolkensteyn:
ok - in that case I suggest to close the PR here, keep the branch on your fork, and resend the PR when time will be appropriate

PR #12, opened 2015-05-26
SONARCSANA-105 Locks should be released

tvsonar:


dbolkensteyn:
This rule is badly specified as is and cannot yet be implemented.

PR #13, opened 2015-05-26
SONARCSANA-113 Any() should be used to test for emptiness

tvsonar:


src/Tests/CSharp.CodeAnalysis.UnitTest/Verifier.cs
```diff
@@ -80,6 +80,7 @@ public static void Verify(string path, DiagnosticAnalyzer diagnosticAnalyzer)
             {
                 var document = workspace.CurrentSolution.AddProject("foo", "foo.dll", LanguageNames.CSharp)
                     .AddMetadataReference(MetadataReference.CreateFromAssembly(typeof(object).Assembly))
+                    .AddMetadataReference(MetadataReference.CreateFromAssembly(typeof(Enumerable).Assembly))
```
dbolkensteyn: @tvsonar Could you briefly explain why this change is required? Will this rule work with the current rule runner?

tvsonar: @dbolkensteyn It was needed because the semantic analyzer can only find the extension methods if it has a reference to the assembly.
The rule runner will not fail. But we should probably add this reference to the ```CompilationHelper``` as well.

dbolkensteyn:
@tvsonar LGTM - I let you decide if any change should be made to the rule runner

PR #15, opened 2015-05-28
SONARCSANA-81 Property assignments should not be made for "readonly" …

tvsonar:
…fields not constrained to reference types

### DIFF
--- a/src/CSharp.CodeAnalysis/Rules.Description/S2934.html
+++ b/src/CSharp.CodeAnalysis/Rules.Description/S2934.html
@@ -1,0 +1,78 @@
+<p>
+    While the properties of a <code>readonly</code> reference type field can still be changed after initialization,
+    those of a <code>readonly</code> value field, such as a <code>struct</code>, cannot.
+</p>
+<p>
+    If the member could be either a <code>class</code> or a <code>struct</code> then assignment to its properties 
+    could be unreliable, working sometimes but not others.
+</p>
+
+<h2>Noncompliant Code Example</h2>
+<pre>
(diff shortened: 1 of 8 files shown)


src/CSharp.CodeAnalysis/Rules.Description/S2934.html
```diff
@@ -0,0 +1,78 @@
+<p>
+    While the properties of a <code>readonly</code> reference type field can still be changed after initialization,
+    those of a <code>readonly</code> value field, such as a <code>struct</code>, cannot.
+</p>
+<p>
+    If the member could be either a <code>code</code> or a <code>struct</code> then assignment to its properties
```
dbolkensteyn: @tvsonar a `code`?

src/CSharp.CodeAnalysis/Rules/PropertiesOfReadonlyGenericsField.cs
```diff
@@ -0,0 +1,137 @@
...
+        private static void ProcessPropertyChange(ExpressionSyntax expression, SemanticModel semanticModel, 
+            SyntaxNodeAnalysisContext context)
+        {
+            var memberAccess = expression as MemberAccessExpressionSyntax;
+            if (memberAccess == null)
+            {
+                return;
+            }
+
+            var propertySymbol = semanticModel.GetSymbolInfo(expression).Symbol as IPropertySymbol;
```
dbolkensteyn: @tvsonar does this also work with `foo.MyField` ?

PR #16, opened 2015-05-28
SONARCSANA-74 Improve reported error locations to minimize underline …

tvsonar:
…length


dbolkensteyn:
@tvsonar OK from a technical point of view - Probably Ann should review it functionnally

PR #18, opened 2015-05-29
Public field

tvsonar:


src/CSharp.CodeAnalysis/Rules/VisibleStaticField.cs
```diff
@@ -0,0 +1,89 @@
...
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.SynchronizationReliability)]
+    [Rule(DiagnosticId, RuleSeverity, Description, IsActivatedByDefault)]
+    [Tags("pitfall")]
+    public class VisibleStaticField : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S2223";
+        internal const string Description = "Non-constant static fields should not be visible";
+        internal const string MessageFormat = "Change the visibility of \"{0}\" or make it \"static\" or \"readonly\".";
+        internal const string Category = "SonarQube";
+        internal const Severity RuleSeverity = Severity.Critical;
```
dbolkensteyn: @tvsonar You think this really is critical?

dbolkensteyn:
@tvsonar The commit history doesn't look good. There should be like 1 PR per rule, and I expect to see a SONARCSANA ticket associated to each commit

src/CSharp.CodeAnalysis/Rules/VisibleInstanceField.cs
```diff
@@ -0,0 +1,88 @@
...
+                    {
+                        c.ReportDiagnostic(Diagnostic.Create(Rule, field.Syntax.Identifier.GetLocation(),
+                            field.Syntax.Identifier.ValueText));
+                    }
+
+                },
+                SyntaxKind.FieldDeclaration);
+        }
+
+        private static bool FieldIsRelevant(IFieldSymbol fieldSymbol)
```
dbolkensteyn: @tvsonar Does this rule check if there is a public field in an internally visible class for example? (if this scenario is possible)

PR #20, opened 2015-06-03
SONARCSANA-70 Display plain text error descriptions in the error window

tvsonar:


src/CSharp.CodeAnalysis.Extra/Rules/ClassName.cs
```diff
@@ -46,8 +46,7 @@ public class ClassName : DiagnosticAnalyzer
 
         internal static DiagnosticDescriptor Rule =
             new DiagnosticDescriptor(DiagnosticId, Description, MessageFormat, Category,
-                RuleSeverity.ToDiagnosticSeverity(), IsActivatedByDefault,
-                helpLinkUri: "http://nemo.sonarqube.org/coding_rules#rule_key=csharpsquid%3AS101");
+                RuleSeverity.ToDiagnosticSeverity(), IsActivatedByDefault);
```
dbolkensteyn: @tvsonar this change seems actually more related to SONARCSANA-21

dbolkensteyn:
@tvsonar I'm not sure this PR is a good idea, it's yet-another-place where a rule description needs to be kept in sync (1. RSPEC, 2. HTML description and now 3. String description within VS)

You were the first to complain during the retrospective that managing 1 & 2 are already painful :)

But I'm fine with this as long as I'm not maintaining it.. ;)

tvsonar:
@dbolkensteyn yes, I know, but I like the additional info in VS. And I couldn't come up with a generic solution to include the description automatically.

PR #24, opened 2015-06-09
SONARCSANA-114 Types and "this" should not be used for locking

tvsonar:


src/CSharp.CodeAnalysis/Rules.Description/S2551.html
```diff
@@ -0,0 +1,31 @@
+<p>
+    Locking on the current object instance (i.e. <code>this</code>), or on a <code>Type</code> object is a bad idea because 
+    any other thread could acquire (or attempt to acquire) the same lock for another unrelated purpose, thus creating deadlocks.
```
dbolkensteyn: @tvsonar please explain me how this creates a deadlock

dbolkensteyn: note to self: this rule is CRITICAL and activated by default

src/CSharp.CodeAnalysis/Rules/LockOnThisOrType.cs
```diff
@@ -0,0 +1,88 @@
...
+        internal const Severity RuleSeverity = Severity.Critical;
+        internal const bool IsActivatedByDefault = true;
+
+        internal static readonly DiagnosticDescriptor Rule = 
+            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, 
+                RuleSeverity.ToDiagnosticSeverity(), IsActivatedByDefault, 
+                helpLinkUri: DiagnosticId.GetHelpLink(),
+                description: Description);
+
+        private const string SystemTypeFullName = "System.Type";
```
dbolkensteyn: @tvsonar Does this make any sense? The Gendarme rule descriptions mentions *a* Type, so any given type: http://www.mono-project.com/docs/tools+libraries/tools/gendarme/rules/concurrency/#donotlockonthisortypesrule

src/CSharp.CodeAnalysis/Rules.Description/S2551.html
```diff
@@ -0,0 +1,31 @@
...
+</p>
+<p>
+    Instead, a new, private object should be created and used for the lock.
+</p>
+
+<h2>Noncompliant Code Example</h2>
+<pre>
+public void MyLockingMethod()
+{
+  lock (this) // Noncompliant
```
dbolkensteyn: @tvsonar I don't see the deadlock in this example

src/CSharp.CodeAnalysis/Rules.Description/S2551.html
tvsonar: @dbolkensteyn The problem here is that to acquire a lock you need to specify an object that you are locking on. It is not a good idea to lock on something that is reachable by many consumers, because they also might lock on it for a different reason, which might result in a deadlock.
The `this` reference and any type are just two examples that can be reached from many places, and thus should be avoided for locking.

src/CSharp.CodeAnalysis/Rules.Description/S2551.html
tvsonar: @dbolkensteyn because there is none here. It just shows what could cause a deadlock. Do you think we should add a full example to RSPEC?

src/CSharp.CodeAnalysis/Rules/LockOnThisOrType.cs
tvsonar: @dbolkensteyn yes, it makes sense. Later, where we check what we are locking on, we are getting the type of the symbol, and check if the type is of `System.Type`.
So when we analyze the following:
```
lock(someExpression)
{
  //...
}
```
We get the type of `someExpression`. If it is `System.Type` then we have an issue. If it is let's say `int`, then we are okay.

dbolkensteyn: OK, that makes sense indeed

src/CSharp.CodeAnalysis/Rules.Description/S2551.html
dbolkensteyn: @tvsonar Yep - but I don't think the RSPEC is clear enough on that. IMO ideally we'll want to show an instance of a real issue

my understanding of the current description is that, as soon as another thread will also lock "this" or the same type, then you will have a deadlock

I'm fine with keeping the severity at CRITICAL level

dbolkensteyn:
LGTM

PR #26, opened 2015-06-11
SONARCSANA-131 Object.ReferenceEquals called with value types

tvsonar:


src/CSharp.CodeAnalysis/Rules/ReferenceEqualsOnValueType.cs
```diff
@@ -0,0 +1,93 @@
...
+                    {
+                        c.ReportDiagnostic(Diagnostic.Create(Rule, invocation.Expression.GetLocation()));
+                    }
+                },
+                SyntaxKind.InvocationExpression);
+        }
+
+        private static bool AnyArgumentIsValueType(InvocationExpressionSyntax invocation, SemanticModel semanticModel)
+        {
+            if (invocation.ArgumentList == null)
```
dbolkensteyn: @tvsonar Is it possible to have null here? Using "null" for empty lists is a well known anti-pattern (we have a rule for it in Java, which BTW we probably also want for C#), and I'd be surprised if Roslyn does it.

If not, it's probably better to pass the ArgumentList directly to this function.

src/CSharp.CodeAnalysis/Rules/ReferenceEqualsOnValueType.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                return false;
+            }
+
+            return invocation.ArgumentList.Arguments.Any(argument =>
```
dbolkensteyn: @tvsonar So you've decided to not make an exception for: `Foo(ref a, ref b)`, `Object.ReferenceEquals(a, b)` - that works for me, as long as it was a deliberate choice :)

dbolkensteyn:
LGTM

src/CSharp.CodeAnalysis/Rules/ReferenceEqualsOnValueType.cs
tvsonar: @dbolkensteyn You guessed it correctly, it can't be null. (But it is actually not a list in this case). C# is already targeted in the mentioned RSPEC (http://jira.sonarsource.com/browse/RSPEC-1168).

PR #27, opened 2015-06-11
SONARCSANA-141 "Unused local variables" rule should not report on "us…

tvsonar:
…ing" statements


src/Tests/CSharp.CodeAnalysis.UnitTest/TestCases/UnusedLocalVariable.cs
```diff
@@ -14,7 +14,7 @@ void F1()
 
             var localRepository = new Cl { packageA, packageB }; // Noncompliant
 
-            using (var x = new StreamReader("")) // Noncompliant
+            using (var x = new StreamReader("")) // Compliant - Microsoft asked for it
```
dbolkensteyn: @tvsonar Let's move that comment to the RSPEC C# subtask

dbolkensteyn:
LGTM

PR #28, opened 2015-06-11
SONARCSANA-127 ThreadStaticAttibute doesn't work on non-static fields

tvsonar:


src/CSharp.CodeAnalysis/Rules/ThreadStaticNonStaticField.cs
```diff
@@ -0,0 +1,104 @@
...
+    public class ThreadStaticNonStaticField : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S3005";
+        internal const string Title = "\"ThreadStatic\" should not be used on non-static fields";
+        internal const string Description =
+            "The \"ThreadStatic\" attribute has no effect on non-static class fields. However, annotating such a " +
+            "field with \"ThreadStatic\" may confuse maintainers about the true behavior of the field.";
+        internal const string MessageFormat = "Remove the \"ThreadStatic\" attribute from this definition.";
+        internal const string Category = "SonarQube";
+        internal const Severity RuleSeverity = Severity.Major;
```
dbolkensteyn: @tvsonar Don't you want to make it critical?

dbolkensteyn: @tvsonar In fact I think this rule is not really about understandability, but about potential bugs: The developer most likely thought that each thread would get its own value, but that is just not what is happening.

I doubt that someone would purposely leave the `ThreadStatic` in, knowing that it has no useful effect

dbolkensteyn:
LGTM

PR #29, opened 2015-06-11
SONARCSANA-128 Thread static field has initializer

tvsonar:


src/CSharp.CodeAnalysis/Rules.Description/S2996.html
```diff
@@ -0,0 +1,37 @@
+<p>
+    When an object has a field annotated with <code>ThreadStatic</code>, that field is shared within a given thread,
+    but unique across threads. That's a good thing, but since a class' static initializer is only invoked for the
```
dbolkensteyn: @tvsonar "That's a good thing" doesn't very technical nor accurate nor relevant :)

dbolkensteyn:
LGTM

PR #31, opened 2015-06-12
SONARCSANA-130 IDisposable local variables should not be returned fro…

tvsonar:
…m a "using" statement

### DIFF
--- a/src/CSharp.CodeAnalysis/Rules/DisposableReturnedFromUsing.cs
+++ b/src/CSharp.CodeAnalysis/Rules/DisposableReturnedFromUsing.cs
@@ -1,0 +1,129 @@
+﻿/*
+ * SonarQube C# Code Analysis
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarQube.CSharp.CodeAnalysis.Common;
+using SonarQube.CSharp.CodeAnalysis.Common.Sqale;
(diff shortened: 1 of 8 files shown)


src/CSharp.CodeAnalysis/Rules/DisposableReturnedFromUsing.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterSyntaxNodeAction(
+                c =>
+                {
+                    var usingStatement = (UsingStatementSyntax) c.Node;
+                    var declaration = usingStatement.Declaration;
+                    var declaredSymbols = ImmutableArray<ISymbol>.Empty;
+                    if (declaration != null)
```
dbolkensteyn: @tvsonar Does Roslyn use null if the node is not present?

src/CSharp.CodeAnalysis/Rules/DisposableReturnedFromUsing.cs
```diff
@@ -0,0 +1,134 @@
...
+                    {
+                        c.ReportDiagnostic(Diagnostic.Create(Rule, usingStatement.UsingKeyword.GetLocation(),
+                            string.Join(", ",
+                                returnedSymbols.Select(s => string.Format("\"{0}\"", s.Name)))));
+                    }
+                },
+                SyntaxKind.UsingStatement);
+        }
+
+        private static ImmutableArray<ISymbol> GetReturnedSymbolSubset(StatementSyntax usingStatement,
```
dbolkensteyn: @tvsonar This seems quite complex. Why can't you simply return all returned symbols by `return` statements within the `using` block, and then check for the non-empty intersection with `declaredSymbols` ?

(so this function would just take the using statement and the sematic model as parameters)

src/CSharp.CodeAnalysis/Rules/DisposableReturnedFromUsing.cs
tvsonar: @dbolkensteyn Yes, either the `Declaration` or the `Expression` is `null`.

src/CSharp.CodeAnalysis/Rules/DisposableReturnedFromUsing.cs
```diff
@@ -0,0 +1,129 @@
...
+                                returnedSymbols.Select(s => string.Format("\"{0}\"", s.Name)))));
+                    }
+                },
+                SyntaxKind.UsingStatement);
+        }
+
+        private static ImmutableHashSet<ISymbol> GetReturnedSymbols(StatementSyntax usingStatement, 
+            SemanticModel semanticModel)
+        {
+            var enclosingSymbol = semanticModel.GetEnclosingSymbol(usingStatement.SpanStart);
```
dbolkensteyn: @tvsonar If I remove this line (and one more hereunder), all tests are still green - so I think they should be removed

src/CSharp.CodeAnalysis/Rules/DisposableReturnedFromUsing.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            return usingStatement.DescendantNodesAndSelf()
+                .OfType<ReturnStatementSyntax>()
+                .Where(ret => semanticModel.GetEnclosingSymbol(ret.SpanStart).Equals(enclosingSymbol))
```
dbolkensteyn: @tvsonar and this one

dbolkensteyn:
LGTM except for the 2 useless lines

tvsonar:
@dbolkensteyn I added an additional test case that requires these 2 lines

PR #36, opened 2015-06-17
SONARCSANA-120 Redundant casts should not be used

tvsonar:


src/Tests/SonarQube.CSharp.CodeAnalysis.RulingTest/Expected/S1905.json
```diff
@@ -0,0 +1,2831 @@
+{
```
dbolkensteyn: @tvsonar How comes that there are so many useless casts in it-sources?

src/CSharp.CodeAnalysis/Rules/RedundantCast.cs
```diff
@@ -0,0 +1,189 @@
...
+                return collectionType.TypeArguments.First();
+            }
+
+            var arrayType = semanticModel.GetTypeInfo(collection).Type as IArrayTypeSymbol;
+            return arrayType != null 
+                ? arrayType.ElementType 
+                : null;
+        }
+
+        internal static bool MethodIsOnIEnumerable(IMethodSymbol methodSymbol, SemanticModel semanticModel)
```
dbolkensteyn: @tvsonar I guess this is the method that you wanted to factorize

dbolkensteyn:
LGTM

src/Tests/SonarQube.CSharp.CodeAnalysis.RulingTest/Expected/S1905.json
tvsonar: @dbolkensteyn mainly because in "SharpCompress\Compressor\BZip2\CRC.cs" we have a huge array of {{int}}s constructed from hexadecimal values, such as {{(int)0x130476dc}}. It was most probably ported from http://www.opensource.apple.com/source/cctools/cctools-667.8.0/libstuff/crc32.c, where the values seem to be unsigned.

PR #37, opened 2015-06-18
SONARCSANA-110 Classes should "Dispose" of members from the classes' …

tvsonar:
…own "Dispose" methods

### DIFF
--- a/src/CSharp.CodeAnalysis/Rules/DisposeFromDispose.cs
+++ b/src/CSharp.CodeAnalysis/Rules/DisposeFromDispose.cs
@@ -1,0 +1,114 @@
+﻿/*
+ * SonarQube C# Code Analysis
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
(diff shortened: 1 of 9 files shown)


src/CSharp.CodeAnalysis/Rules/DisposeFromDispose.cs
```diff
@@ -0,0 +1,114 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Rule); } }
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterSyntaxNodeAction(
+                c =>
+                {
+                    var invocation = (InvocationExpressionSyntax) c.Node;
+                    var memberAccess = invocation.Expression as MemberAccessExpressionSyntax;
+                    if (memberAccess == null)
```
dbolkensteyn: @tvsonar I think that you're forcing the field name to be prefixed by `this` here

dbolkensteyn:
LGTM

PR #38, opened 2015-06-19
SONARCSANA-111 Methods named Dispose should implement IDisposable.Dis…

tvsonar:
…pose

### DIFF
--- a/src/CSharp.CodeAnalysis/Rules/DisposeNotImplementingDispose.cs
+++ b/src/CSharp.CodeAnalysis/Rules/DisposeNotImplementingDispose.cs
@@ -1,0 +1,220 @@
+﻿/*
+ * SonarQube C# Code Analysis
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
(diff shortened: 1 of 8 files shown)


src/CSharp.CodeAnalysis/Rules/DisposeNotImplementingDispose.cs
```diff
@@ -0,0 +1,205 @@
...
+                    var disposeMethod =
+                        (IMethodSymbol) analysisContext.Compilation.GetSpecialType(SpecialType.System_IDisposable)
+                            .GetMembers(DisposeMethodName)
+                            .Single();
+                    
+                    var disposeMethodsCalledFromDispose = ImmutableDictionary<INamedTypeSymbol, ImmutableHashSet<IMethodSymbol>>.Empty;
+                    var implementingDisposeMethods = ImmutableHashSet<IMethodSymbol>.Empty;
+                    var allDisposeMethods = ImmutableHashSet<IMethodSymbol>.Empty;
+
+                    analysisContext.RegisterSymbolAction(
```
dbolkensteyn: @tvsonar Each callback defined here has its own purpose. However at the moment, this purpose has to be "reverse-engineered" from the implementation.

I'm thinking that, in the future, we'll want to keep lambdas small, and might want to extract those callbacks into dedicated methods. Those methods can then be named in such a way that their purpose is explicit.

Then, the big picture of this giant `RegisterCompilationStartAction` can become clear, as it will be just a combination of 3 well defined steps.

PR #39, opened 2015-06-22
SONARCSANA-162, and Ruling test move

tvsonar:


dbolkensteyn:
LGTM, except on a few points:
 * I don't really see why the project level .gitignore needs to be changed - this probably is a mistake. Worse case a sub-gitignore file should be used
 * The internal release script can probably dropped

PR #40, opened 2015-06-23
SONARCSANA-153 Constructors should only call non-overridable methods

tvsonar:
After implementing this rule, I realized that it also exists in FxCop. Its ID is CA2214 (http://source.roslyn.codeplex.com/#Microsoft.AnalyzerPowerPack.CSharp/Usage/CSharpCA2214DiagnosticAnalyzer.cs), and is already implemented by the Roslyn team. (It is not a general compiler error, and it is also implemented by Resharper)


dbolkensteyn:
LGTM

PR #41, opened 2015-06-23
SONARCSANA-151, SONARCSANA-161 Silly bitwise operations and refactoring for constants

tvsonar:


src/CSharp.CodeAnalysis/Rules.Description/S2437.html
```diff
@@ -0,0 +1,7 @@
+<p>
+    Certain bit operations are just silly and should not be performed because their results are predictable.
+</p>
+<p>
+    Specifically, using <code>& -1</code> with any value will always result in the original value, as will
```
dbolkensteyn: html escaping

src/CSharp.CodeAnalysis/Rules/SillyBitwiseOperation.cs
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        private static bool TryConvert(object o, out int value)
+        {
+            try
+            {
+                value = Convert.ToInt32(o);
+                return true;
+            }
+            catch
```
dbolkensteyn: @tvsonar you're going to want to catch a specific exception instead of all

src/CSharp.CodeAnalysis/Rules/SillyBitwiseOperation.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        internal static bool TryGetConstantIntValue(ExpressionSyntax expression, out int value)
+        {
+            var multiplier = 1;
+            var expr = expression;
+            var unary = expr as PrefixUnaryExpressionSyntax;
+            while (unary != null)
+            {
+                var op = unary.OperatorToken;
+                if (op.IsKind(SyntaxKind.MinusToken))
```
dbolkensteyn: @tvsonar i'd use a whitelist here, to allow just for minus and plus. i'm not sure how other unary operators such as `!` or `~` will behave

dbolkensteyn:
LGTM

PR #42, opened 2015-06-24
SONARCSANA-136 Instance methods should not write to "static" fields

tvsonar:


src/CSharp.CodeAnalysis/Rules/StaticFieldWrittenFromInstanceMember.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        private static bool IsStaticFieldModification(ExpressionSyntax expression, ISymbol declaration, SemanticModel semanticModel)
+        {
+            var fieldSymbol = semanticModel.GetSymbolInfo(expression).Symbol as IFieldSymbol;
+            if (!FieldIsRelevant(fieldSymbol))
+            {
+                return false;
+            }
+
+            var enclosingSymbol = semanticModel.GetEnclosingSymbol(expression.SpanStart);
```
dbolkensteyn: @tvsonar is this check wanted?

dbolkensteyn: i could remove it with all tests passing

tvsonar: @dbolkensteyn I extended the test cases

dbolkensteyn: @tvsonar I think the exception should be removed instead of covered, it is better from a functional point of view IMO:

```
        public void DoSomething()
        {
            count++;  // Noncompliant
            var action = new Action(() =>
            {
                count++; // Noncompliant
            });
            countInstance++;
        }

        public static void DoSomethingStatic()
        {
            count++;
            var action = new Action(() =>
            {
                count++;
            });
        }
```

This is how the rule behaves without this exception. I don't see a good reason to exclude the lambda modification within the non-static method.

The static method does not register any syntax node action, so it does anyway not trigger on the lambda.

PR #45, opened 2015-06-25
SONARCSANA-154 An abstract class should have both abstract and concre…

tvsonar:
…te methods

### DIFF
--- a/src/CSharp.CodeAnalysis/Rules/ClassShouldNotBeAbstract.cs
+++ b/src/CSharp.CodeAnalysis/Rules/ClassShouldNotBeAbstract.cs
@@ -1,0 +1,132 @@
+﻿/*
+ * SonarQube C# Code Analysis
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


src/CSharp.CodeAnalysis/Rules/ClassShouldNotBeAbstract.cs
```diff
@@ -0,0 +1,140 @@
...
+            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, 
+                RuleSeverity.ToDiagnosticSeverity(), IsActivatedByDefault, 
+                helpLinkUri: DiagnosticId.GetHelpLink(),
+                description: Description);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Rule); } }
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterCompilationStartAction(
```
dbolkensteyn: @tvsonar i could remove the `RegisterCompilationStartAction` without breaking any test

tvsonar: @dbolkensteyn Yes you can remove it. The only difference is in case of `partial classes`. This way we only call the `GetMembers()` on each class symbol once. If we operate on the class declarations, we will call it multiple times for partial classes.
On the other hand, this way we need to get the `DeclaringSyntaxReferences` and keep a list of symbols.

I'm not sure which is more performant, it's probably just a matter of taste now. Should I simplify it just use a `RegisterSyntaxNodeAction`? WDYT?

tvsonar: @dbolkensteyn I've added your version to a second commit in this PR. It simplifies the rule, so I'd probably go with this one.

dbolkensteyn:
LGTM

PR #46, opened 2015-06-25
SONARCSANA-152 Child class members should not shadow parent class mem…

tvsonar:
…bers

### DIFF
--- a/src/CSharp.CodeAnalysis/Rules/FieldShadowsParentField.cs
+++ b/src/CSharp.CodeAnalysis/Rules/FieldShadowsParentField.cs
@@ -1,0 +1,114 @@
+﻿/*
+ * SonarQube C# Code Analysis
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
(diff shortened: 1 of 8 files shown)


src/CSharp.CodeAnalysis/Rules/FieldShadowsParentField.cs
```diff
@@ -0,0 +1,115 @@
...
+            {
+                return;
+            }
+
+            var fieldName = fieldSymbol.Name;
+            var fieldNameLower = fieldSymbol.Name.ToLower();
+            var declaringType = fieldSymbol.ContainingType;
+            var baseType = declaringType.BaseType;
+            while (baseType != null &&
+                   (baseType.SpecialType != SpecialType.System_Object ||
```
dbolkensteyn: @tvsonar the logic looks strange to me here (the `||`) - i think a `&&` was intented. But in any case, you probably can just remove the check of `System.Object` ? I don't expect that one to declare lots of members - and it's misleading as well to re-use these

dbolkensteyn:
LGTM

src/CSharp.CodeAnalysis/Rules/FieldShadowsParentField.cs
tvsonar: @dbolkensteyn you're right. I've removed the check for `System.Object`

PR #47, opened 2015-06-26
SONARCSANA-155 Modulus results should not be checked for direct equality

tvsonar:


src/CSharp.CodeAnalysis/Rules/EqualityCheckOnModulus.cs
```diff
@@ -0,0 +1,142 @@
...
+            {
+                currentExpression = parenthesized.Expression;
+                parenthesized = currentExpression as ParenthesizedExpressionSyntax;
+            }
+
+            var binary = currentExpression as BinaryExpressionSyntax;
+            return binary != null && binary.IsKind(SyntaxKind.ModuloExpression);
+        }
+
+        private static bool TryConvert(object o, out int value)
```
dbolkensteyn: @tvsonar this part also duplicates silly operator rule and the IndexOf > 0 rule, but has a slightly worse implementation -  so they should be put in common

dbolkensteyn:
LGTM except for the duplication which should be fixed

PR #48, opened 2015-06-26
SONARCSANA-149 "IndexOf" checks should not be for positive numbers

tvsonar:


src/CSharp.CodeAnalysis/Rules/IndexOfCheckAgainstZero.cs
```diff
@@ -0,0 +1,198 @@
...
+                if (possibleTypes.Contains(baseType))
+                {
+                    return true;
+                }
+                baseType = baseType.BaseType;
+            }
+            return false;
+        }
+
+        //todo merge this with SillyBitwiseOperation
```
dbolkensteyn: @tvsonar notice the remaining todo here

src/CSharp.CodeAnalysis/Rules/IndexOfCheckAgainstZero.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        //todo merge this with SillyBitwiseOperation 
+        private static bool TryConvert(object o, out int value)
+        {
+            try
+            {
+                value = Convert.ToInt32(o);
+                return true;
+            }
+            catch (Exception exception)
```
dbolkensteyn: @tvsonar this is exception as control flow - i don't know a better solution myself, but there must be one :)

dbolkensteyn:
LGTM

PR #49, opened 2015-06-29
SONARCSANA-164 Rule S124 should have parameters in the generates rule…

tvsonar:
…s.xml


src/Tests/SonarQube.CSharp.CodeAnalysis.UnitTest/ConfigurationTest.cs
```diff
@@ -66,13 +66,14 @@ public void Configuration()
...
 
             var commentAnalyzer = analyzers.OfType<CommentRegularExpression>().Single();
-            commentAnalyzer.Rules.Should().HaveCount(2);
-            commentAnalyzer.Rules[0].Descriptor.Id.ShouldBeEquivalentTo("TODO");
-            commentAnalyzer.Rules[0].Descriptor.MessageFormat.ToString().ShouldBeEquivalentTo("Fix this TODO");
-            commentAnalyzer.Rules[0].RegularExpression.ShouldBeEquivalentTo(".*TODO.*");
-            commentAnalyzer.Rules[1].Descriptor.Id.ShouldBeEquivalentTo("FIXME");
-            commentAnalyzer.Rules[1].Descriptor.MessageFormat.ToString().ShouldBeEquivalentTo("Fix this FIXME");
-            commentAnalyzer.Rules[1].RegularExpression.ShouldBeEquivalentTo(".*FIXME.*");
+            var ruleInstances = commentAnalyzer.RuleInstances.OfType<CommentRegularExpressionRule>().ToList();
```
tvsonar: This cast is unneeded.

src/CSharp.CodeAnalysis.Common/SonarQube.Settings/IRuleTemplateInstance.cs
```diff
@@ -0,0 +1,29 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+    using Microsoft.CodeAnalysis;
```
dbolkensteyn: @tvsonar wrong indentation

src/CSharp.CodeAnalysis.Common/RuleDetailBuilder.cs
```diff
@@ -60,44 +60,33 @@ private static RuleDetail GetRuleDetail(Type analyzerType)
                 IsTemplate = rule.Template
```
dbolkensteyn: @tvsonar minor: at the beginning of this method, you should be able to replace the `FirstOrDefault()` with a call to `Single()`, remove the null check and the return null

src/CSharp.CodeAnalysis.Common/RuleDetailBuilder.cs
```diff
@@ -60,44 +60,33 @@ private static RuleDetail GetRuleDetail(Type analyzerType)
...
-                    });
-            }
-
-            var tags = analyzerType.GetCustomAttributes<TagsAttribute>().FirstOrDefault();
-
-            if (tags != null)
-            {
-                ruleDetail.Tags.AddRange(tags.Tags);
-            }
+            GetParameters(analyzerType, ruleDetail, rule.Template);
```
dbolkensteyn: @tvsonar why is a `Template` property present in the rule annotation, when you now actually require an interface to be implemented for rule templates?

src/CSharp.CodeAnalysis.Common/RuleDetailBuilder.cs
```diff
@@ -116,7 +105,44 @@ private static RuleDetail GetRuleDetail(Type analyzerType)
...
+
+        private static void GetParameters(Type analyzerType, RuleDetail ruleDetail, bool isTemplate)
+        {
+            var typeToGetParametersFrom = analyzerType;
+            if (isTemplate)
+            {
+                var templateInterface = analyzerType.GetInterfaces()
+                    .First(type => type.IsGenericType &&
+                                   type.GetGenericTypeDefinition() == typeof (IRuleTemplate<>));
+
```
dbolkensteyn: @tvsonar I would add a check at this stage to verify that no rule parameter is defined on `analyzerType` (which is an easy mistake to make)

src/CSharp.CodeAnalysis.Common/RuleDetailBuilder.cs
```diff
@@ -116,7 +105,44 @@ private static RuleDetail GetRuleDetail(Type analyzerType)
...
+                typeToGetParametersFrom = templateInterface.GetGenericArguments().First();
+            }
+
+            var parameters = typeToGetParametersFrom.GetProperties()
```
dbolkensteyn: @tvsonar i have the feeling that `RuleParameterAttribute` doesn't have to be repeated here and in the following foreach, and I'd replace the `First` call by `Single()` so that it fails if the attribute is defined twice on the property

src/CSharp.CodeAnalysis.Common/RuleDetailBuilder.cs
tvsonar: @dbolkensteyn I added a separate test for that instead. (`TemplateRule_With_Direct_Parameters()`)

dbolkensteyn:
@tvsonar I don't seem to find where in the 2nd commit you are checking that the missing descriptions bug isn't being re-introduced?

tvsonar:
I'll need to merge some commits in this PR.

dbolkensteyn:
LGTM

dbolkensteyn:
Just pay attention to potentially failing ITs due to the renaming after the merge

PR #82, opened 2015-07-22
#57 Delegates should not be subtracted

### LINKED ISSUE #57: Rule: Delegate subtraction
Replaces http://jira.sonarsource.com/browse/LINTVS-93

In C# the operators `+`, `+=`, `-`, and `-=` are defined for delegates. These operate on lists of delegates. So subtraction might not result in what you'd expect.

```
var sum = x + y;
sum(); //executes x and y
(sum - x)(); //executes y;
(sum - (y + x))();// executes x and y, because [xy]-[yx]=[xy] as [xy] doesn't contain [yx] ordered list
```

As you can see line 3 works as expected, line 4 doesn't. We have no easy way of finding a difference between them in the analyzers, so we'll report on both of them. (R# does the same.)

R# ID: *DelegateSubtraction*
more info: https://confluence.jetbrains.com/display/ReSharper/Delegate+subtraction+has+unpredictable+semantics

Implements http://jira.sonarsource.com/browse/RSPEC-3172

tvsonar:
Fixed #57

### DIFF
--- a/src/SonarLint/Rules.Description/S3172.html
+++ b/src/SonarLint/Rules.Description/S3172.html
@@ -1,0 +1,32 @@
+<p>
+    In C#, delegates can be added together to chain their execution, and subtracted to remove their execution from the chain.
+</p>
+<p>
+    Subtracting a chain of delegates from another one might yield unexpected results as shown hereunder - and is likely to be a bug
+</p>
+
+<h2>Noncompliant Code Example</h2>
+<pre>
+MyDelegate first, second, third, fourth;
+first = () => Console.Write("1");
+second = () => Console.Write("2");
+third = () => Console.Write("3");
+fourth = () => Console.Write("4");
+
+MyDelegate chain1234 = first + second + third + fourth; // Compliant - chain sequence = "1234"
+MyDelegate chain12 = chain1234 - third - fourth; // Compliant - chain sequence = "12"
+
+// The chain sequence of "chain23" will be "1234" instead of "23"!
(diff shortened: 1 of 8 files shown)


src/CSharp.CodeAnalysis/Rules/DelegateSubtraction.cs
```diff
@@ -0,0 +1,81 @@
...
+using SonarQube.CSharp.CodeAnalysis.Common.Sqale;
+using SonarQube.CSharp.CodeAnalysis.Helpers;
+
+namespace SonarQube.CSharp.CodeAnalysis.Rules
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [SqaleConstantRemediation("30min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.LogicReliability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags("suspicious")]
```
dbolkensteyn: @tvsonar `CRITICAL` severity should be used with the `bug` tag, and `MAJOR` at most for the `suspicious` one.

It seems to me that  `bug` will be more appropriate for this rule - I've updated the RSPEC accordingly

dbolkensteyn:
Let's discuss this rule

dbolkensteyn:
I've updated http://jira.sonarsource.com/browse/RSPEC-3172 as discussed

dbolkensteyn:
LGTM

PR #84, opened 2015-07-22
#83 Rule S2930 reports false positives for generated WinForms

### LINKED ISSUE #83: Rule S2930 reports false positives for generated WinForms
The problem seems to be that one of the members is an interface that implements `IDisposable`, and we are checking only for implementations of `IDisposable.Dispose()` and not for exact matches.

tvsonar:
Fixes #83


dbolkensteyn:
LGTM

PR #87, opened 2015-07-23
#63 Don’t cast from an interface to a concrete type

### LINKED ISSUE #63: Rule: Don’t cast from an interface to a concrete type
Replaces http://jira.sonarsource.com/browse/LINTVS-170

Whenever you need to cast from an interface to a concrete type that means that something is wrong with the abstractions you are using, something is missing from the interface. So you should fix the interface instead of casting, because casting can throw an exception during runtime. 

Attachment: https://gist.github.com/dbolkensteyn/0ac2cc784069eb1526e7

Implements http://jira.sonarsource.com/browse/RSPEC-3215

tvsonar:
Fixes #63

### DIFF
--- a/src/SonarLint/Rules/CastConcreteTypeToInterface.cs
+++ b/src/SonarLint/Rules/CastConcreteTypeToInterface.cs
@@ -1,0 +1,96 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarLint.Common;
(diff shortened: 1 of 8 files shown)


dbolkensteyn:
@tvsonar Can you please rebase this one before I test it? (to avoid running into the issues due to the project renaming)

tvsonar:
@dbolkensteyn Done.

src/SonarLint/Rules/CastConcreteTypeToInterface.cs
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    var castExpression = (BinaryExpressionSyntax)c.Node;
+                    var castedTo = c.SemanticModel.GetTypeInfo(castExpression.Right).Type;
+                    var castedFrom = c.SemanticModel.GetTypeInfo(castExpression.Left).Type;
+                    ReportIssue(castedTo, castedFrom, c);
+                },
+                SyntaxKind.AsExpression);
+        }
+
+        public static void ReportIssue(ITypeSymbol castedTo, ITypeSymbol castedFrom,
```
dbolkensteyn: @tvsonar the method name here is not ideal: the method doesn't just report an issue as its name indicates, but it actually does it conditionally - it therefore would make more sense to call it `CheckSomething()`

src/SonarLint/Rules/CastConcreteTypeToInterface.cs
```diff
@@ -0,0 +1,97 @@
...
+    [Tags("design")]
+    public class CastConcreteTypeToInterface : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S3215";
+        internal const string Title = "\"interface\" instances should not be cast to concrete types";
+        internal const string Description =
+            "Needing to cast from an interface to a concrete type indicates that something is wrong with the abstractions in use, " +
+            "likely that something is missing from the interface. Instead of casting to a discrete type, the missing functionality " +
+            "should be added to the interface. Otherwise there is the risk of runtime exceptions.";
+        internal const string MessageFormat = "Remove this cast and edit the \"interface\" to add the missing functionality.";
```
dbolkensteyn: @tvsonar this issue message is strange: "interface" should be written without the quotes around it - else it sounds like we're talking about a fake interface...

src/SonarLint/Rules/CastConcreteTypeToInterface.cs
```diff
@@ -0,0 +1,97 @@
...
+        internal const string Category = "SonarQube";
+        internal const Severity RuleSeverity = Severity.Major;
+        internal const bool IsActivatedByDefault = false;
```
dbolkensteyn: @tvsonar I've updated the rule to be enabled by default as discussed

dbolkensteyn:
LGTM - but see the comments

Issue #85, opened 2015-07-22
Do not report issues on *.g.cs or other generated filenames


tvsonar:
Roslyn should ideally handle the reporting of issues on generated code out of the box, but this is not yet the case:
* https://github.com/dotnet/roslyn/issues/3940
* https://github.com/dotnet/roslyn/issues/1751

A very common scenario is `*.Designer.cs` files - which are generated by Visual Studio for Windows Form files for example.

With this issue, we want to exclude any file name containing (case insensitively):
* .g.
* .generated.
* .designer.
* TemporaryGeneratedFile_

This issue relates to #164

dbolkensteyn:
Also have a look at http://geekswithblogs.net/terje/archive/2008/11/10/hiding-generated-code-from-code-analysis-metrics-and-test-coverage.aspx

dbolkensteyn:
What other tools are doing:

So at the moment everyone is implementing their own logic to do the filtering:
* https://github.com/DotNetAnalyzers/StyleCopAnalyzers/commit/823a4c07c91b63a575d4f7f7f395c5094dd1ab63
* https://github.com/code-cracker/code-cracker/blob/77ada1021fc0d7121c4011665ac8402e003ea9f5/src/CSharp/CodeCracker/Extensions/CSharpGeneratedCodeAnalysisExtensions.cs
* https://github.com/DustinCampbell/CSharpEssentials/commit/3ba72c08b99dfe95b68c7339162f08b61f3ec417
* https://github.com/icsharpcode/RefactoringEssentials/commit/21b7ecf3e329a7484c7339b3e5a8b3e97e43183c

It would also be nice to incorporate the "Suppress results from generated code" option from the Code Analysis project settings. But that only seems to be an FxCop flag.

dbolkensteyn:
Issue updated as discussed

PR #88, opened 2015-07-23
#68 Redundant parentheses should not be used

### LINKED ISSUE #68: Rule: Redundant parentheses should not be used
Replaces http://jira.sonarsource.com/browse/LINTVS-177

In C# when you call a constructor, you can specify property values as well for the initialized object:
`new SomeClass(1,2,3) \{ Property1 = 5, Property2 = 15\}`

But if there is no constructor argument this looks like the following:
`new SomeClass() \{ Property1 = 5, Property2 = 15\}`, where `()` is not needed so could be simplified to `new SomeClass \{ Property1 = 5, Property2 = 15\}`

R# ID: *RedundantEmptyObjectCreationArgumentList*

Implements http://jira.sonarsource.com/browse/RSPEC-3235

tvsonar:
Fixes #68


tvsonar:
@dbolkensteyn This rule reports loads of issues on generated code. So we should really invest into #85, and then check if the situation got better here or not. Maybe we should exclude this rule from the default profile.

tvsonar:
@ganncamp Please make sure that when you validate this PR, you also check the new error locations of S1854, and S2996.

dbolkensteyn:
@tvsonar It is a bad idea to mix several pull requests. This one should only have been about the new rule, and another one (with an associated GitHub issue) should be there to update the behavior of the other rules

src/SonarLint/Rules/RedundantParentheses.cs
```diff
@@ -0,0 +1,85 @@
...
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.Readability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags("unused")]
+    public class RedundantParentheses : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S3235";
+        internal const string Title = "Redundant parentheses should not be used";
+        internal const string Description =
+            "Redundant parentheses are simply wasted keystrokes, and should be removed.";
+        internal const string MessageFormat = "Remove the redundant parentheses from this call.";
```
dbolkensteyn: @tvsonar I've updated the message into: `Remove these redundant parentheses.`

Indeed, there is no method call in the places where redundant parentheses are reported by this method.

PR #90, opened 2015-07-23
#86 Rule S2197 should not report on unsigned values

### LINKED ISSUE #86: Rule S2197 should not report on unsigned values
The rule "Modulus results should not be checked for direct equality" says that the result of a modulus can be negative. But this is not the case for unsigned values.

tvsonar:
Fixes #86

### DIFF
--- a/src/SonarLint/Rules.Description/S2197.html
+++ b/src/SonarLint/Rules.Description/S2197.html
@@ -1,26 +1,27 @@
 <p>
-    When the modulus of a negative number is calculated, the result will either be negative or zero. Thus, comparing the 
-    modulus of a variable for equality with a positive number (or a negative one) could result in false negatives.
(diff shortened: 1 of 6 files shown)


src/SonarLint/Rules.Description/S2197.html
```diff
@@ -1,26 +1,27 @@
...
+public boo IsOdd(int x) 
+{
+  return x % 2 == 1;  // Noncompliant; if x is an odd negative, x % 2 == -1
 }
 </pre>
 
 <h2>Compliant Solution</h2>
 <pre>
-public bool isOdd(int x)
+public bool IsOdd(int x)
```
dbolkensteyn: @tvsonar no need for the trailing whitespace

dbolkensteyn:
LGTM

PR #97, opened 2015-07-27
Removing the SonarLint.Common assembly

tvsonar:


dbolkensteyn:
LGTM - manually tested the VSIX and NuGet installation & uninstallation

PR #98, opened 2015-07-27
#72 Anonymous delegates should not be used to unsubscribe from Events

### LINKED ISSUE #72: Rule: Event unsubscription via anonymous delegate
Replaces http://jira.sonarsource.com/browse/LINTVS-94

Sample from http://stackoverflow.com/questions/8803064/event-unsubscription-via-anonymous-delegate:
```
if (((bool)e.NewValue))
{
    listView.PreviewTextInput += (obj,args) =>
        listView_PreviewTextInput(obj,args,listView);  // subscribes to notifications with anonymous delegate
}
else
{
    listView.PreviewTextInput -= (obj, args) => 
        listView_PreviewTextInput(obj, args, listView); //Noncompliant; unsubscribes with anonymous delegate - this is not the same anon. delegate that was added
}
```

Unsubscribing this way won't work, because a new delegate will be created which won't be equal to the one in the subscription. So you should keep a reference to the original delegate, and unsubscribe with that one:
```
EventHandler func = (o,args) =>
        listView_PreviewTextInput(o,args,listView);
if (((bool)e.NewValue))
{
    listView.PreviewTextInput += func;
}
else
{
    listView.PreviewTextInput -= func;
}
```

http://stackoverflow.com/questions/183367/unsubscribe-anonymous-method-in-c-sharp
http://stackoverflow.com/questions/8803064/event-unsubscription-via-anonymous-delegate

R# ID: *EventUnsubscriptionViaAnonymousDelegate*

Implements http://jira.sonarsource.com/browse/RSPEC-3244

tvsonar:
Fixes #72

### DIFF
--- a/src/Tests/SonarLint.UnitTest/TestCases/AnonymousDelegateEventUnsubscribe.cs
+++ b/src/Tests/SonarLint.UnitTest/TestCases/AnonymousDelegateEventUnsubscribe.cs
@@ -1,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Tests.Diagnostics
+{
+    class AnonymousDelegateEventUnsubscribe
+    {
+        public delegate void ChangedEventHandler(object sender, EventArgs e);
+        public delegate void ChangedEventHandler2(object sender);
+        public event ChangedEventHandler Changed;
+        public event ChangedEventHandler2 Changed2;
+
+        void Test()
+        {
+            Changed += (obj, args) => { };
+            Changed -= (obj, args) => { }; //Noncompliant
+
+            Changed -= (obj, args) => Console.WriteLine(); // Noncompliant - single statement
+            Changed -= (obj, args) => delegate () { }; // Noncompliant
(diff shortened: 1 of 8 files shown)


src/Tests/SonarLint.UnitTest/TestCases/AnonymousDelegateEventUnsubscribe.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    class AnonymousDelegateEventUnsubscribe
+    {
+        public delegate void ChangedEventHandler(object sender, EventArgs e);
+        public event ChangedEventHandler Changed;
+        
+        void Test()
+        {
+            Changed += (obj, args) => { };
+            Changed -= (obj, args) => { }; //Noncompliant
```
dbolkensteyn: @tvsonar I'd also test the other syntaxes:

```
Changed -= (obj, args) => Console.WriteLine(); // Noncompliant - single statement
Changed -= (obj, args) => delegate() {}; // Noncompliant
```

dbolkensteyn:
LGTM

PR #99, opened 2015-07-27
#74 "value" parameter is not used

### LINKED ISSUE #74: Rule: "value" parameter is not used
Replaces http://jira.sonarsource.com/browse/LINTVS-183

It is usually a bug not to use the `value` in a property setter, array indexer, or event's `add/remove`.

Property setter: https://msdn.microsoft.com/en-us/library/a1khb4f8.aspx
Indexer: https://msdn.microsoft.com/en-us/library/6x16t2tx.aspx
Event: https://msdn.microsoft.com/en-us/library/cc713648.aspx

R# ID: *ValueParameterNotUsed*

Implements http://jira.sonarsource.com/browse/RSPEC-3237

tvsonar:
Fixes #74, #104

### DIFF
--- a/src/SonarLint/Helpers/ProjectTypeHelper.cs
+++ b/src/SonarLint/Helpers/ProjectTypeHelper.cs
@@ -1,0 +1,39 @@
+/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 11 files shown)


tvsonar:
Fixes #74

dbolkensteyn:
Re-opened:

1. I've updated http://jira.sonarsource.com/browse/RSPEC-3237 - this clearly must be a bug detection rule
2. Single statement methods with just a `throw` statement should be ignored
3. This rule should not be executed in test projects - depends on #104

dbolkensteyn:
LGTM

PR #101, opened 2015-07-27
#76 Overriden GetHashCode/Equals calls base 'Object.GetHashCode/Equal…

### LINKED ISSUE #76: Rule: Overriden GetHashCode/Equals calls base 'Object.GetHashCode/Equals()'
Replaces http://jira.sonarsource.com/browse/LINTVS-194

The point is to raise an issue when base (super) is called and class inherits directly from Object.

If you override `GetHashCode()` or `Equals()` that most probably means that you want to provide the same hash value for different instances of the class when they are logically equivalent.

Calling `base.GetHashCode()` in the implementation is generally a good idea, because the base class might already implement a logical hash code generation algorithm for itself. However if the `base` class is `System.Object`, then calling its `GetHashCode` defeats the purpose of the override because `Object.GetHashCode()`'s value is based on the reference handle of the current object. 

```
public class Point
{
  private readonly int x;
  public MyClass(int x)
  {
    this.x = x;
  }
  public override int GetHashCode()
  {
    return x.GetHashCode() ^ base.GetHashCode(); //Noncompliant
  }
}
```
Compliant:
```
public class Point
{
  private readonly int x;
  public MyClass(int x)
  {
    this.x = x;
  }
  public override int GetHashCode()
  {
    return x.GetHashCode();
  }
}
```

R# ID:*BaseObjectGetHashCodeCallInGetHashCode*, *BaseObjectEqualsIsObjectEquals*

Implements http://jira.sonarsource.com/browse/RSPEC-3249

tvsonar:
…s()'

Fixes #76

### DIFF
--- a/src/SonarLint/Rules.Description/S3249.html
+++ b/src/SonarLint/Rules.Description/S3249.html
@@ -1,0 +1,38 @@
+<p>
+    Making a <code>base</code> call in an overridden method is generally a good idea, but not in <code>GetHashCode</code> and
+    <code>Equals</code> for classes that directly extend <code>Object</code> because those methods are based on the object reference.
+    Meaning that no two <code>Objects</code> that use those <code>base</code> methods will ever be equal or have the same hash.
+</p>
+
+<h2>Noncompliant Code Example</h2>
+<pre>
+public class Point
+{
+  private readonly int x;
+  public MyClass(int x)
+  {
+    this.x = x;
+  }
+  public override int GetHashCode()
+  {
+    return x.GetHashCode() ^ base.GetHashCode(); //Noncompliant
+  }
+}
+</pre>
+
+<h2>Compliant Solution</h2>
+<pre>
+public class Point
+{
+  private readonly int x;
+  public MyClass(int x)
+  {
+    this.x = x;
+  }
+  public override int GetHashCode()
+  {
+    return x.GetHashCode();
+  }
+}
+</pre>
+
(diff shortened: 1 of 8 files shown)


src/SonarLint/Rules/GetHashCodeEqualsOverride.cs
```diff
@@ -0,0 +1,118 @@
...
+            context.RegisterCodeBlockStartAction<SyntaxKind>(
+                cb =>
+                {
+                    var methodDeclaration = cb.CodeBlock as MethodDeclarationSyntax;
+                    if (methodDeclaration == null)
+                    {
+                        return;
+                    }
+
+                    var methodSymbol = cb.SemanticModel.GetDeclaredSymbol(methodDeclaration);
```
dbolkensteyn: @tvsonar `.OwningSymbol` ?

src/SonarLint/Rules.Description/S3249.html
```diff
@@ -0,0 +1,38 @@
+<p>
+    Making a <code>base</code> call in an overridden method is generally a good idea, but not in <code>GetHashCode</code> and
```
dbolkensteyn: @tvsonar I've slightly evolved the title and description to make it explicit that this rule only runs when directly extending `Object`

src/SonarLint/Rules/GetHashCodeEqualsOverride.cs
```diff
@@ -0,0 +1,118 @@
...
+                                return;
+                            }
+
+                            var baseCall = memberAccess.Expression as BaseExpressionSyntax;
+                            if (baseCall == null)
+                            {
+                                return;
+                            }
+
+                            var objectType = invokedMethod.ContainingType;
```
dbolkensteyn: @tvsonar nice!

dbolkensteyn:
LGTM, nice rule @tvsonar !

PR #106, opened 2015-07-28
#65 Drop the S2325 rule which generates too many false positives

### LINKED ISSUE #65: Drop the S2325 rule which generates too many false positives
S2325 reports false positives on 
* test methods
* MVC and WebApi action methods
* Serializable and DataContract attributed properties.
* And basically on any method and property that needs to be reached through reflection, because most/many/several reflection based frameworks don't handle static properties/methods.

We could ignore 
* methods/properties that have attributes on them, 
* properties of classes with attributes
* methods of classes that derive from `Controller`.

But there will always be cases that report false positives, or result in false negatives. The overall gain from this rule is minimal, so considering these, we decided to drop the rule.

tvsonar:
Fixes #65


dbolkensteyn:
LGTM

PR #116, opened 2015-07-29
Issue #110 Fade out redundant casts and fix error location

### LINKED ISSUE #110: Fade out redundant casts and fix error location

tvsonar:
Fixes #110


dbolkensteyn:
LGTM, however:

1. Unit tests should be updated to check for the change of the error location
2. The static case should ideally fade also both parentheses (but not the argument) in `Enumerable.OfType<int>(list)` - but I guess this is not possible without creating 2 distinct issues?

tvsonar:
@dbolkensteyn Yes, the report location needs to be one single location.

PR #117, opened 2015-07-29
Issue #108 Fade out unnecessary boolean literals, and fixing quality flaw

### LINKED ISSUE #108: Fade out unnecessary boolean literals

tvsonar:
Fixes #108

### DIFF
--- a/src/SonarLint/Rules/BooleanLiteralUnnecessary.cs
+++ b/src/SonarLint/Rules/BooleanLiteralUnnecessary.cs
@@ -28,6 +28,7 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using SonarLint.Common;
 using SonarLint.Common.Sqale;
 using SonarLint.Helpers;
+using Microsoft.CodeAnalysis.Text;
 
 namespace SonarLint.Rules
 {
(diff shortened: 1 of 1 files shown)


dbolkensteyn:
The error location should include the binary operator:

{code}
SomeBooleanFunction() == true;
                                        ~~~~~~ this whole part is unused
{code}

The same logic goes for `!=`, `&&`, `||` and `!`

src/SonarLint/Rules/BooleanLiteralUnnecessary.cs
```diff
@@ -40,36 +41,98 @@ public class BooleanLiteralUnnecessary : DiagnosticAnalyzer
...
+                        c.ReportDiagnostic(Diagnostic.Create(Rule, logicalNot.Operand.GetLocation(), logicalNot.Operand.ToString()));
+                        return;
                     }
                 },
-                SyntaxKind.TrueLiteralExpression,
-                SyntaxKind.FalseLiteralExpression);
+                SyntaxKind.LogicalNotExpression);
+        }
+
+        private static void ReportIssue(BinaryExpressionSyntax binaryExpression, bool leftSide, ExpressionSyntax expressionToCheck,
```
dbolkensteyn: @tvsonar this method doesn't do what it saids - it doesn't simply report an issue: It should likely more be called "CheckSomething" - or the `SyntaxFactory.AreEquivalent()` moved out

src/SonarLint/Rules/BooleanLiteralUnnecessary.cs
```diff
@@ -40,36 +41,98 @@ public class BooleanLiteralUnnecessary : DiagnosticAnalyzer
...
+                },
+                SyntaxKind.EqualsExpression,
+                SyntaxKind.LogicalAndExpression);
+
+            context.RegisterSyntaxNodeAction(
+                c =>
+                {
+                    var binaryExpression = (BinaryExpressionSyntax)c.Node;
+
+                    ReportIssue(binaryExpression, true, FalseExpression, true, c);
```
dbolkensteyn: @tvsonar it is quite a bad idea to pass boolean values to methods: looking back at the code, it is not clear what this boolean flag does.

A better way is to add one new properly-named method for the `true` case, another one for the `false` one, which then both simply call the real method with the right flag

dbolkensteyn:
LGTM

PR #120, opened 2015-07-30
Issue #85 Don't report issues on generated code

### LINKED ISSUE #85: Do not report issues on *.g.cs or other generated filenames

tvsonar:
Fixes #85


src/Tests/SonarLint.UnitTest/Verifier.cs
```diff
@@ -87,6 +87,23 @@ public static void Verify(string path, DiagnosticAnalyzer diagnosticAnalyzer)
             }
         }
 
+        internal static void VerifyEmpty(string name, string content, DiagnosticAnalyzer diagnosticAnalyzer)
```
dbolkensteyn: @tvsonar this method should go to `DiagnosticAnalyzerContextHelperTest` : the only place where it is expected to be used

src/SonarLint/Rules/GenericTypeParameterUnused.cs
```diff
@@ -74,7 +74,8 @@ public override void Initialize(AnalysisContext context)
                         }
 
                         var declarationSymbol = c.SemanticModel.GetDeclaredSymbol(c.Node);
-                        if (declarationSymbol == null)
+                        if (declarationSymbol == null ||
+                            declarationSymbol.IsGenerated(c.CancellationToken))
```
dbolkensteyn: @tvsonar :-1: such a mechanism is not worth to be introduced for a single rule : let's either re-use of of the other exclusion mechanism if possible - or live with the few false positives this rule might generate for now

src/SonarLint/Helpers/DiagnosticAnalyzerContextHelper.cs
```diff
@@ -0,0 +1,285 @@
...
+            {
+                context.ReportDiagnostic(diagnostic);
+            }
+        }
+
+        #endregion
+
+        #region ISymbol.IsGenerated
+
+        public static bool IsGenerated(this ISymbol symbol, CancellationToken cancellationToken)
```
dbolkensteyn: :-1: for this method - see the comment below in `GenericTypeParameterUnused` where it is used

src/SonarLint/Helpers/DiagnosticAnalyzerContextHelper.cs
```diff
@@ -0,0 +1,285 @@
...
+            where TLanguageKindEnum : struct
+        {
+            return context.CodeBlock.SyntaxTree.IsGenerated(context.CancellationToken);
+        }
+
+        #endregion
+
+        #region ReportDiagnosticInNonGenerated
+
+        public static void ReportDiagnosticInNonGenerated(this CompilationAnalysisContext context, Diagnostic diagnostic)
```
dbolkensteyn: @tvsonar what about the name `ReportDiagnostic**If**NonGenerated` ?

src/SonarLint/Helpers/DiagnosticAnalyzerContextHelper.cs
```diff
@@ -0,0 +1,285 @@
...
+        {
+            if (!diagnostic.Location.SourceTree.IsGenerated(context.CancellationToken))
+            {
+                context.ReportDiagnostic(diagnostic);
+            }
+        }
+
+        public static void ReportDiagnosticInNonGenerated(this SymbolAnalysisContext context, Diagnostic diagnostic)
```
dbolkensteyn: @tvsonar same as above

src/SonarLint/Helpers/DiagnosticAnalyzerContextHelper.cs
```diff
@@ -0,0 +1,285 @@
...
+                        action(c);
+                    }
+                });
+        }
+
+        #endregion
+
+        #region  *AnalysisContext.IsGenerated
+
+        public static bool IsGenerated(this SyntaxNodeAnalysisContext context)
```
dbolkensteyn: @tvsonar should be `private`

src/SonarLint/Helpers/DiagnosticAnalyzerContextHelper.cs
```diff
@@ -0,0 +1,285 @@
...
+        {
+            return context.Node.SyntaxTree.IsGenerated(context.CancellationToken);
+        }
+        public static bool IsGenerated(this SyntaxTreeAnalysisContext context)
```
dbolkensteyn: @tvsonar should be `private`

src/SonarLint/Helpers/DiagnosticAnalyzerContextHelper.cs
```diff
@@ -0,0 +1,285 @@
...
+        {
+            return context.Tree.IsGenerated(context.CancellationToken);
+        }
+        public static bool IsGenerated<TLanguageKindEnum>(this CodeBlockStartAnalysisContext<TLanguageKindEnum> context)
```
dbolkensteyn: @tvsonar should be `private`

src/SonarLint/Helpers/DiagnosticAnalyzerContextHelper.cs
```diff
@@ -0,0 +1,285 @@
...
+
+            return symbol.DeclaringSyntaxReferences.Any(syntaxReference =>
+                syntaxReference.SyntaxTree.IsGenerated(cancellationToken));
+        }
+
+        #endregion
+
+        #region SyntaxTree.IsGenerated
+
+        private static WeakReference<ImmutableDictionary<SyntaxTree, bool>> cache =
```
dbolkensteyn: @tvsonar I think this is a mistake : the whole immutable dictionary will get garbage collected. Isn't there a weak dictionary where individual elements can disappear if there is memory contention?

src/SonarLint/Helpers/DiagnosticAnalyzerContextHelper.cs
```diff
@@ -0,0 +1,285 @@
...
+            new WeakReference<ImmutableDictionary<SyntaxTree, bool>>(ImmutableDictionary<SyntaxTree, bool>.Empty);
+
+        private static bool IsGenerated(this SyntaxTree tree, CancellationToken cancellationToken)
```
dbolkensteyn: let's discuss the actual strategy with @duncanMS

src/SonarLint/Helpers/DiagnosticAnalyzerContextHelper.cs
tvsonar: @dbolkensteyn Works for me.

src/SonarLint/Helpers/DiagnosticAnalyzerContextHelper.cs
tvsonar: @dbolkensteyn  you are right. I thought it's okay to get rid of the complete dictionary. But as I think about it, this might never get GC'd at all. There is always at least one tree in the dictionary that is being referenced from Roslyn, so the dictionary can't be GC'd.
There is a `ConditionalWeakTable` class, which is not exactly a dictionary, but can attach properties to objects without holding a strong reference to the "key". I'll change it to that one, I think StyleCop is using exactly that.

src/SonarLint/Rules/GenericTypeParameterUnused.cs
tvsonar: @dbolkensteyn `ISymbol.IsGenerated` is a 5 line method, I don't think that's such a heavy mechanism, and it removes possible false positives. If this was a symbol analyzer, I would have created a `RegisterSymbolActionInNonGenerated()`, and this method would still be used in only one single analyzer. Whenever will have more symbol based analyzers, we can reuse the `ISymbol.IsGenerated`.

src/SonarLint/Helpers/DiagnosticAnalyzerContextHelper.cs
```diff
@@ -0,0 +1,190 @@
...
+
+            return symbol.DeclaringSyntaxReferences.Any(syntaxReference =>
+                syntaxReference.SyntaxTree.IsGenerated(cancellationToken));
+        }
+
+        #endregion
+
+        #region SyntaxTree.IsGenerated
+
+        private static ConditionalWeakTable<SyntaxTree, StrongBox<bool?>> cache =
```
dbolkensteyn: @tvsonar do we need this, checking on filenames is pretty fast? If that's not fast enough - it's better to cache based on the filename than on the `SyntaxTree` - because syntax trees will be different with every change made to the file

src/SonarLint/Rules/GenericTypeParameterUnused.cs
dbolkensteyn: @tvsonar I'm still against it - it's not just 5 lines: the unit test here is missing, so this method call looks like dead code - moreover, this didn't remove any false positive on ruling - so actually all tests are green without this call

dbolkensteyn:
LGTM

PR #123, opened 2015-07-31
Issue #62 Multiple "OrderBy" calls should not be used

### LINKED ISSUE #62: Rule: Do not chain "OrderBy" calls
Replaces http://jira.sonarsource.com/browse/LINTVS-165

Calling `OrderBy` methods in a chain doesn't make sense, because only the last one will have effect.
```
var x = personList
  .OrderBy(person => person.Age)
  .OrderBy(person => person.Name)
  .ToList();
```
Here the list will be sorted by person name.

If you want to order same aged people by name as well, you need to use `ThenBy`
```
var x = personList
  .OrderBy(person => person.Age)
  .ThenBy(person => person.Name)
  .ToList();
```

Resharper ID of this rule is: *MultipleOrderBy*

Implements http://jira.sonarsource.com/browse/RSPEC-3169

tvsonar:
Fixes #62


dbolkensteyn:
LGTM

PR #125, opened 2015-07-31
Issue #73 "GC.SuppressFinalize" is invoked for type without destructor

### LINKED ISSUE #73: Rule: "GC.SuppressFinalize" is invoked for type without destructor
Replaces http://jira.sonarsource.com/browse/LINTVS-191

You can ask the CLR not to call the finalizer on an object by calling `GC.SuppressFinalize()` with the object as a parameter. However, if there is no finalizer defined for the type, there is no point in calling this method.

R# ID: *GCSuppressFinalizeForTypeWithoutDestructor*

Implements http://jira.sonarsource.com/browse/RSPEC-3234

tvsonar:
Fixes #73

### DIFF
--- a/src/SonarLint/Rules.Description/S3234.html
+++ b/src/SonarLint/Rules.Description/S3234.html
@@ -1,0 +1,31 @@
+<p>
+    <code>GC.SuppressFinalize</code> asks the Common Language Runtime not to call the finalizer of an object. This is useful when
+    implementing the dispose pattern where object finalization is already handled in <code>IDisposable.Dispose</code>. However, it
+    has no effect if there is no finalizer defined in the object's type, so using it in such cases is just confusing.
+</p>
+<p>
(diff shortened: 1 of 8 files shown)


src/SonarLint/Rules.Description/S3234.html
```diff
@@ -0,0 +1,31 @@
+<p>
+    <code>GC.SuppressFinalize</code> asks the Common Language Runtime not to call the finalizer of an object. This is useful when
+    implementing the dispose pattern where object finalization is already handled in <code>IDisposable.Dispose</code>. However, it
+    has no effect if there is no finalizer defined in the object's type, so using it in such cases is just confusing.
+</p>
+<p>
+    This rule raises an issue when <code>GC.SuppressFinalize</code> is called for objects of <code>sealed</code> types.
```
dbolkensteyn: @tvsonar the description is incorrect here: it must be `sealed` and not have a finalizer

dbolkensteyn:
LGTM except for the little issue with the description which I let you fix

PR #128, opened 2015-08-03
Issue #71 Generic type parameters should be co/contravariant when possible

### LINKED ISSUE #71: Rule: Generic type parameter could be defined with "in" or "out" modifier
Replaces http://jira.sonarsource.com/browse/LINTVS-184

From MSDN: 
You can mark a generic parameter `out` if the type parameter is used only as a return type of interface methods and not used as a type of method arguments. (There are some exceptions) https://msdn.microsoft.com/en-us/library/dd469487.aspx
Similarly it can be marked `in` if the type is only used as method argument type. (https://msdn.microsoft.com/en-us/library/dd469484.aspx)

Good summary here on interface covariance and contravariance: http://stackoverflow.com/questions/10956993/out-t-vs-t-in-generics

Implementation notes: both interfaces and delegates should be handled

R# ID: *TypeParameterCanBeVariant*

let's see an example for the `in`. This might not be the most meaningful example but it shows the concept:
```
public class Fruit
{ }
public class Orange : Fruit
{ }

interface IConsumer<in T> where T: Fruit
{
    bool Eat(T fruit);
}
public class Person : IConsumer<Fruit>
{
    public bool Eat(Fruit fruit)
    {
        return true;
    }
}

class Program
{
    public static void Main()
    {
        IConsumer<Fruit> fruitConsumer = new Person();

        fruitConsumer.Eat(new Fruit());
        fruitConsumer.Eat(new Orange());

        IConsumer<Orange> orangeConsumer = fruitConsumer; //This one would raise a compiler error if there was no "in" in the IConsumer<in T>
        orangeConsumer.Eat(new Orange());
    }
}
```

A better example for `out`. Consider the following:
```
Orange[] oranges = new[] { new Orange(), new Orange() };
Fruit[] fruits = oranges; //This works because of the array covariance.
var firstFruit = fruits.First();
```
This works with arrays, so it should also work with `IEnumerable`s. But previously we had no covariance with generics, so `MyGeneric<Base>` had nothing to do with `MyGeneric<Derived>`, even though `Derived : Base`.

`IEnumerable<T>` is defined as `public interface IEnumerable<out T> : IEnumerable`, so we can write the following
```
IEnumerable<Orange> oranges = new[] { new Orange(), new Orange() };
IEnumerable<Fruit> fruits = oranges; //This works because of the "out"
var firstFruit = fruits.First();
```
Which is nice. And at the same time we can't get to runtime exceptions contrary to arrays:
```
Orange[] oranges = new[] { new Orange(), new Orange() };
Fruit[] fruits = oranges; //This works because of the "out"
fruits[0] = new Apple(); // this one fails in runtime. But if we are using IEnumerable, we don't even have this option.
```
The covariant interface protects against this runtime failure.

Implements http://jira.sonarsource.com/browse/RSPEC-3246

tvsonar:
Fixes #71

### DIFF
--- a/src/SonarLint/Rules.Description/S3246.html
+++ b/src/SonarLint/Rules.Description/S3246.html
@@ -1,0 +1,35 @@
+<p>
+    In the interests of making code as usable as possible, interfaces and delegates with generic parameters should use the
+    <code>out</code> and <code>in</code> modifiers when possible to make the interfaces and delegates covariant and contravariant,
+    respectively.
+</p>
+<p>
+    The <code>out</code> keyword can be used when the type parameter is used only as a return type in the interface or delegate.
+    Doing so makes the parameter covariant, and allows interface and delegate instances created with a sub-type to be used as
+    instances created with a base type. The most notable example of this is <code>IEnumerable&lt;out T&gt;</code>, which allows
+    the assignment of an <code>IEnumerable&lt;string&gt;</code> instance to an <code>IEnumerable&lt;object&gt;</code> variable,
+    for instance.
+</p>
+<p>
+    The <code>in</code> keyword can be used when the type parameter is used only as a method parameter in the interface or a
+    parameter in the delegate. Doing so makes the parameter contravariant, and allows interface and delegate instances created
+    with a base type to be used as instances created with a sub-type. I.e. this is the inversion of covariance. The most notable
+    example of this is the <code>Action&lt;in T&gt;</code> delegate, which allows the assignment of an
+    <code>Action&lt;object&gt;</code> instance to a <code>Action&lt;string&gt;</code> variable, for instance.
+</p>
+
+<h2>Noncompliant Code Example</h2>
+<pre>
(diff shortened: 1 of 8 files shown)


tvsonar:
@dbolkensteyn Do we need to add some sort of copyright notice that this is based on `Microsoft.CodeAnalysis.CSharp.Symbols.VarianceSafety`?

dbolkensteyn:
@tvsonar let's check that with Freddy (it seems indeed required)

@tvsonar Duncan's suggestion to keep the file as close to the original one is a good one - and one we should go for. It will allow us in the future to identify that these bits were copied (and check if that still is required) - and also synchronize with updates the Roslyn team will make more easily.

tvsonar:
@dbolkensteyn Okay, I've assigned the corresponding issue to @fmallet

dbolkensteyn:
email sent @tvsonar to know how to handle the duplication.

PR #133, opened 2015-08-04
Issue #131 Rules on disposables fail when the "IDisposable.Dispose" method can't be found

### LINKED ISSUE #131: Rules on disposables fail when the "IDisposable.Dispose" method can't be found

tvsonar:
Fixes #131


dbolkensteyn:
LGTM

PR #134, opened 2015-08-04
Issue #81 Fields should not be initialized to default values

### LINKED ISSUE #81: Rule: Fields should not be initialized to default values
Replaces http://jira.sonarsource.com/browse/LINTVS-98

A reference type field is automatically initialized to `null`, while a value type is initialized to the default value of the type. So initializing a field to its default value is useless. This initialization can happen both in a field initializer or in the constructor.
R# only reports on field initializers, and not constructor ones (it might not be that simple to recognize the latter).

```
class X
{
  public int field = 0; //Noncompliant, it is already 0, it has a minor performance impact that we reinitialize
  public object o = null; //Noncompliant (R# doesn't warn on this one)
}
```

R# ID: *RedundantDefaultFieldInitializer*

http://www.codeproject.com/Articles/10439/Gain-performance-by-not-initializing-variables
http://blog.codinghorror.com/for-best-results-dont-initialize-variables/

Implements http://jira.sonarsource.com/browse/RSPEC-3052

tvsonar:
Fixes #81


src/SonarLint/Rules/FieldInitializedToDefault.cs
```diff
@@ -0,0 +1,163 @@
...
+        internal const string Title = "Fields should not be initialized to default values";
+        internal const string Description =
+            "The compiler automatically initializes class fields to their default values before setting them with any " +
+            "initialization values, so not only is there no need to explicitly set a field to its default value, but doing " +
+            "so is slightly less performant. Further, under the logic that cleaner code is better code, it's considered " +
+            "poor style to do so.";
+        internal const string MessageFormat = "Remove this initialization to \"{0}\", the compiler will do that for you.";
+        internal const string Category = "SonarQube";
+        internal const Severity RuleSeverity = Severity.Minor;
+        internal const bool IsActivatedByDefault = false;
```
dbolkensteyn: @tvsonar let's enable this one by default for C# - it won't generate noise because it only shades the initializer.

And as this one didn't trigger any issue on Ruling - i don't expect that much noise even in SonarQube

src/SonarLint/Rules/FieldInitializedToDefault.cs
```diff
@@ -0,0 +1,163 @@
...
+                case SpecialType.System_Single:
+                case SpecialType.System_UInt16:
+                case SpecialType.System_UInt32:
+                case SpecialType.System_UInt64:
+                    int constantValue;
+                    return SillyBitwiseOperation.TryGetConstantIntValue(variable.Initializer.Value, out constantValue) &&
+                        constantValue == 0;
+            }
+
+            return false;
```
dbolkensteyn: @tvsonar sonarlint says to move this to the `default` case of the switch

src/SonarLint/Rules/FieldInitializedToDefault.cs
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        private static object GetDefaultValue(ITypeSymbol type)
```
dbolkensteyn: @tvsonar this method is actually unused... and nothing is reporting it in VS? We must add this rule in SonarLint VS!

src/SonarLint/Rules.Description/S3052.html
```diff
@@ -0,0 +1,23 @@
...
+    The compiler automatically initializes class fields to their default values before setting them with any initialization
+    values, so not only is there no need to explicitly set a field to its default value, but doing so is slightly less
+    performant. Further, under the logic that cleaner code is better code, it's considered poor style to do so.
+</p>
+
+<h2>Noncompliant Code Example</h2>
+<pre>
+class X
+{
+  public int field = 0; // Noncompliant; defaults to 0. Reinitializing incurs minor performance penalty
```
dbolkensteyn: @tvsonar please drop the performance penalty part... I've just checked and this simply gets drop at bytecode generation time

src/SonarLint/Rules.Description/S3052.html
```diff
@@ -0,0 +1,23 @@
+<p>
+    The compiler automatically initializes class fields to their default values before setting them with any initialization
+    values, so not only is there no need to explicitly set a field to its default value, but doing so is slightly less
+    performant. Further, under the logic that cleaner code is better code, it's considered poor style to do so.
```
dbolkensteyn: @tvsonar please drop the performance reason

src/SonarLint/Rules/FieldInitializedToDefault.cs
```diff
@@ -0,0 +1,163 @@
...
+                },
+                SyntaxKind.FieldDeclaration);
+        }
+
+        private static bool CheckDefaultExpressionInitializer(VariableDeclaratorSyntax variable, IFieldSymbol variableSymbol,
+            SemanticModel semanticModel)
+        {
+            var defaultValue = variable.Initializer.Value as DefaultExpressionSyntax;
+            return defaultValue != null &&
+                variableSymbol.Type.Equals(semanticModel.GetTypeInfo(defaultValue.Type).Type);
```
dbolkensteyn: @tvsonar this check is not required : `struct` cannot inherit from one another, and `default(ClassA) == default(ClassB) == null`

So there is a false negative here currently:
{code}
class A {}
class B : A {}

public A MyField = default(B); // Compiles - but not reported
{code}

dbolkensteyn:
LGTM - but see comments

PR #136, opened 2015-08-05
Issue #75 Unused method parameters should be removed

### LINKED ISSUE #75: Rule: Unused method parameter
Replaces http://jira.sonarsource.com/browse/LINTVS-186

If a method's parameter is not used, then it can be removed from the signature. (Except for interface implementations, overrides, virtual, abstract, partial methods)

R# ID: *UnusedParameter*

Implements http://jira.sonarsource.com/browse/RSPEC-1172

tvsonar:
Fixes #75


src/SonarLint/Rules/MethodParameterUnused.cs
```diff
@@ -0,0 +1,130 @@
...
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.Understandability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags("misra", "unused")]
+    public class MethodParameterUnused : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S1172";
+        internal const string Title = "Unused method parameters should be removed";
+        internal const string Description =
+            "Unused parameters are misleading. Whatever the value passed to such parameters is, the behavior will be the same.";
+        internal const string MessageFormat = "Remove this unused method parameter.";
```
dbolkensteyn: @tvsonar can you add the parameter's name to the issue message so that it is also usable from the SonarQube UI?

dbolkensteyn:
LGTM - but see comment

PR #137, opened 2015-08-05
Issue #56 Inner class members should not shadow outer class "static" or type members

### LINKED ISSUE #56: Rule: Inner class members should not shadow outer class "static" or type members
Replaces http://jira.sonarsource.com/browse/LINTVS-97

```
class Outer
{
  public static int A;

  public class Inner
  {
    public int A; //Noncompliant
    public int MyProp
    {
      get { return A; }  // Which A does this return?
    }
  }
}
```

```
class Outer
{
  public static int B;

  public class Inner
  {
    public int A;

    public int MyProp
    {
      get { return A; }
    }
  }
}
```

The main problem here is that if I have a property that uses `A` in the inner class, and then I decide to rename `A` to `Aaaa`, then the code still compiles and executes fine, but the functionality is changed as in this case the property references the `Outer.A`. (Check `MyProp` in the noncompliant sample)

R# rule ID: *MemberHidesStaticFromOuterClass*

Implements http://jira.sonarsource.com/browse/RSPEC-3218

tvsonar:
Fixes #56

### DIFF
--- a/src/SonarLint/Rules/MemberShadowsOuterStaticMember.cs
+++ b/src/SonarLint/Rules/MemberShadowsOuterStaticMember.cs
@@ -1,0 +1,285 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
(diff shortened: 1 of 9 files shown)


src/SonarLint/Rules/MemberShadowsOuterStaticMember.cs
```diff
@@ -0,0 +1,285 @@
...
+    [Tags("design", "pitfall")]
+    public class MemberShadowsOuterStaticMember : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S3218";
+        internal const string Title = "Inner class members should not shadow outer class \"static\" or type members";
+        internal const string Description =
+            "It's possible to name the members of an inner class the same as the \"static\" members of its enclosing class - " +
+            "possible, but a bad idea. That's because maintainers may be confused about which members are being used where. " +
+            "Instead the inner class' members should be renamed and all the references updated.";
+        internal const string MessageFormat = "Rename this {0}.";
```
dbolkensteyn: @tvsonar as discussed, the issue message should be updated to explain why

dbolkensteyn:
LGTM except for the issue message

PR #138, opened 2015-08-05
Issue #61 Don’t hide inherited members

### LINKED ISSUE #61: Rule: Don’t hide inherited members
Replaces http://jira.sonarsource.com/browse/LINTVS-169

In C# if you have a derived class with members of the same signature as in the base class, you should use the `new` keyword to specify that you want to hide the base member. If you don't specify it CS0108 compiler warning will be shown. 

However, hiding a member can result in strange issues. So instead of hiding a member investigate why the member needed to be hidden.

For such an issue see the attached sample from one of our MS partners.

The basic issue is that if you have a single object and you reference it through both the derived and the base type, different method/property implementations will be called, even different fields will be accessed:
```
public  class Base
{
    public int i;
}

public class Derived : Base
{
    public new int i;
}

//....

var d = new Derived();
var b = (Base) x;
d.i = 5;
Console.WriteLine(b.i); //prints 0
```

Attachment: https://gist.github.com/dbolkensteyn/95ea63ebaac941bbce8a

~~Implements http://jira.sonarsource.com/browse/RSPEC-2387~~

**UPDATE:**
Implements https://jira.sonarsource.com/browse/RSPEC-3434 Note, fields are already handled by RSPEC-2387.

tvsonar:
Fixes #61


src/SonarLint/Rules.Description/S2387.html
```diff
@@ -1,43 +1,79 @@
 <p>
-    Having a variable with the same name in two unrelated classes is fine, but do the same thing within a class hierarchy
-    and you'll get confusion at best, chaos at worst. Perhaps even worse is the case where a child class field varies from 
-    the name of a parent class only by case.
+    Shadowing parent class members by creating fields, properties and methods with the same signatures as non-<code>virtual</code>
```
dbolkensteyn: All the signature clashes are already handled out of the box by Visual Studio: CS0108

src/SonarLint/Rules.Description/S2387.html
```diff
@@ -1,43 +1,79 @@
...
+    parent class members can result in seemingly strange behavior if an instance of the child class is cast to the parent class.
```
dbolkensteyn: this whole description is different than the main RSPEC - please align both (your version does seem better)

dbolkensteyn:
This rule needs to be fully reworked because it largely overlaps with CS0108

dbolkensteyn:

PR #140, opened 2015-08-06
Fixing rule metadata

tvsonar:


src/SonarLint.Extra/Rules.Description/S101.html
```diff
@@ -3,18 +3,14 @@
 This rule allows to check that all class names match a provided regular expression.
 </p>
 
-<p>The following code:</p>
-
+<h2>Noncompliant Code Example</h2>
+<p>With default provided regular expression <code>^[A-Z][a-zA-Z0-9]*$</code>:</p>
```
dbolkensteyn: @tvsonar this does not match the actual default value

dbolkensteyn:
@tvsonar Can you please configure the whitespace removal tool to do so only on manually edited lines - it pollutes diffs quite a lot - making it harder to spot real changes

@tvsonar There was one wrong default value issue for which a C# sub-task should be created - but the rest LGTM

PR #188, opened 2015-08-11
Issue #151 Add codefix for S2933 (field should be readonly)

### LINKED ISSUE #151: Add codefix for S2933 (field should be readonly)

tvsonar:
Fixes #151

### DIFF
--- a/src/SonarLint/Rules/FieldShouldBeReadonlyCodeFixProvider.cs
+++ b/src/SonarLint/Rules/FieldShouldBeReadonlyCodeFixProvider.cs
@@ -1,0 +1,91 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


tvsonar:
From the SyntaxTree, we can get all the diagnostics, and then filter them by ID, check for common field declarations, and then we have all the issues belonging to a single declaration.

src/SonarLint/Rules/FieldShouldBeReadonlyCodeFixProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+                        Title),
+                    context.Diagnostics);
+            }
+        }
+
+        private static async Task<Document> AddReadonlyKeyword(Document document, FieldDeclarationSyntax fieldDeclaration,
+            CancellationToken cancellationToken)
+        {
+            var root = await document.GetSyntaxRootAsync(cancellationToken);
+            var newFieldDeclaration = fieldDeclaration.AddModifiers(
```
dbolkensteyn: @tvsonar this is quite cool, it adds it automatically in the usual order

src/SonarLint/Rules/FieldShouldBeReadonlyCodeFixProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+                if (fieldDeclaration == null)
+                {
+                    return;
+                }
+
+                context.RegisterCodeFix(
+                    CodeAction.Create(
+                        Title,
+                        c => AddReadonlyKeyword(context.Document, fieldDeclaration, c),
+                        Title),
```
dbolkensteyn: @tvsonar as discussed - let's remove this

dbolkensteyn:
LGTM - but the analyzer could IMO also suggest to make `private static` fields `readonly` - I've created a new issue for that: https://github.com/SonarSource/sonarlint-vs/issues/206

PR #184, opened 2015-08-11
Issue #78 - Redundant argument with default value - Analyzer, CodeFix

tvsonar:
Fixes #78

PR #189, opened 2015-08-11
Issue #144 Add codefix for S1905 (redundant casts)

### LINKED ISSUE #144: Add codefix for S1905 (redundant casts)

tvsonar:
Fixes #144

### DIFF
--- a/src/SonarLint/Rules/RedundantCastCodeFixProvider.cs
+++ b/src/SonarLint/Rules/RedundantCastCodeFixProvider.cs
@@ -1,0 +1,102 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeFixes;
+using System.Threading.Tasks;
+using System.Linq;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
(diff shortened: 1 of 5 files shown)


src/Tests/SonarLint.UnitTest/Rules/RedundantCastTest.cs
```diff
@@ -32,5 +32,16 @@ public void RedundantCast()
         {
             Verifier.VerifyAnalyzer(@"TestCases\RedundantCast.cs", new RedundantCast());
         }
+
+        [TestMethod]
+        [TestCategory("CodeFix")]
+        public void RedundantCast_CodeFix()
```
tvsonar: @dbolkensteyn This one fails at the moment, it needs the modification from #184

src/SonarLint/Rules/RedundantCastCodeFixProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        private static async Task<Document> RemoveStaticMemberCall(Document document, MemberAccessExpressionSyntax memberAccess,
+            CancellationToken cancellationToken)
+        {
+            var invocation = memberAccess.Parent as InvocationExpressionSyntax;
+            if (invocation == null ||
+                !invocation.ArgumentList.Arguments.Any())
+            {
+                return document;
```
dbolkensteyn: @tvsonar it seems to me like this check should be done before the code fix is suggested. It's misleading to offer a codefix, and to later not fix anything in the end

src/SonarLint/Rules/RedundantCastCodeFixProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+            return document.WithSyntaxRoot(newRoot);
+        }
+
+        private static async Task<Document> RemoveExtensionMethodCall(Document document, InvocationExpressionSyntax invocation,
+            CancellationToken cancellationToken)
+        {
+            var memberAccess = invocation.Expression as MemberAccessExpressionSyntax;
+            if (memberAccess == null)
+            {
+                return document;
```
dbolkensteyn: @tvsonar same thing here

src/SonarLint/Rules/RedundantCastCodeFixProvider.cs
dbolkensteyn: or perhaps this actually is deadcode - in which case this hypothesis is better replaced by a `Deubg.Assert()` or something that will help to fix the code

dbolkensteyn: it actually seems like deadcode - the analyzer won't report if this hypothesis isn't true

dbolkensteyn:
LGTM, but `x.OfType<int>().Cast<int>().ToList();` should be solvable all at once

dbolkensteyn:
Let's got ahead with the current code fix - I've created another issue to improve it later on https://github.com/SonarSource/sonarlint-vs/issues/208

src/SonarLint/Rules/RedundantCastCodeFixProvider.cs
tvsonar: @dbolkensteyn As you pointed out it is dead code, so I simply removed it.

PR #190, opened 2015-08-11
Fixing code fix verifier

tvsonar:


src/Tests/SonarLint.UnitTest/Verifier.cs
```diff
@@ -58,24 +59,48 @@ public static void VerifyAnalyzerInTest(string path, DiagnosticAnalyzer diagnost
...
+                Assert.AreNotEqual(0, maxAttempts);
 
-                    document = ApplyFix(document, actions.ElementAt(0));
-                    diagnostics = GetDiagnostics(document.Project.GetCompilationAsync().Result, diagnosticAnalyzer).ToList();
+                var previous = "";
+                var actual = document.GetSyntaxRootAsync().Result.GetText().ToString();
+                var codeFixExecutedAtLeastOnce = false;
 
-                    Assert.AreEqual(attempts - i - 1, diagnostics.Count);
+                while (previous != actual)
```
dbolkensteyn: @tvsonar looks like a `do ... while` could be more appropriate as you always want the first iteration to go through

src/Tests/SonarLint.UnitTest/Verifier.cs
```diff
@@ -58,24 +59,48 @@ public static void VerifyAnalyzerInTest(string path, DiagnosticAnalyzer diagnost
...
+
+                        codeFixProvider.RegisterCodeFixesAsync(context).Wait();
+
+                        var codeActionToExecute = actions.ElementAt(0);
+                        if (codeFixTitle != null)
+                        {
+                            codeActionToExecute = actions.SingleOrDefault(action => action.Title == codeFixTitle);
+                            if (codeActionToExecute == null)
+                            {
+                                diagnosticIndexToFix++;
```
dbolkensteyn: @tvsonar hm not sure to get this - will it move on to another codefix if it didn't find the requested one?

src/Tests/SonarLint.UnitTest/Verifier.cs
```diff
@@ -58,24 +59,48 @@ public static void VerifyAnalyzerInTest(string path, DiagnosticAnalyzer diagnost
...
+                var actual = document.GetSyntaxRootAsync().Result.GetText().ToString();
+                var codeFixExecutedAtLeastOnce = false;
 
-                    Assert.AreEqual(attempts - i - 1, diagnostics.Count);
+                while (previous != actual)
+                {
+                    previous = actual;
+                    var diagnosticIndexToFix = 0;
+
+                    while (diagnosticIndexToFix < maxAttempts)
```
dbolkensteyn: @tvsonar I think this should be extracted into another method - that will specifically fix a given diagnostic, with an optionally given code fix

src/Tests/SonarLint.UnitTest/Verifier.cs
dbolkensteyn: @tvsonar IMO if you want to go for a fixed point approach (as opposed as going once through each initial diagnostic), then you need to get the new diagnostics at each stage. Applying a code fix could result in new diagnostics being introduced. This however introduces the risk to potentially not terminate. Moreover, we just said today that we don't really like to have users to apply code fix in steps - they should be applicable all at once. So it seems an overkill.

Maybe it's safer for now to simply foreach on all diagnostics

tvsonar: @dbolkensteyn That's exactly what is happening here: 
* we get the diagnostics
* get the first, and check for fixes
* if the input fix can be applied, then apply it, and get the new list of diagnostics, and continue
* if the input fix cannot be applied, then go to the next diagnostic. 

to not get into infinite loops, we limit the search for a viable diagnostic with the max number. And also, if after trying to apply a fix we get the same code as the input, then we exit.

src/Tests/SonarLint.UnitTest/Verifier.cs
```diff
@@ -50,36 +51,58 @@ public static void VerifyAnalyzerInTest(string path, DiagnosticAnalyzer diagnost
...
+                    for (int diagnosticIndexToFix = 0; diagnosticIndexToFix < diagnostics.Count; diagnosticIndexToFix++)
+                    {
+                        var codeActionsForDiagnostic = GetCodeActionsForDiagnostic(codeFixProvider, document, diagnostics[diagnosticIndexToFix]);
 
-                    document = ApplyFix(document, actions.ElementAt(0));
-                    diagnostics = GetDiagnostics(document.Project.GetCompilationAsync().Result, diagnosticAnalyzer).ToList();
+                        CodeAction codeActionToExecute;
+                        if (!TryGetCodeActionToApply(codeFixTitle, codeActionsForDiagnostic, out codeActionToExecute))
+                        {
+                            continue;
```
dbolkensteyn: @tvsonar it seems a bit strange to always end the loop iteration with a `continue;` or a `break`;

I'd reverse the condition, get rid of the `continue;` and move the actual applying of the code within this `if`

dbolkensteyn:
LGTM

PR #191, opened 2015-08-12
Issue #143 Add codefix for S3005 (non static field with ThreadStaticAttribute)

### LINKED ISSUE #143: Add codefix for S3005 (non static field with ThreadStaticAttribute)

tvsonar:
Fixes #143

### DIFF
--- a/src/SonarLint/Rules/ThreadStaticNonStaticFieldCodeFixProvider.cs
+++ b/src/SonarLint/Rules/ThreadStaticNonStaticFieldCodeFixProvider.cs
@@ -1,0 +1,72 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System.Collections.Immutable;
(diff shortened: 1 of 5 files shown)


src/SonarLint/Rules/RedundantParenthesesCodeFixProvider.cs
```diff
@@ -43,7 +43,7 @@ public sealed override FixAllProvider GetFixAllProvider()
         {
```
dbolkensteyn: ahum this is completely unreleaded and should go into the redundant parentheses PR?

src/SonarLint/Rules/ThreadStaticNonStaticFieldCodeFixProvider.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                nodeToRemove = attributeList;
+            }
+
+            context.RegisterCodeFix(
+                CodeAction.Create(
+                    Title,
+                    c => RedundantParenthesesCodeFixProvider.RemoveUnusedCode(context.Document, nodeToRemove,
+                        SyntaxRemoveOptions.KeepNoTrivia, c),
+                    Title),
```
dbolkensteyn: @tvsonar as discussed - let's drop this

src/SonarLint/Rules/ThreadStaticNonStaticFieldCodeFixProvider.cs
```diff
@@ -0,0 +1,71 @@
...
+            var attributeList = nodeToRemove.Parent as AttributeListSyntax;
+            if (attributeList != null && attributeList.Attributes.Count == 1)
+            {
+                nodeToRemove = attributeList;
+            }
+
+            context.RegisterCodeFix(
+                CodeAction.Create(
+                    Title,
+                    c => RedundantParenthesesCodeFixProvider.RemoveUnusedCode(context.Document, nodeToRemove,
```
dbolkensteyn: @tvsonar I don't think it's worth to reuse the code of `RedundantParenthesesCodeFixProvider` here - it's more misleading than helping. Both analyzers are completely unrelated

dbolkensteyn:
LGTM

One small thing: On `[System.ThreadStatic, /*x*/ MTAThread /*y*/]` - we're losing the `/*x*/` comment - but that's minor

tvsonar:
@dbolkensteyn The comment is a trailing trivia of the comma token. I don't think it is worth the trouble to keep it, because we would need to filter out the whitespace trivias, and only keep the  comment ones. It is going to be unnecessarily complex. (We already have to branch based on single/multiple attributes in the attribute list.

PR #193, opened 2015-08-12
Issue #173 Fade out and add code fix to catch blocks with only a throw in them (S2737)

### LINKED ISSUE #173: Fade out and add code fix to catch blocks with only a throw in them (S2737)
```
try
{
 doSomething();
}
catch(Exception)
{
  throw;
}
```
is equivalent to 
```
doSomething();
```

tvsonar:
Fixes #173

### DIFF
--- a/src/SonarLint/Rules/CatchRethrowCodeFixProvider.cs
+++ b/src/SonarLint/Rules/CatchRethrowCodeFixProvider.cs
@@ -1,0 +1,92 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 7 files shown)


src/SonarLint/Rules/CatchRethrowCodeFixProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+                return;
+            }
+
+            if (tryStatement.Catches.Count > 1 ||
+                tryStatement.Finally != null)
+            {
+                context.RegisterCodeFix(
+                    CodeAction.Create(
+                        Title,
+                        c => RedundantParenthesesCodeFixProvider.RemoveUnusedCode(context.Document, syntaxNode,
```
dbolkensteyn: @tvsonar let's not use the `RedundantParenthesesCodeFixProvider` as discussed

dbolkensteyn:
LGTM

PR #198, opened 2015-08-13
Issue #69 Fix error location for S1125 and add ternary operator support

### LINKED ISSUE #69: Rule: Fix error location for S1125 and add ternary operator support
Replaces http://jira.sonarsource.com/browse/LINTVS-182

```
var x = condition? a : false;
x = condition? a : true;
x = condition? false : true;
```
are equivalent to
```
var x = condition && a;
x = condition || a;
x = !condition;
```

R# ID: *SimplifyConditionalTernaryExpression*

Implements http://jira.sonarsource.com/browse/RSPEC-1125

tvsonar:
Fixes #69

### DIFF
--- a/src/SonarLint/Rules/BooleanLiteralUnnecessary.cs
+++ b/src/SonarLint/Rules/BooleanLiteralUnnecessary.cs
@@ -38,12 +38,10 @@ namespace SonarLint.Rules
     public class BooleanLiteralUnnecessary : DiagnosticAnalyzer
     {
         internal const string DiagnosticId = "S1125";
-        internal const string Title = "Literal boolean values should not be used in condition expressions";
+        internal const string Title = "Boolean literals should not be redundant";
         internal const string Description =
-            "Remove literal boolean values from conditional expressions to improve readability. Anything that " +
-            "can be tested for equality with a boolean value must itself be a boolean value, and boolean values " +
-            "can be tested atomically.";
-        internal const string MessageFormat = "Remove the literal \"{0}\" boolean value.";
+            "Redundant Boolean literals should be removed from expressions to improve readability.";
+        internal const string MessageFormat = "Remove the unnecessary Boolean literal(s).";
         internal const string Category = "SonarQube";
         internal const Severity RuleSeverity = Severity.Minor;
         internal const bool IsActivatedByDefault = true;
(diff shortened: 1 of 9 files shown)


src/SonarLint/Rules/BooleanLiteralUnnecessary.cs
```diff
@@ -29,6 +29,7 @@
 using SonarLint.Common.Sqale;
 using SonarLint.Helpers;
 using Microsoft.CodeAnalysis.Text;
+using System;
```
dbolkensteyn: @tvsonar unused using introduced - there are 2 others of them in this file

dbolkensteyn:
The code fix is suggesting either `true` or `!true` in this case, whereas I actually expect it to only suggest `false`. In particular, it is very disturbing that `true` is generated, as it changes the functional meaning of the program.

```
false == true
```

On the following code, two issues are reported on the `false` boolean literal, and none on the `true`. One of them suggests to remove the `false`, and the other one the `true` - which is strange.

```
if (true || false) { ... }
``` 

The RSPEC is very specific to conditions, but in fact this rule applies to any boolean expression. I will update it to remove references to "conditions".

src/SonarLint/Rules/BooleanLiteralUnnecessaryCodeFixProvider.cs
```diff
@@ -0,0 +1,295 @@
...
+using Microsoft.CodeAnalysis.Formatting;
+using Microsoft.CodeAnalysis.Text;
+using System;
+
+namespace SonarLint.Rules
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public class BooleanLiteralUnnecessaryCodeFixProvider : CodeFixProvider
+    {
+        internal const string Title = "Remove the unnecessary Boolean literal(s)";
```
dbolkensteyn: @tvsonar i'd use a lower case `b` for boolean

tvsonar: @dbolkensteyn I think "Boolean" is correct: https://en.wikipedia.org/wiki/Boolean_data_type, but we could use "bool" instead.

dbolkensteyn:
I've made a couple of changes to the RSPEC.

Regarding the strange behaviors, my feeling is that you'll want to report a single issue on chains of redundant Boolean literals such as `false == true == true == false` or `true || false || ...` and avoid the strange behaviors.

src/SonarLint/Rules/BooleanLiteralUnnecessaryCodeFixProvider.cs
dbolkensteyn: I thought it was used as a noun here, not a datatype. But I'm fine with it, I've edited the RSPEC for consistency.

dbolkensteyn:
LGTM

PR #204, opened 2015-08-17
Issue #202 Add support to the Elvis operator in S2930 (Dispose of unused fields)

### LINKED ISSUE #202: Add support to the Elvis operator in S2930 (Dispose of unused fields)
I get a false positive SonarLint issue on the following code:
```
namespace Test
{
    class MyClass : IDisposable
    {
        private Mutex _mutex; // false positive S2930

        public MyClass()
        {
             _mutex = new Mutex(true, "MyAppInstance");
        }

        public void Dispose()
        {
           _mutex?.Dispose();
           _mutex = null;
        }
    }
}
```

This is a false positive because SonarLint doesn't recognise the null operator (?). 

If I rewrite the code in the `Dispose` method to `if (_mutex != null) _mutex.Dispose();` Then the SonarLint issue is gone

tvsonar:
Fixes #202

### DIFF
--- a/src/SonarLint/Rules/DisposableNotDisposed.cs
+++ b/src/SonarLint/Rules/DisposableNotDisposed.cs
@@ -97,6 +97,11 @@ namespace SonarLint.Rules
                     SyntaxKind.InvocationExpression);
 
                 analysisContext.RegisterSyntaxNodeAction(
+                    c => CollectDisposedSymbol((ConditionalAccessExpressionSyntax)c.Node, c.SemanticModel, disposeMethod,
+                    IsDisposableLocalSymbol, ref localsDisposed),
+                    SyntaxKind.ConditionalAccessExpression);
+
+                analysisContext.RegisterSyntaxNodeAction(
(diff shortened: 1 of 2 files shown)


src/SonarLint/Rules/DisposableNotDisposed.cs
```diff
@@ -139,6 +144,11 @@ public override void Initialize(AnalysisContext context)
                     DisposableMemberInNonDisposableClass.IsNonStaticNonPublicDisposableField, ref fieldsDisposed),
                     SyntaxKind.InvocationExpression);
 
+                analysisContext.RegisterSyntaxNodeAction(
```
dbolkensteyn: @tvsonar this code doesn't seem to be covered by tests

dbolkensteyn:
LGTM except for the unit test cases

src/SonarLint/Rules/DisposableNotDisposed.cs
tvsonar: @dbolkensteyn Ohh, you are right, I forgot to initialize `field7` in the test.

PR #215, opened 2015-08-19
Issue #153 Add codefix for S2934 (property assignment of a generic readonly field)

### LINKED ISSUE #153: Add codefix for S2934 (property assignment of a generic readonly field)

tvsonar:
Fixes #153

### DIFF
--- a/src/SonarLint/Rules/GenericReadonlyFieldPropertyAssignmentCodeFixProvider.cs
+++ b/src/SonarLint/Rules/GenericReadonlyFieldPropertyAssignmentCodeFixProvider.cs
@@ -1,0 +1,187 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeFixes;
+using System.Threading.Tasks;
+using System.Linq;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Formatting;
+using Microsoft.CodeAnalysis.CSharp;
+using SonarLint.Helpers;
+using System.Collections.Generic;
+using System;
+
+namespace SonarLint.Rules
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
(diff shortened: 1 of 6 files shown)


src/SonarLint/Rules/GenericReadonlyFieldPropertyAssignmentCodeFixProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Formatting;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace SonarLint.Rules
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public class GenericReadonlyFieldPropertyAssignmentCodeFixProvider : CodeFixProvider
+    {
+        internal const string TitleRemove = "Remove assignment";
```
dbolkensteyn: @tvsonar i wouldn't suggest this code fix - it is likely to break code

src/SonarLint/Rules/GenericReadonlyFieldPropertyAssignmentCodeFixProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+        internal const string TitleAddClassConstraint = "Add reference type constraint";
```
dbolkensteyn: @tvsonar this seems like the better one - but why isn't it being offered?

src/Tests/SonarLint.UnitTest/TestCases/GenericReadonlyFieldPropertyAssignment.AddConstraint.Fixed.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Tests.TestCases
```
duncanpMS: There's an obscure corner case with multiple generic arguments that seems to generate a false positive. I think it's too obscure to worry about, though.

    class P2<A, B> where A : class, IPoint where B : A
    {
        readonly A pointA;
        readonly B pointB;

        public P2(A a, B b)
        {
            this.pointA = a;
            this.pointB = b;
        }

        public void Add(int i)
        {
            pointA.X += i;
            pointB.X += i + i;   // diagnostic fires here
        }
    }
`

src/Tests/SonarLint.UnitTest/TestCases/GenericReadonlyFieldPropertyAssignment.AddConstraint.Fixed.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            this.point = point;
+            this.point.X = 1; // Compliant, we are in the constructor
+        }
+
+        public void MovePointVertically(int newX)
+        {
+            point.X = newX; //Noncompliant; if point is a struct, then nothing happened
+            point.X++; //Noncompliant; if point is a struct, then nothing happened
+            Console.WriteLine(point.X);
```
duncanpMS: Another corner case (against, not very common):
            int i = point.X = newX;
            point.X = i = newX;
The "remove" constraint strips out the whole line so the assignment to *i* is lost.

duncanpMS:
LGTM

src/Tests/SonarLint.UnitTest/TestCases/GenericReadonlyFieldPropertyAssignment.AddConstraint.Fixed.cs
tvsonar: Created #459 to cover this.

src/Tests/SonarLint.UnitTest/TestCases/GenericReadonlyFieldPropertyAssignment.AddConstraint.Fixed.cs
tvsonar: Fixed this by only proposing the fix if the assignment is directly in a statement. Created #460 for the proper assignment removal.

PR #219, opened 2015-08-19
Issue #218 Remove false positives from S2934 (Generic readonly field property assignment)

### LINKED ISSUE #218: Remove false positives from S2934 (Generic readonly field property assignment)
* In the constructor it is allowed to assign to these properties
* we are reporting on struct restricted generic type parameters as well.

tvsonar:
Fixes #218


dbolkensteyn:
LGTM

PR #225, opened 2015-08-20
Issue #148 Change metadata, and add codefix to S1940 (inverted boolean check)

### LINKED ISSUE #148: Change metadata, and add codefix to S1940 (inverted boolean check)

tvsonar:
Fixes #148

### DIFF
--- a/src/SonarLint/Rules/BooleanCheckInverted.cs
+++ b/src/SonarLint/Rules/BooleanCheckInverted.cs
@@ -44,8 +44,8 @@ namespace SonarLint.Rules
             "be made instead.";
         internal const string MessageFormat = "Use the opposite operator (\"{0}\") instead.";
         internal const string Category = "SonarQube";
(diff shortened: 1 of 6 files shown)


dbolkensteyn:
@tvsonar Is there a good reason not to enable this rule by default? I think it should be minor but enabled.

dbolkensteyn:
LGTM, but this code fix, by removing parentheses, can change the meaning of the program.

For example: `var a = !!(args.Length == 0)` will get fixed into `!args.Length != 0` which then doesn't compile.

With dynamics, it should be possible to get programs that still compile but whose meaning will change, as in:
```
dynamic foo = null;
var a = !(foo == 0) + 3;
```

that will become:
```
var a = foo != 0 + 3
```

which will be evaluated not as `(foo != 0) + 3` as originally intended, but as `foo != (0 + 3)`

I think this should be fixed as part of another issue though

PR #385, opened 2015-08-31
Issue #142 Add codefix for rule S2996 (ThreadStatic with initializer)

### LINKED ISSUE #142: Add codefix for rule S2996 (ThreadStatic with initializer)

tvsonar:
Fixes #142


dbolkensteyn:
I'm not convinced about this code fix action: IMO we should offer the code fix that properly initializes the field for all thread .

I've checked R# and it is not offering a codefix for this issue.

However, I'm ok with keeping this code fix, but only as a 2nd option.

dbolkensteyn:

PR #390, opened 2015-08-31
Issue #381 S1940 should not report on implementations of operator != …

### LINKED ISSUE #381: S1940 should not report on implementations of operator != overloads
Rule S1940 reports on the correct implementation of an != operator.  THis is the one case where this good rule do not apply.
image

Ref.: https://msdn.microsoft.com/en-US/library/ms173147(v=vs.80).aspx

tvsonar:
…overloads

Fixes #381


dbolkensteyn:
LGTM

PR #392, opened 2015-08-31
Issue #181 Printf-style format strings should not lead to unexpected …

### LINKED ISSUE #181: Rule: Printf-style format strings should not lead to unexpected behavior at runtime
http://jira.sonarsource.com/browse/RSPEC-2275

tvsonar:
…behavior at runtime

Fixes #181

### DIFF
--- a/src/SonarLint/Rules/StringFormatArgumentNumberMismatch.cs
+++ b/src/SonarLint/Rules/StringFormatArgumentNumberMismatch.cs
@@ -1,0 +1,179 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarLint.Common;
+using SonarLint.Common.Sqale;
+using SonarLint.Helpers;
+using Microsoft.CodeAnalysis.Text;
+using System.Linq;
+
+namespace SonarLint.Rules
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [SqaleConstantRemediation("10min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.InstructionReliability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags(Tag.Bug, Tag.Pitfall)]
+    public class StringFormatArgumentNumberMismatch : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S2275";
+        internal const string Title = "Format strings should be passed the correct number of arguments";
+        internal const string Description =
+            "Use fewer arguments than are expected in your format string, and you'll get an error at runtime. Use more arguments " +
+            "than are expected, and you probably won't get the output you expect. Either way, it's a bug.";
+        internal const string MessageFormat = "The passed arguments do not match the format string.";
+        internal const string Category = "SonarLint";
+        internal const Severity RuleSeverity = Severity.Critical;
+        internal const bool IsActivatedByDefault = true;
+
+        internal static readonly DiagnosticDescriptor Rule =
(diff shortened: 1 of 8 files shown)


src/SonarLint/Rules/StringFormatArgumentNumberMismatch.cs
```diff
@@ -0,0 +1,134 @@
...
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c =>
+                {
+                    var invocation = (InvocationExpressionSyntax)c.Node;
+                    var methodSymbol = c.SemanticModel.GetSymbolInfo(invocation).Symbol as IMethodSymbol;
+
+                    if (methodSymbol == null ||
+                        methodSymbol.ContainingType.SpecialType != SpecialType.System_String ||
+                        methodSymbol.Name != "Format" ||
+                        invocation.ArgumentList.Arguments.Count <= 1)
```
dbolkensteyn: @tvsonar should be == 0, `string.Format("{0}")` is actually missing an argument

dbolkensteyn:
As discussed:

1. `string.Format("{0}")` should report a warning.
2. `string.Format("{}")` should report a warning on the string literal stating that the literal is invalid
3. `string.Format("{0}", "a", "b")` should report just on the unused "b"

src/Tests/SonarLint.UnitTest/TestCases/StringFormatArgumentNumberMismatch.cs
```diff
@@ -0,0 +1,44 @@
...
+                11); // Noncompliant; too many arguments
+            s = string.Format("{0} {1} {2}", new[] { 1, 2 }); // Compliant, not recognized
+            s = string.Format("{0} {1} {2}", new object[] { 1, 2 }); // Compliant, not recognized
+            var pattern = "{0} {1} {2}";
+            s = string.Format(pattern, 1, 2); // Compliant, not recognized
+
+            const string pattern2 = "{0} {1} {2}";
+            s = string.Format(pattern2, // Noncompliant
+                1, 2);
+            s = string.Format(null, pattern2, 1, 2); // Noncompliant
```
duncanpMS: All of the permutations for invalid parameters with a constant format string are checked, but there's no check that the valid string doesn't produce a false positive i.e.
_s = string.Format(null, pattern2, 1, 2, 3); // compliant_

src/Tests/SonarLint.UnitTest/TestCases/StringFormatArgumentNumberMismatch.cs
```diff
@@ -0,0 +1,44 @@
...
+            s = string.Format(null, pattern2, 1, 2, 3,
+                4); // Noncompliant
+            s = string.Format(null, "{3}", 1); // Noncompliant
+            s = string.Format(null,
+                "{3}", // Noncompliant
+                1, 2, 3);
+            s = string.Format(null, "{2}", 1, 2, 3); // Not recognized
+            s = string.Format(null, "{}"); // not recognized
```
duncanpMS: This was one of the examples listed by Dinesh that should be handled. Is this rule focusing validly-formatted strings with parameter errors, rather than on detecting invalid format strings?

src/Tests/SonarLint.UnitTest/TestCases/StringFormatArgumentNumberMismatch.cs
```diff
@@ -0,0 +1,44 @@
...
+            s = string.Format(null, "{}", // not recognized
+                1,2,3);
+            s = string.Format(null, "{0}");  // Noncompliant
+            s = string.Format(null, "{0}{2}{1}"); // Noncompliant
+            s = string.Format(null, arg0: 1); // not recognized
+
+            s = string.Format(null, "{2000}"); // not recognized
+
+            s= string.Format(", 1 , 2, 3); // compliant
```
duncanpMS: Additional test cases:
* some valid cases to check for false positives e.g.
_s = string.Format("no format");
s = string.Format("{0}", 1); // valid 
s = string.Format("{1}", "{0}", 1, 2); // valid 

* recognised invalid cases with arrays e.g.
_int[] intArray = new int[] { };
s = string.Format("{0}", intArray, intArray);_

src/SonarLint/Rules/StringFormatArgumentNumberMismatch.cs
```diff
@@ -0,0 +1,175 @@
...
+                        }
+
+                        var spanStart = argument.SpanStart;
+                        var location = Location.Create(invocation.SyntaxTree, new TextSpan(
+                            spanStart,
+                            invocation.ArgumentList.Arguments.Last().Span.End - spanStart));
+                        c.ReportDiagnostic(Diagnostic.Create(Rule, location));
+                        return;
+                    }
+                },
```
duncanpMS: A few comments would have helped me understand the code faster e.g.
Line 86: _return :// can't statically check the override that supplies args in an array_
Line 94: _return: // can't check non-constant format strings_
Line 104: _// Check the format with the supplied number of args_
Line 107: _// Must be insufficient arguments_

duncanpMS:
LGTM

src/Tests/SonarLint.UnitTest/TestCases/StringFormatArgumentNumberMismatch.cs
tvsonar: @duncanpMS, We discussed it with @dbolkensteyn, and we couldn't come up with an easy way of detecting all the cases. One way of doing it would be to parse the constant string literal, and extract the placeholders. But that's a big task, as you'd need to handle curly brace escaping too. But that way the custom formatters could also be checked.

duncanpMS: @tvsonar I agree, just thought I'd ask since it was called out explicitly in the issue. The only easy way would be to hard-code this check as a special case, but that's not a great solution and it's not a big enough problem to warrant it.

PR #395, opened 2015-09-01
Issue #176 Methods with caller info attributes should not be invoked …

### LINKED ISSUE #176: Rule: Methods with caller info attributes should not be invoked with explicit arguments
http://jira.sonarsource.com/browse/RSPEC-3236

tvsonar:
…with explicit arguments

Fixes #176


dbolkensteyn:
LGTM

PR #396, opened 2015-09-01
Issue #174 Trivial properties should be auto-implemented

### LINKED ISSUE #174: Rule: Trivial properties should be auto-implemented
http://jira.sonarsource.com/browse/RSPEC-2292

tvsonar:
Fixes #174

### DIFF
--- a/src/SonarLint/Rules/PropertyToAutoProperty.cs
+++ b/src/SonarLint/Rules/PropertyToAutoProperty.cs
@@ -1,0 +1,189 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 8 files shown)


src/SonarLint/Rules/PropertyToAutoProperty.cs
```diff
@@ -0,0 +1,193 @@
...
+                    if (getter == null || setter == null)
+                    {
+                        return;
+                    }
+
+                    IFieldSymbol getterField = null;
+                    IFieldSymbol setterField = null;
+                    if (TryGetFieldFromGetter(getter, c.SemanticModel, out getterField) &&
+                        TryGetFieldFromSetter(setter, c.SemanticModel, out setterField) &&
+                        getterField != null &&
```
dbolkensteyn: @tvsonar IMO if the `Try` operation returned true, then the field should be set to a non-null value already

src/SonarLint/Rules/PropertyToAutoProperty.cs
```diff
@@ -0,0 +1,193 @@
...
+                return false;
+            }
+
+            var statement = setter.Body.Statements[0] as ExpressionStatementSyntax;
+            if (statement == null)
+            {
+                return false;
+            }
+
+            var assignment = statement.Expression as AssignmentExpressionSyntax;
```
dbolkensteyn: @tvsonar `field += value` will also pass this check, whereas we probably only what to detect `field = value`

src/SonarLint/Rules/PropertyToAutoProperty.cs
```diff
@@ -0,0 +1,193 @@
...
+
+            var identifier = memberAccess.Expression as IdentifierNameSyntax;
+            if (identifier == null)
+            {
+                field = null;
+                return false;
+            }
+
+            var type = semanticModel.GetSymbolInfo(identifier).Symbol as INamedTypeSymbol;
+            if (type == null ||
```
dbolkensteyn: @tvsonar this isn't covered by the tests, I guess `anotherInstance.field = value` would cover it

dbolkensteyn:
LGTM, but see comments

PR #397, opened 2015-09-02
AppVeyor: Use Visual Studio 2015 - Fixes #92

### LINKED ISSUE #92: force the usage of MSBuild 14 during the build
The size of the .vsix file is much larger when the project is being build with MSBuild 12 - so we should enforce the usage of MSBuild 14.

dbolkensteyn:
Fixes #92


tvsonar:
This has been merged into master.

PR #399, opened 2015-09-02
Issue #171 "Type.IsAssignableFrom()" should not be used to check obje…

### LINKED ISSUE #171: Rule: "Type.IsAssignableFrom()" should not be used to check object type
http://jira.sonarsource.com/browse/RSPEC-2219

**This RSPEC needs rewrite.**

R# rule ID: _UseMethodIsInstanceOfType_

There are certain cases when a type checking can be simplified, for example:
```
var expr = ...
Type t = ...
var b = t.IsAssignableFrom(expr.GetType());
```
is equivalent to
```
var expr = ...
Type t = ...
var y = t.IsInstanceOfType(expr); // this is simpler, has one method call less
```

furthermore, if we have the following:
```
var b = typeof(String).IsAssignableFrom(expr.GetType());
```
This is equivalent to both of the following:
```
var b = typeof(String).IsInstanceOfType(expr); //this is the same as above
b = expr is String; //preferred, we can do this because the type is known at compile time
```

tvsonar:
…ct type

Fixes #171

### DIFF
--- a/src/SonarLint/Rules.Description/S2219.html
+++ b/src/SonarLint/Rules.Description/S2219.html
@@ -1,0 +1,29 @@
+<p>
+    To check the type of an object there are at least three options:
+    <ul>
+        <li>The simplest and shortest one uses the <code>expr is SomeType</code> operator</li>
+        <li>The slightly longer <code>typeInstance.IsInstanceOfType(expr)</code></li>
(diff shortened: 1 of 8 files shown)


dbolkensteyn:
LGTM

dbolkensteyn:
LGTM

PR #400, opened 2015-09-02
Issue #382 S2123 reacts on fields and properties

### LINKED ISSUE #382: S2123 reacts on fields and properties
Rule 2123 is valuable for local variables, but should not trigger for fields and properties that are accessible later. 
The example below show one case:
image
(If counter is made into a property, we get the same result.)

If, however, the NextIndex method is made into an expression body, the warning disappears. 
image

Also, what is strange is that pressing the "Show potential fixes" doesn't show anything from this rule, but do show fixes from another rule (not Sonar).  I am not sure if this is Sonar or Roslyn, but should be looked at. 

The effect of pressing that is:
image

tvsonar:
Fixes #382


dbolkensteyn:
LGTM

PR #401, opened 2015-09-02
Issue #230 Codefix for S3240 can convert to code that has compiler er…

### LINKED ISSUE #230: Codefix for S3240 can convert to code that has compiler errors
```
Name n;
if (condition)
{
  n = new VerySpecialName();
}
else
{
  n = new SpecialName();
}
```
will be converted to 
```
Name n;
n = condition ? new VerySpecialName() : new SpecialName();
```
but because the `whenTrue` is of `VerySpecialName`, the `whenFalse` should also be of the same type.

tvsonar:
…rors

Fixes #230

### DIFF
--- a/src/SonarLint/Rules/ConditionalSimplificationCodeFixProvider.cs
+++ b/src/SonarLint/Rules/ConditionalSimplificationCodeFixProvider.cs
@@ -29,6 +29,7 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Formatting;
 using Microsoft.CodeAnalysis.CSharp;
 using SonarLint.Helpers;
+using System;
 
 namespace SonarLint.Rules
 {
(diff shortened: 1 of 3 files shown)


src/SonarLint/Rules/ConditionalSimplificationCodeFixProvider.cs
```diff
@@ -247,6 +309,13 @@ public sealed override FixAllProvider GetFixAllProvider()
 
             return methodCall1.WithArgumentList(newArgumentList);
         }
+
+        private static bool TypesMatch(ExpressionSyntax expr1, ExpressionSyntax expr2, SemanticModel semanticModel)
+        {
+            var type1 = semanticModel.GetTypeInfo(expr1).Type;
+            var type2 = semanticModel.GetTypeInfo(expr2).Type;
+            return type1 != null && type1.Equals(type2);
```
dbolkensteyn: the rule is that there should be an implicit conversion between the two types, which is weaker than strict equality

dbolkensteyn:
I think the issue is actually with the analyzer, not with the code fix. I've updated the issue accordingly.

src/Tests/SonarLint.UnitTest/TestCases/ConditionalSimplification.cs
```diff
@@ -84,6 +84,52 @@ int Test(object a, object b, object y, bool condition)
                 return 2;
             else
                 return 3;
+
```
duncanpMS: I've opened this version of the unfixed file in VS with the updated code provider and tried to apply the fix e.g. on line 54. I'm getting a gold bar with the message _'ConditionalSimplificationCodeFixProvider' encountered an error and has been disabled. Enable Enable and ignore future errors_.

src/SonarLint/Rules/ConditionalSimplificationCodeFixProvider.cs
```diff
@@ -54,60 +55,102 @@ public sealed override FixAllProvider GetFixAllProvider()
...
+                                return Task.FromResult(context.Document.WithSyntaxRoot(newRoot));
+                            }),
+                        context.Diagnostics);
+                }
             }
         }
 
+        private static T RemoveAnnotation<T>(T node, SyntaxAnnotation annotation) where T: SyntaxNode
+        {
+            var annotated = node.GetAnnotatedNodes(annotation).First();
```
duncanpMS: This call can fail for some of the examples in the test file.

src/Tests/SonarLint.UnitTest/TestCases/ConditionalSimplification.cs
```diff
@@ -84,6 +84,52 @@ int Test(object a, object b, object y, bool condition)
...
+            Base elem;
+            if (condition) //Noncompliant
+            {
+                elem = new A();
+            }
+            else
+            {
+                elem = new B();
+            }
```
duncanpMS: A couple of additional test cases:
* if /else without { }
* comments in the if/else block

duncanpMS:
LGTM (apart from the exception being thrown :-))

src/SonarLint/Rules/ConditionalSimplificationCodeFixProvider.cs
tvsonar: @duncanpMS Good catch. It seems to me that this only happens in VS, and not during test execution. And the reason seems to be that VS is running the CodeAction multiple times, and for the second time, the annotation is already removed. I changed it to `FirstOrDefault`, and a null check.

PR #403, opened 2015-09-02
Enable SQ analysis on PR

dbolkensteyn:


dbolkensteyn:
well well FxCop's execution is leading to the overall failure on MSBuild on SonarLint. VS ... independently of the SonarQube integration:

```
msbuild /p:RunCodeAnalysis=true
echo %errorlevel%
```

That's not exactly good news nor what I expected

dbolkensteyn:
@tvsonar Can you have a look at why FxCop is failing?

tvsonar:
@dbolkensteyn can you exclude the VISX project from SQ analysis? (as discussed we couldn't find any better short term solution)

dbolkensteyn:
This solution isn't good enough - Dory is now available online and we should be able to use it.

PR #407, opened 2015-09-04
Issue #383 Improve the metadata of S2760 not to show it as a bug

### LINKED ISSUE #383: Improve the metadata of S2760 not to show it as a bug
Rule S2760 seems to miss that intermediate methods may affect a condition check.
The rule says that:
image
But the code in questions is like:
image
Both Position and Length are properties of the file object, which may have been changed by the call in (1), so the test in (2) is not "sequential".

tvsonar:
Fixes #383


duncanpMS:
LGTM

Issue #211, opened 2015-08-19
S1172 should not report on event handlers


jessehouwing:
S1172 should not flag unused members if they are part of an interface method or when they are  passed to a delegate/eventhandler.

Classic example of this is the (sender, event) passed to an Event handling function. Many people don't use the sender and often not even the event, this is not really a problem.

Alternative syntax, such as += (s,e) => MethodWithoutParameters(); works, but had the problem that unsubscribing from the event requires us to store the delegate so that we can use it in the -= call later on.

tvsonar:
@jessehouwing I see your point with the event handlers. I've added a [comment](http://stackoverflow.com/questions/32078974/sonarlint-rule-s1172-unused-method-parameters-should-be-removed-and-eventhandl#comment52071559_32078974) to the [SO](http://stackoverflow.com/questions/32078974/sonarlint-rule-s1172-unused-method-parameters-should-be-removed-and-eventhandl) question where the discussion started.

Here is the relevant part: My feeling is that if you don't need the `sender` and the `EventArgs` at all, then you could use a custom delegate which doesn't have those. But this only works if you control the type of the event. Do you think this goes against event handling best practices?

Additionally: Can you provide some code samples for the false positives with interface implementations? This rule should not trigger an issue, if the method is `abstract`, `virtual`, `override`, or if it is an interface implementation.

njannink:
Its a design guideline that event delegates always should be (derived) of the type EventHandler. So that means you will always get the sender and EventArgs. [MSDN](https://msdn.microsoft.com/en-us/library/vstudio/ms229011(v=vs.100).aspx)

So even tho the subscribers don't use those parameters all events in the .Net framework are build using this paradigm and so by my opinion SonarLint should be able to handle this. 

I think Resharper correctly recognises event handling methods and ignores the unused parameters of these.

tvsonar:
Thanks for the MSDN link. 

It seems to me that Resharper does at least two things differently
* doesn't report on any non-`private` method
* for `private` methods it is checking if the method is assigned to a anything (local variable, field, event, ...)

These reduce the number of false positives, but also introduce a lot of false negatives. Instead, I would probably go with identifying event handler method signatures.

jessehouwing:
Thanks for chiming in with the design guidelines. That's what I'd have done. I'll have to look up the case where it triggered on an interface variable. I've refactored that codebase quite a bit over the last week.

jessehouwing:
[...]

PR #410, opened 2015-09-11
Issue #211 In S1172 ignore methods that are passed to delegates ...

andwi:
…or event handlers

A project I work on has very large number of warnings due to this issue. So I thought this is a great time to start learning about Roslyn code analyzers and try to fix it myself.

I think I've got a solution that works. I have verified against a modified version of the existing MethodParameterUnused test. But the ItSources_Match_Expected_All test fails and I have no idea how to geet this test green. I would appreciate some help on this.

### DIFF
--- a/src/SonarLint/Rules.Description/S1172.html
+++ b/src/SonarLint/Rules.Description/S1172.html
@@ -20,7 +20,7 @@ void DoSomething(int a)
 
 <h2>Exceptions</h2>
 <p>
-    <code>virtual</code>, <code>override</code> methods and interface implementations are ignored.
+    <code>virtual</code>, <code>override</code> methods, interface implementations and methods that are passed to a delegate or eventhandler.
 </p>
 <pre>
 override void DoSomething(int a, int b) // no issue reported on b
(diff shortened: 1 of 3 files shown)


tvsonar:
@andwi Thanks for the PR. To fix the ruling test, you need to update the corresponding expected file: https://github.com/SonarSource/sonarlint-vs/blob/master/src/Tests/SonarLint.RulingTest/Expected/S1172.json
When you run the `ItSources_Match_Expected_All` test, it generates the actual files in your `sonarlint-vs\src\Tests\SonarLint.RulingTest\bin\Debug\Actual\` folder. Here you'll find the S1172.json file which needs to be compared to the expected one. Check some/all of the differences to see if you only removed cases that you wanted. If that is the case, then overwrite the expected file and the test should be green.

andwi:
@tvsonar Thanks for the explanation. I found some issues in the code by going over the expected file. Will soon update the PR.

dbolkensteyn:
Any news on this @andwi ? It's been a while, I'm going to have to close this PR if the issue isn't resolved

dbolkensteyn:
@tvsonar Is this something that, from a functional point of view, you'd like to do? If so, could you please create an issue?

PR #415, opened 2015-09-18
Issue #414 - S3052 Decimals initialized to non-zero reported as defau…

### LINKED ISSUE #414: S3052 Decimals initialized to non-zero reported as default value (false positive)
S3052 generates a warning for code such as:

        private const decimal BandFractionalWidth = .2m;

Warning	S3052	Remove this initialization to "BandFractionalWidth", the compiler will do that for you.

tvsonar:
…lt value (false positive)

Fixes #414

### DIFF
--- a/src/Tests/SonarLint.UnitTest/TestCases/FieldInitializedToDefault.cs
+++ b/src/Tests/SonarLint.UnitTest/TestCases/FieldInitializedToDefault.cs
@@ -8,19 +8,19 @@ namespace Tests.Diagnostics
 
     class FieldInitializedToDefault<T>
     {
-        public int field = 0; // Noncompliant
-        public float fieldF = 0.0; // Noncompliant
-        public char fieldC = '\0'; // Noncompliant
-        public long fieldL = 0L; // Noncompliant
-        public long fieldL2 = default(long); // Noncompliant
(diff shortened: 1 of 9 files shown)


src/Tests/SonarLint.UnitTest/TestCases/FieldInitializedToDefault.cs
```diff
@@ -20,5 +20,11 @@ class FieldInitializedToDefault<T>
 
         public T gen = default(T); // Noncompliant
         public Dummy stru = default(Dummy); // Noncompliant
+
+        public double fieldD1 = 0; // Noncompliant
+        public double fieldD2 = 0.0; // Noncompliant
+        public double fieldD3 = .0; // Noncompliant
+        public decimal fieldD4 = .0m; // Noncompliant
+        public decimal fieldD5 = .2m;
```
duncanpMS: There are a number of types that aren't tested (byte, sbyte, ushort, uint, ulong). The new ExpressionNumericConverter is complicated to have its own set of tests, but the next best thing would be to make sure all of the types are covered here (including values with multipliers).

duncanpMS:
LGTM

PR #416, opened 2015-09-18
Issue #150 Handle direct inequality in S1244 (floating point equality…

### LINKED ISSUE #150: Add codefix S1244 (floating point equality check)

tvsonar:
… check) and add codefix

Fixes #150


src/Tests/SonarLint.UnitTest/TestCases/EqualityOnFloatingPoint.cs
```diff
@@ -16,6 +16,9 @@ void test(float f, Double d)
             if (f == 3.14F) //Noncompliant
             {
             }
+            if (f != 3.14F) //Noncompliant
```
duncanpMS: The short description for the diagnostic says "Use <= or >= instead of == or !=", but the code fix uses a range check which is slightly confusing. Could the diagnostic message refer to a range check too, or would that make the short message too long?

src/Tests/SonarLint.UnitTest/TestCases/EqualityOnFloatingPoint.Fixed.cs
```diff
@@ -0,0 +1,50 @@
...
+            var i = 3;
+            if (i <= 3 && i >= 3)
+            {
+            }
+
+            if (i < 4 || i > 4)
+            {
+            }
+
+            if (Math.Abs(f - 3.146) > double.Epsilon) // Noncompliant indirect inequality test
```
duncanpMS: [MSDN](https://msdn.microsoft.com/en-us/library/system.double(v=vs.110).aspx#Equality) suggests not using single/double.Epsilon for equality checks ("_except for values that are zero, we do not recommend its use in tests for equality_"). Will this code fix be appropriate in most user scenarios, or does the solution need to be more context-specific (i.e. using an appropriate epsilon value for the particular scenario)?

src/SonarLint/Rules/EqualityOnFloatingPointCodeFixProvider.cs
```diff
@@ -0,0 +1,166 @@
...
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var diagnostic = context.Diagnostics.First();
+            var diagnosticSpan = diagnostic.Location.SourceSpan;
+            var syntaxNode = root.FindNode(diagnosticSpan);
+            var semanticModel = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+
+            var isEquality = diagnostic.Properties[EqualityOnFloatingPoint.OperatorEqualsKey] == EqualityOnFloatingPoint.Equality;
+
+            var binary = (BinaryExpressionSyntax)syntaxNode;
```
duncanpMS: Just an observation: the code fixes assume that the diagnostic has been created by the expected analyser, so it knows what type of syntax node has been selected. However, the diagnostic id is just a simple string and since these are generally short there isn't much protection against id collisions between different analysers (someone else might use the convention "S" + number). It might be worth considering a more defensive approach i.e. a soft rather than a hard cast on the syntax node.

src/SonarLint/Rules/EqualityOnFloatingPointCodeFixProvider.cs
```diff
@@ -0,0 +1,166 @@
...
+                            return Task.FromResult(context.Document.WithSyntaxRoot(newRoot));
+                        }),
+                    context.Diagnostics);
+            }
+        }
+
+        private static SyntaxNode ChangeIndirectEqualityToMathAbs(SyntaxNode root, BinaryExpressionSyntax binary,
+            SemanticModel semanticModel, TextSpan diagnosticSpan, bool isEquality)
+        {
+            var comparison = EqualityOnFloatingPoint.TryGetBinaryExpression(binary.Left);
```
duncanpMS: You're calling a _TryGet_ method but not checking for null on return. Are you just calling the _TryGet_ because it's a convenient method that already existed, and in this context it should never return null? If so, it might be worth adding an assertion to document/check this.

duncanpMS:
The changes to the analyser and associated tests look good.
The code fix and tests look good; the only issue is about the appropriateness of using Epsilon in the comparison.
Manually tested.

src/SonarLint/Rules/EqualityOnFloatingPointCodeFixProvider.cs
tvsonar: Good point, created a separate issue for it: #453

src/SonarLint/Rules/EqualityOnFloatingPointCodeFixProvider.cs
tvsonar: Yes, it should never return null.

tvsonar:
As we discussed, the epsilon should change according to the context. So we might not go through with the code fix at all.

dbolkensteyn:
I agree with @duncanpMS (and the MSDN) - we shouldn't offer a code fix for this issue @tvsonar . Using `Double.Epsilon` isn't actually much better than `==` or `!=`

PR #422, opened 2015-09-21
Issue #374 "params" should be used on overrides

### LINKED ISSUE #374: "params" should be used on overrides
```
class Base
{
  public virtual void Method(params int[] numbers)
  {
    ...
  }
}
class Derived : Base
{
  public override void Method(int[] numbers) // Noncompliant, the params is missing.
  {
    ...
  }
}
```
base class has overridden/removed the `params` behavior. Now must pass explicit array.
major
suspicious

R# ID: BaseMemberHasParams

https://jira.sonarsource.com/browse/RSPEC-3262

tvsonar:
Fixes #374

### DIFF
--- a/src/Tests/SonarLint.UnitTest/TestCases/MethodOverrideNoParams.Fixed.cs
+++ b/src/Tests/SonarLint.UnitTest/TestCases/MethodOverrideNoParams.Fixed.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Tests.Diagnostics
+{
+    abstract class Base
+    {
+        public virtual void Method(params int[] numbers)
+        {
+        }
+        public virtual void Method(string s,
+            params int[] numbers)
+        {
+        }
+        public abstract void Method(string s, string s1,
+            params int[] numbers);
+    }
+    abstract class Derived : Base
+    {
+        public override void Method(params int[] numbers) // Noncompliant, the params is missing.
+        {
+        }
+        public override void Method(string s,
+            params int[] numbers) // Noncompliant
+        {
+        }
+        public override void Method(string s, string s1,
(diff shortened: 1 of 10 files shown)


src/Tests/SonarLint.UnitTest/TestCases/MethodOverrideNoParams.cs
```diff
@@ -0,0 +1,25 @@
...
+        }
+    }
+
+    class Derived2 : Base
+    {
+        public override void Method(params int[] numbers)
+        {
+        }
+    }
+}
```
duncanpMS: Potential additional test case: abstract method

src/Tests/SonarLint.UnitTest/TestCases/MethodOverrideNoParams.Fixed.cs
```diff
@@ -0,0 +1,25 @@
...
+    class Derived : Base
+    {
+        public override void Method(params int[] numbers) // Noncompliant, the params is missing.
+        {
+        }
+    }
+
+    class Derived2 : Base
+    {
+        public override void Method(params int[] numbers)
```
duncanpMS: If the arguments are indented on a separate line from the method name, the indentation is lost after applying the fix.
It looks a bit odd, but I guess that is how the editor thinks the doc should be edited (the formatting doesn't change if you select the whole doc and select Edit, Advanced, Format Document.

duncanpMS:
LGTM

src/Tests/SonarLint.UnitTest/TestCases/MethodOverrideNoParams.Fixed.cs
tvsonar: Thanks, I fixed this by adding back the trivia from the old parameter to the new one.

PR #423, opened 2015-09-22
Issue #367 Gratuitious modifiers should be removed

### LINKED ISSUE #367: Gratuitious modifiers should be removed
In C# a single class can be defined at multiple locations if the class has `partial` modifier. This helps tools to generate some part of the class's functionality. But if a class is defined as `partial`, but is located only in one place, then the `partial` keyword is redundant.
Noncompliant:
```
partial class X
{
}
```
Compliant:
```
class X
{
}
```
or:
File1:
```
partial class X
{

}
```
File2:
```
partial class X
{

}
```

R# ID: PartialTypeWithSinglePart

https://jira.sonarsource.com/browse/RSPEC-2333

tvsonar:
Fixes #367

### DIFF
--- a/src/SonarLint/Rules.Description/S2333.html
+++ b/src/SonarLint/Rules.Description/S2333.html
@@ -1,0 +1,30 @@
+<p>
+    Unnecessary keywords simply clutter the code and should be removed. Specifically:
+    <ul>
+        <li><code>virtual</code> on members of classes that are never overridden</li>
+        <li><code>partial</code> on type declarations that are completely defined in one place</li>
+        <li><code>sealed</code> on members of <code>sealed</code> classes</li>
+    </ul>
+</p>
+
+<h2>Noncompliant Code Example</h2>
+<pre>
+internal partial class MyClass // Noncompliant;
+{
+  public virtual void DoSomething()  // Noncompliant; virtual is gratuitous
+  {
+    // ...
+  }
+}
+</pre>
+
+<h2>Compliant Solution</h2>
+<pre>
+internal class MyClass
+{
+  public void DoSomething()
+  {
+    // ...
+  }
+}
+</pre>
--- a/src/SonarLint/Rules/RedundantModifier.cs
+++ b/src/SonarLint/Rules/RedundantModifier.cs
@@ -1,0 +1,220 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarLint.Common;
+using SonarLint.Common.Sqale;
+using SonarLint.Helpers;
+
+namespace SonarLint.Rules
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [SqaleConstantRemediation("2min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.Understandability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags(Tag.Unused)]
+    public class RedundantModifier : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S2333";
+        internal const string Title = "Redundant modifiers should be removed";
+        internal const string Description =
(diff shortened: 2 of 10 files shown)


src/SonarLint/Rules/RedundantModifier.cs
```diff
@@ -0,0 +1,186 @@
...
+                            return;
+                        }
+
+                        if (!classWithBaseType.ContainsKey(classSymbol))
+                        {
+                            classWithBaseType = classWithBaseType.Add(classSymbol, classSymbol.BaseType);
+                        }
+                    }, SymbolKind.NamedType);
+
+                    analysisContext.RegisterCompilationEndAction(c =>
```
tvsonar: @dbolkensteyn  This action is not triggered in VS, but it is in the test. I couldn't yet figure out why.

tvsonar: I fixed it by changing the severity in VS. Created a separate issue to track this problem: Issue #448

src/SonarLint/Rules.Description/S2333.html
```diff
@@ -0,0 +1,30 @@
+<p>
+    Unneeded keywords simply clutter the code and should be removed. Specifically:
+    <ul>
+        <li><code>virtual</code> on members of classes that are never overridden</li>
```
duncanpMS: "on members of non-public classes..." might be a better description.

src/SonarLint/Rules/RedundantModifier.cs
```diff
@@ -0,0 +1,191 @@
...
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [SqaleConstantRemediation("2min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.Understandability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags(Tag.Unused)]
+    public class RedundantModifier : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S2333";
+        internal const string Title = "Gratuitious modifiers should be removed";
```
duncanpMS: Terminological consistency: this appears to be the only rule that uses "gratuitous" instead of "redundant" (and the code fix for this rule uses "redundant" too).

src/SonarLint/Rules/RedundantModifier.cs
```diff
@@ -0,0 +1,191 @@
...
+        internal const string Description =
+            "Unneeded keywords simply clutter the code and should be removed.";
```
duncanpMS: "Unnecessary keywords" reads more naturally.

src/SonarLint/Rules.Description/S2333.html
```diff
@@ -0,0 +1,30 @@
...
+        <li><code>partial</code> on classes that are completely defined in one place</li>
```
duncanpMS: FYI interfaces and structs can also be partial (although I can't remember having seen either in practice).

src/SonarLint/Rules/RedundantModifierCodeFixProvider.cs
```diff
@@ -0,0 +1,67 @@
...
+            return WellKnownFixAllProviders.BatchFixer;
+        }
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+            var diagnostic = context.Diagnostics.First();
+            var diagnosticSpan = diagnostic.Location.SourceSpan;
+            var token = root.FindToken(diagnosticSpan.Start);
```
duncanpMS: You could assert that the token is one of the expected types.

src/Tests/SonarLint.UnitTest/TestCases/RedundantModifier.cs
```diff
@@ -0,0 +1,26 @@
...
+    internal class Override : Partial2Part
+    {
+        public override void MyOverridenMethod() { }
+    }
+    sealed class SealedClass : Partial2Part
+    {
+        public virtual void MyNotOverridenMethod() { } //Noncompliant, reported by VS as well, but we need it for the codefix
+        public override sealed void MyOverridenMethod() { } //Noncompliant
+    }
+}
```
duncanpMS: A couple of additional test cases:
* virtual properties (currently no coverage)
* virtual on method of public / sealed classes (checking for false positives)

src/SonarLint/Rules/RedundantModifier.cs
```diff
@@ -0,0 +1,191 @@
...
+                analysisContext =>
+                {
+                    var classWithBaseType = ImmutableDictionary<INamedTypeSymbol, INamedTypeSymbol>.Empty;
+
+                    analysisContext.RegisterSymbolAction(
+                        c =>
+                        {
+                            var classSymbol = c.Symbol as INamedTypeSymbol;
+                            if (classSymbol == null ||
+                                classSymbol.TypeKind != TypeKind.Class)
```
duncanpMS: Potential minor optimisation: could public/sealed etc classes be skipped at this point too?

duncanpMS:
LGTM

src/Tests/SonarLint.UnitTest/TestCases/RedundantModifier.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        public virtual void MyOverridenMethod() { }
+    }
+    internal class Override : Partial2Part
+    {
+        public override void MyOverridenMethod() { }
+    }
+    sealed class SealedClass : Partial2Part
+    {
+        public virtual void MyNotOverridenMethod() { } //Noncompliant, reported by VS as well, but we need it for the codefix
```
duncanpMS: Could you just provide a code fix for the VS diagnostic, rather than creating a duplicate diagnostic for this case?

src/SonarLint/Rules.Description/S2333.html
tvsonar: Yes, I also thought about `InternalsVisibleTo`, but I don't think it's worth the time investment.

src/SonarLint/Rules.Description/S2333.html
tvsonar: Thanks, I changed the description and the implementation accordingly.

PR #425, opened 2015-09-25
Issue #388 S1118 should not report issues if the class has members th…

### LINKED ISSUE #388: Improve error message of S1118, and it should not report issues if the class has members that use an instance of a class
Error message: say explicitely that the constructor should be hidden by making it protected or private.

Like operator overloads, properties with the type of the class. These require the class to be instantiated, so the class can't be marked static.
Also, in
```
public class X
{
  public static void M(X x) {}
}
```
parameter `x` needs to be instantiated somewhere, so based on the visibility of the class this might be outside the class, which means that we can even propose a private constructor. So it is safer to just not report on these methods.

tvsonar:
…at use an instance of a class

Fixes #388

### DIFF
--- a/src/SonarLint/Rules/ClassWithOnlyStaticMember.cs
+++ b/src/SonarLint/Rules/ClassWithOnlyStaticMember.cs
@@ -26,6 +26,7 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using SonarLint.Common;
 using SonarLint.Common.Sqale;
 using SonarLint.Helpers;
+using System.Collections.Generic;
 
 namespace SonarLint.Rules
 {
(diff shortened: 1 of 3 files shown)


src/SonarLint/Rules/ClassWithOnlyStaticMember.cs
```diff
@@ -38,19 +39,15 @@ public class ClassWithOnlyStaticMember : DiagnosticAnalyzer
...
         internal const string Title = "Utility classes should not have public constructors";
-
         internal const string Description =
             "Utility classes, which are collections of \"static\" members, are not meant to be instantiated. Even \"abstract\" " +
             "utility classes, which can be extended, should not have \"public\" constructors. C# adds an implicit public " +
             "constructor to every class which does not explicitly define at least one constructor. Hence, at least one " +
             "\"protected\" constructor should be defined if you wish to subclass this utility class. Or the \"static\" keyword " +
             "should be added to the class declaration to prevent subclassing.";
-
         internal const string MessageFormatConstructor = "Hide this public constructor.";
```
duncanpMS: Not changed in this commit, but it's not immediately obvious what "hide" means in this context.
Suggestions: "Hide this constructor by making it protected" or "Change the constructor visibility to 'protected'"

src/Tests/SonarLint.UnitTest/TestCases/ClassWithOnlyStaticMember.cs
```diff
@@ -88,4 +88,22 @@ public static string Concatenate(string s1, string s2)
...
+    public class StringUtils5
+    {
+        public static void Concatenate(StringUtils5 p)
+        {
+        }
+    }
+    public class StringUtils6
+    {
+        public static StringUtils6 Prop { get; set; }
+    }
```
duncanpMS: Missing test case: field.

duncanpMS:
LGTM

PR #430, opened 2015-09-25
Issue #156 Add codefix for S2328 (GetHashCode uses mutable field)

### LINKED ISSUE #156: Add codefix for S2328 (GetHashCode uses mutable field)

tvsonar:
Fixes #156

### DIFF
--- a/src/Tests/SonarLint.UnitTest/TestCases/GetHashCodeMutable.Fixed.cs
+++ b/src/Tests/SonarLint.UnitTest/TestCases/GetHashCodeMutable.Fixed.cs
@@ -6,15 +6,15 @@ namespace Tests.Diagnostics
 {
     public class AnyOther
     {
-        public int Field;
+        public readonly int Field;
     }
 
     public class GetHashCodeMutable : AnyOther
     {
         public readonly DateTime birthday;
         public const int Zero = 0;
-        public int age;
-        public string name;
+        public readonly int age;
+        public readonly string name;
         int foo, bar;
 
         public GetHashCodeMutable()
--- a/src/SonarLint/Rules/GetHashCodeMutable.cs
+++ b/src/SonarLint/Rules/GetHashCodeMutable.cs
@@ -45,7 +45,7 @@ namespace SonarLint.Rules
             "the object is stored, the object immediately becomes mis-filed in the " +
             "\"Hashtable\". Any subsequent test to see if the object is in the \"Hashtable\" " +
             "will return a false negative.";
-        internal const string MessageFormat = "Remove this use of \"{0}\".";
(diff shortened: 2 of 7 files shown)


src/Tests/SonarLint.UnitTest/TestCases/GetHashCodeMutable.Fixed.cs
```diff
@@ -0,0 +1,40 @@
...
+            return hash;
+        }
+        public int SomeMethod()
+        {
+            int hash = Zero;
+            hash += this.age.GetHashCode();
+            return hash;
+        }
+    }
+}
```
duncanpMS: Additional test case:
* int foo, bar; // can't make readonly in this case

src/Tests/SonarLint.UnitTest/TestCases/GetHashCodeMutable.Fixed.cs
```diff
@@ -0,0 +1,40 @@
...
+            hash += age.GetHashCode(); // Noncompliant
+            hash += this.name.GetHashCode(); // Noncompliant
+            hash += this.birthday.GetHashCode();
+            hash += Field; // Noncompliant
+            return hash;
+        }
+        public int SomeMethod()
+        {
+            int hash = Zero;
+            hash += this.age.GetHashCode();
```
duncanpMS: Adding readonly will make the code uncompilable if the variable is assigned to other than in the constructor. However, that's probably ok; the user can preview the change, and can easily undo the change.

duncanpMS:
LGTM

src/Tests/SonarLint.UnitTest/TestCases/GetHashCodeMutable.Fixed.cs
```diff
@@ -0,0 +1,40 @@
...
+    public class GetHashCodeMutable : AnyOther
+    {
+        public readonly DateTime birthday;
+        public const int Zero = 0;
+        public readonly int age;
+        public readonly string name;
+
+        public GetHashCodeMutable()
+        {
+            any = new AnyOther();
```
duncanpMS: *any* isn't declared.

src/Tests/SonarLint.UnitTest/TestCases/GetHashCodeMutable.Fixed.cs
tvsonar: @duncanpMS This is expected. We never want to produce code that is semantically wrong, but compiles. Also, it would be very-very nice to produce code that is semantically correct and compiles, but at the moment for this rule this is not always the case. There is a rule to check if a field can be marked readonly, whenever we decide how to handle rules that depend on each other, this issue will be fixed. I've created #458 to note this issue.

PR #431, opened 2015-09-25
Issue #408 Remove Microsoft.CodeAnalysis.Workspaces.Desktop.dll from …

### LINKED ISSUE #408: Remove Microsoft.CodeAnalysis.Workspaces.Desktop.dll from the Runner
Microsoft.CodeAnalysis.Workspaces.Desktop.dll is needed for the codefixes, but the runner doesn't need them, and doesn't initialize any of them, so if the DLL is not needed we should remove it.

tvsonar:
…the Runner

Fixes #408


duncanpMS:
LGTM

PR #433, opened 2015-09-28
Issue #208 Use improved FixAllProvider for S1905

### LINKED ISSUE #208: Use improved FixAllProvider for S1905
Relates to #144 and #141 (and #207)

```
x.OfType<int>().Cast<int>().ToList();
```

should be fixed at once in the whole document.

tvsonar:
Fixes #208

### DIFF
--- a/src/SonarLint/Rules/RedundantCastCodeFixProvider.cs
+++ b/src/SonarLint/Rules/RedundantCastCodeFixProvider.cs
@@ -25,6 +25,7 @@ using System.Threading.Tasks;
 using System.Linq;
 using Microsoft.CodeAnalysis.CodeActions;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarLint.Common;
 
 namespace SonarLint.Rules
 {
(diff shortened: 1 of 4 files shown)


src/SonarLint/Rules/RedundantCastCodeFixProvider.cs
```diff
@@ -50,10 +56,9 @@ public sealed override FixAllProvider GetFixAllProvider()
 
             var diagnostic = context.Diagnostics.First();
             var diagnosticSpan = diagnostic.Location.SourceSpan;
-            var syntaxNode = root.FindNode(diagnosticSpan);
+            var syntaxNode = root.FindNode(diagnosticSpan, getInnermostNodeForTie: true);
```
duncanpMS: Presumably the old code didn't handle some scenarios correctly - if so, could the failing case be added to the unit test for this fixer?

src/SonarLint/Rules/RedundantCastCodeFixProvider.cs
```diff
@@ -61,41 +66,58 @@ public sealed override FixAllProvider GetFixAllProvider()
...
             {
-                context.RegisterCodeFix(
-                    CodeAction.Create(
-                        Title,
-                        c => RemoveStaticMemberCall(context.Document, root, memberAccess)),
-                    context.Diagnostics);
-                return;
+                return root.ReplaceNode(
+                    syntaxNode,
+                    syntaxNode.WithoutAnnotations(DocumentBasedFixAllProvider<RedundantCast>.Annotation));
```
duncanpMS: Making the document fix-all provider responsible for removing the annotations might make it possible to have a single "RemoveCall" method.
If it is necessary to have both, it might help to rename the batch one to make the usage clearer e.g. "RemoveCallInBatch".

duncanpMS:
Not changed is this commit, but it is first example in supporting [html](http://vs.sonarlint.org/rules/index.html#version=1.3.0&ruleId=S1905) documentation for the rule handled by the S1905, or is just handled by IDE0004?

PR #434, opened 2015-09-28
Issue #228 Use improved FixAllProvider for S2437

### LINKED ISSUE #228: Use improved FixAllProvider for S2437
Relates to #159, #141

tvsonar:
Fixes #228

### DIFF
--- a/src/SonarLint/Common/DocumentBasedFixAllProvider.cs
+++ b/src/SonarLint/Common/DocumentBasedFixAllProvider.cs
@@ -24,16 +24,18 @@ using System.Threading.Tasks;
 using System.Linq;
 using Microsoft.CodeAnalysis.CodeActions;
 using System;
+using System.Collections.Generic;
+using System.Collections;
 
 namespace SonarLint.Common
 {
     internal sealed class DocumentBasedFixAllProvider<T> : FixAllProvider
     {
         private readonly string title;
-        private readonly Func<SyntaxNode, SyntaxNode, SyntaxNode> calculateNewRoot;
-        internal static readonly SyntaxAnnotation annotation = new SyntaxAnnotation(typeof(T).Name);
+        private readonly Func<SyntaxNode, SyntaxNode, Diagnostic, SyntaxNode> calculateNewRoot;
+        private static readonly string AnnotationKind = typeof(T).Name;
 
-        public DocumentBasedFixAllProvider(string title, Func<SyntaxNode, SyntaxNode, SyntaxNode> calculateNewRoot)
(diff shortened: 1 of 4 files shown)


src/SonarLint/Common/DocumentBasedFixAllProvider.cs
```diff
@@ -84,20 +86,66 @@ public override Task<CodeAction> GetFixAsync(FixAllContext fixAllContext)
...
                 var node = annotatedNodes.First();
-                root = calculateNewRoot(root, node);
-                annotatedNodes = root.GetAnnotatedNodes(annotation);
+                var annotation = node.GetAnnotations(AnnotationKind).First();
+                var diagnostic = diagnosticIdPairs.GetByB(annotation.Data);
+                root = calculateNewRoot(root, node, diagnostic);
+
+                annotatedNodes = root.GetAnnotatedNodes(AnnotationKind);
             }
             return root;
```
duncanpMS: It's not immediately obvious what this class/method is for; a few comments about the problem being solved and the solution would help. My understanding is as follows:

*We want to fix all issues in the document. However, as soon as the first issue is fixed we'll have changed the tree so we no longer know which other nodes to fix.
To solve this, we annotate the nodes in the original tree with a unique id that identifies the diagnostic instance. The annotations are preserved when the tree is rewritten.

Annotation data has to be a string, and diagnostic instances don't have a suitable string id property so we'll need to create a unique id for each one and track the relationship ourselves.*

src/SonarLint/Rules/SillyBitwiseOperationCodeFixProvider.cs
```diff
@@ -55,53 +61,60 @@ public sealed override FixAllProvider GetFixAllProvider()
...
-                        {
-                            var newRoot = root.ReplaceNode(assignment, assignment.Left)
-                                .WithAdditionalAnnotations(Formatter.Annotation);
-                            return Task.FromResult(context.Document.WithSyntaxRoot(newRoot));
-                        }),
-                    context.Diagnostics);
-                return;
+                return root.ReplaceNode(
+                    current,
+                    current.WithoutAnnotations(DocumentBasedFixAllProvider<SillyBitwiseOperation>.AnnotationKind));
```
duncanpMS: It might be better if the DocumentBasedFixAllProvider stripped out the annotations itself rather than requiring each code fix provider to do it (e.g. after line 118; the fix-all provider knows which diagnostic id has just been fixed, so it could strip out just that annotation).

duncanpMS:
LGTM

src/SonarLint/Rules/SillyBitwiseOperationCodeFixProvider.cs
tvsonar: Good point.

PR #438, opened 2015-09-28
Issue #196 Provide a way to unit test code fixes (fix all)

### LINKED ISSUE #196: Provide a way to unit test code fixes (fix all)

tvsonar:
Fixes #196


src/Tests/SonarLint.UnitTest/Verifier.cs
```diff
@@ -201,6 +246,56 @@ private static Document ApplyCodeFix(Document document, CodeAction codeAction)
             return actions;
         }
 
+        /// <summary>
+        /// Taken from Roslyn
+        /// http://source.roslyn.codeplex.com/#Microsoft.CodeAnalysis.Features/CodeFixes/FixAllOccurrences/FixAllCodeActionContext.DiagnosticProvider.cs
+        /// </summary>
```
duncanpMS: The Roslyn code requires the preservation of the copyright notice and disclaimer.
It might be better to put the copied class into a separate file.

duncanpMS:
LGTM

src/Tests/SonarLint.UnitTest/Verifier.cs
tvsonar: I've created a ticket in Roslyn to make their implementation public: https://github.com/dotnet/roslyn/issues/5687

sharwell: @duncanpMS A top-level notice like the following should meet the requirements for Roslyn, without needing to include it here in the file. In fact, if you are distributing the code in compiled form, you would need to include a notice like this because any notice embedded in individual source files would be lost.

https://github.com/DotNetAnalyzers/StyleCopAnalyzers/commit/601af15d755c7ffe4d31faa15b5636c42691a86a

:memo: IANAL

tvsonar: @sharwell Thanks for pointing to those lines.

tvsonar:
@dbolkensteyn I had to rework a bit this PR, because there was a new codefix provider that needed to work on `SyntaxToken`s. I left this change in a separate commit, so it's enough to review that one. Thanks

src/SonarLint/Common/FixAllProviders/DocumentBasedFixAllProviderForTokens.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+        }
+
+        protected override SyntaxToken GetReportedElement(Diagnostic d, SyntaxNode root) => root.FindToken(d.Location.SourceSpan.Start);
+        protected override bool IsElementMissing(SyntaxToken element) => element.IsMissing;
+
+        protected override SyntaxNode RemovePossiblyLeftAnnotation(SyntaxNode root, SyntaxAnnotation annotation)
+        {
+            var newToken = root.GetAnnotatedTokens(annotation).FirstOrDefault();
+            return newToken == null
```
dbolkensteyn: @tvsonar Visual Studio warns that this is always false

src/SonarLint/Common/FixAllProviders/DocumentBasedFixAllProviderForTokens.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        public DocumentBasedFixAllProviderForTokens(string title, Func<SyntaxNode, SyntaxToken, Diagnostic, SyntaxNode> calculateNewRoot)
+            : base(title, calculateNewRoot)
+        {
+        }
+
+        protected override SyntaxToken GetReportedElement(Diagnostic d, SyntaxNode root) => root.FindToken(d.Location.SourceSpan.Start);
+        protected override bool IsElementMissing(SyntaxToken element) => element.IsMissing;
+
+        protected override SyntaxNode RemovePossiblyLeftAnnotation(SyntaxNode root, SyntaxAnnotation annotation)
```
dbolkensteyn: @tvsonar it's not very clear from this method's name what it is supposed to do

src/SonarLint/Common/FixAllProviders/DocumentBasedFixAllProviderForTokens.cs
sharwell: :bulb: Related to DotNetAnalyzers/StyleCopAnalyzers#539.

src/SonarLint/Common/FixAllProviders/DocumentBasedFixAllProvider.cs
```diff
@@ -28,13 +28,13 @@
...
+    internal abstract class DocumentBasedFixAllProvider<TAnalyzer, TElement> : FixAllProvider
     {
         private readonly string title;
-        private readonly Func<SyntaxNode, SyntaxNode, Diagnostic, SyntaxNode> calculateNewRoot;
-        internal static readonly string AnnotationKind = typeof(T).Name;
+        private readonly Func<SyntaxNode, TElement, Diagnostic, SyntaxNode> calculateNewRoot;
+        protected static readonly string AnnotationKind = typeof(TAnalyzer).Name;
 
-        public DocumentBasedFixAllProvider(string title, Func<SyntaxNode, SyntaxNode, Diagnostic, SyntaxNode> calculateNewRoot)
+        protected DocumentBasedFixAllProvider(string title, Func<SyntaxNode, TElement, Diagnostic, SyntaxNode> calculateNewRoot)
```
dbolkensteyn: @tvsonar this giant `Func<>` is starting to look bad - perhaps it's better to introduce a proper interface with named and documented parameters instead

src/SonarLint/Common/FixAllProviders/DocumentBasedFixAllProvider.cs
```diff
@@ -87,52 +87,39 @@ public override Task<CodeAction> GetFixAsync(FixAllContext fixAllContext)
...
             {
-                var node = annotatedNodes.First();
-
-                var annotation = node.GetAnnotations(AnnotationKind).First();
+                var element = annotatedElements.First();
+                var annotation = GetAnnotations(element).First();
                 var diagnostic = diagnosticAnnotationPairs.GetByB(annotation);
-                root = calculateNewRoot(root, node, diagnostic);
-
+                root = calculateNewRoot(root, element, diagnostic);
```
dbolkensteyn: @tvsonar I don't like that this `calculateNewRoot()` is ultimately implemented by every code fixer, and that its only entry point is actually through the fix all provider. Can't the fix all provider use the `RegisterCodeFixesAsync()` entry point somehow?

src/SonarLint/Common/FixAllProviders/DocumentBasedFixAllProvider.cs
```diff
@@ -0,0 +1,249 @@
...
+using Microsoft.CodeAnalysis.CodeFixes;
+using System.Threading.Tasks;
+using System.Linq;
+using Microsoft.CodeAnalysis.CodeActions;
+using System.Collections.Generic;
+using System;
+
+namespace SonarLint.Common
+{
+    internal class DocumentBasedFixAllProvider : FixAllProvider
```
dbolkensteyn: @tvsonar why Document, and not `Annotation` for instance? Don't all fix all providers work on the document level?

dbolkensteyn:
LGTM

src/SonarLint/Common/FixAllProviders/DocumentBasedFixAllProvider.cs
tvsonar: @dbolkensteyn No, not all of them are Document based. When we execute the `CodeAction` that returns the changed solution. And from the solution we grab the document which we processed, but this would not work if we needed to change two documents at the same time. In this case we would need to work on solution level.

But you are right, it would be nice to include in the name that it is annotation based. Fo you have anything in mind? The `AnnotatedDocumentBasedFixAllProvider` seems rather long.

PR #439, opened 2015-09-28
Issue #435 Use improved FixAllProvider for S2737

### LINKED ISSUE #435: Use improved FixAllProvider for S2737

tvsonar:
Fixes #435

### DIFF
--- a/src/SonarLint/Rules/CatchRethrowCodeFixProvider.cs
+++ b/src/SonarLint/Rules/CatchRethrowCodeFixProvider.cs
@@ -26,6 +26,7 @@ using System.Linq;
 using Microsoft.CodeAnalysis.CodeActions;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Formatting;
+using SonarLint.Common;
 
 namespace SonarLint.Rules
 {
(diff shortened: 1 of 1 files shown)


tvsonar:
This PR depends on #196. Currently the batch fixer is not tested. And if it was, it would fail. We need to add an expected file for the batch fixer, as that is different from the result of the iterative fixing approach.

src/SonarLint/Rules/CatchRethrowCodeFixProvider.cs
```diff
@@ -59,14 +65,37 @@ public sealed override FixAllProvider GetFixAllProvider()
...
+                }),
                 context.Diagnostics);
         }
 
+        private static SyntaxNode CalculateNewRoot(SyntaxNode root, SyntaxNode currentNode)
+        {
+            var tryStatement = currentNode.Parent as TryStatementSyntax;
+
+            return tryStatement == null
+                ? root.ReplaceNode(currentNode, currentNode.WithoutAnnotations(
```
duncanpMS: I haven't managed to hit the case where the try statement is null. In what circumstances would this happen, and is the code clearing up the annotation in this case just be defensive?

duncanpMS:
LGTM

src/SonarLint/Rules/CatchRethrowCodeFixProvider.cs
tvsonar: It's just defensive. We are not proposing the fix if the parent is not a try statement. But we can get here from the fix all provider too, so I added the same check here as well. 

The annotation is removed if we are changing the try to a new one. But if for some reason we didn't find a `try` as the parent, then we need to be sure to remove the annotation, otherwise the `DocumentBasedFixAllProvider` would try to fix the same node again and again.

PR #442, opened 2015-09-29
Issue #441 Only format code that has been modified in a codefix

### LINKED ISSUE #441: Only format code that has been modified in a codefix
For example rule S3240 formats the whole syntax tree, instead of just the changed parts.

When the change is 
* a replace, we should only format the new node
* a remove, we shouldn't format, or should format the parent
* an insert, should format the newly added element, and maybe the perent.

tvsonar:
Fixes #441

### DIFF
--- a/src/SonarLint/Rules/CatchRethrowCodeFixProvider.cs
+++ b/src/SonarLint/Rules/CatchRethrowCodeFixProvider.cs
@@ -73,8 +73,11 @@ namespace SonarLint.Rules
                 Title,
                 c =>
                 {
-                    var newRoot = root.ReplaceNode(tryStatement, tryStatement.Block.Statements)
-                        .WithAdditionalAnnotations(Formatter.Annotation);
+                    var newParent = tryStatement.Parent.ReplaceNode(
+                        tryStatement,
+                        tryStatement.Block.Statements.Select(st => st.WithAdditionalAnnotations(Formatter.Annotation)));
+
+                    var newRoot = root.ReplaceNode(tryStatement.Parent, newParent);
                     return Task.FromResult(context.Document.WithSyntaxRoot(newRoot));
                 });
         }
--- a/src/SonarLint/Rules/IfConditionalAlwaysTrueOrFalseCodeFixProvider.cs
+++ b/src/SonarLint/Rules/IfConditionalAlwaysTrueOrFalseCodeFixProvider.cs
@@ -73,8 +73,8 @@ namespace SonarLint.Rules
             {
                 var block = ifStatement.Statement as BlockSyntax;
                 newRoot = block == null
-                    ? root.ReplaceNode(ifStatement, ifStatement.Statement)
-                    : root.ReplaceNode(ifStatement, block.Statements);
+                    ? root.ReplaceNode(ifStatement, ifStatement.Statement.WithAdditionalAnnotations(Formatter.Annotation))
+                    : root.ReplaceNode(ifStatement, block.Statements.Select(st => st.WithAdditionalAnnotations(Formatter.Annotation)));
             }
             else
             {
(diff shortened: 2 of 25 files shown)


src/SonarLint/Rules/CatchRethrowCodeFixProvider.cs
```diff
@@ -73,8 +73,10 @@ private static CodeAction CreateActionWithRemovedTryStatement(CodeFixContext con
                 Title,
                 c =>
                 {
-                    var newRoot = root.ReplaceNode(tryStatement, tryStatement.Block.Statements)
+                    var newParent = tryStatement.Parent.ReplaceNode(tryStatement, tryStatement.Block.Statements)
                         .WithAdditionalAnnotations(Formatter.Annotation);
+
+                    var newRoot = root.ReplaceNode(tryStatement.Parent, newParent);
```
duncanpMS: Everything in the new parent block is reformatted, including "incorrectly" formatted code that was outside the try block that was removed. This might be slightly more than the user was expecting, but it's definitely an improvement of the previous behaviour.

src/SonarLint/Rules/IfConditionalAlwaysTrueOrFalseCodeFixProvider.cs
```diff
@@ -86,12 +88,14 @@ private static SyntaxNode CalculateNewRoot(SyntaxNode root, IfStatementSyntax if
                 {
                     var block = ifStatement.Else.Statement as BlockSyntax;
                     newRoot = block == null
-                        ? root.ReplaceNode(ifStatement, ifStatement.Else.Statement)
-                        : root.ReplaceNode(ifStatement, block.Statements);
+                        ? root.ReplaceNode(ifStatement, ifStatement.Else.Statement.WithAdditionalAnnotations(Formatter.Annotation))
+                        : root.ReplaceNode(ifStatement.Parent,
+                            ifStatement.Parent.ReplaceNode(ifStatement, block.Statements)
+                                .WithAdditionalAnnotations(Formatter.Annotation));
```
duncanpMS: The two fixes in this class do the same as _CreateActionWithRemovedTryStatement_ above i.e. format everything in the new parent, not just the _block.Statements_ nodes.

src/SonarLint/Rules/SillyBitwiseOperationCodeFixProvider.cs
```diff
@@ -95,8 +94,8 @@ public sealed override FixAllProvider GetFixAllProvider()
                         c =>
                         {
                             var isReportingOnLeft = bool.Parse(diagnostic.Properties[SillyBitwiseOperation.IsReportingOnLeftKey]);
-                            var newRoot = root.ReplaceNode(binary, isReportingOnLeft?binary.Right :binary.Left)
-                                .WithAdditionalAnnotations(Formatter.Annotation);
+                            var newRoot = root.ReplaceNode(binary,
+                                (isReportingOnLeft ? binary.Right : binary.Left).WithAdditionalAnnotations(Formatter.Annotation));
```
duncanpMS: FYI I'm finding this style of formatting slightly easier to follow than the version in lines 80-81 of this file i.e. joining the _.WithAdditionalAnnotations(...)_ to the node being formatted, rather than having them on separate lines.

src/Tests/SonarLint.UnitTest/TestCases/IfConditionalAlwaysTrueOrFalse.cs
```diff
@@ -4,7 +4,7 @@ namespace Tests.Diagnostics
 {
     public class IfConditionalAlwaysTrueOrFalse
     {
-        public void DoSomething() { throw new NotSupportedException(); }
+        public void DoSomething(  ) { throw new NotSupportedException(); }
         public IfConditionalAlwaysTrueOrFalse(bool a, bool b)
         {
             if (a == b)
```
duncanpMS: Not changed in this fix, but the test code that should be kept and the test code that should be removed both call "DoSomething()" i.e. the test will still pass if the wrong conditions are being removed. If the code that should be kept called a different method then the test would guarantee that the correct code is being removed.

duncanpMS:
LGTM

src/SonarLint/Rules/CatchRethrowCodeFixProvider.cs
tvsonar: Formatting the block and then replacing the statements won't work, but formatting each statement individually does.

PR #445, opened 2015-09-29
Issue #164 Don't report issues on files containing generated code attributes

### LINKED ISSUE #164: Don't report issues on files containing generated code attributes
This issue relates to #85 and pushes it one level further: Instead of looking only at the filename to know whether or not the file is generated (and should therefore be excluded), we also will look at the file contents.

Specifically, the idea is to look for the usage of the following attributes - which are typically used to indicate generated code:
* `[GeneratedCode]` - https://msdn.microsoft.com/en-us/library/system.codedom.compiler.generatedcodeattribute(v=vs.110).aspx
* `[DebuggerNonUserCode]` - https://msdn.microsoft.com/en-us/library/system.diagnostics.debuggernonusercodeattribute(v=vs.110).aspx
* `[ExcludeFromCodeCoverage]` - https://msdn.microsoft.com/en-us/library/system.diagnostics.codeanalysis.excludefromcodecoverageattribute(v=vs.110).aspx
* `[CompilerGenerated]` - https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.compilergeneratedattribute(v=vs.110).aspx

@fmallet This can quite easily be implemented in SonarLint VS. However, this one would be a bit harder to implement on the MSBuild Runner side.

So far we have seen one project, Akka.NET, which has files not ending with generated extensions, but that actually contain these attributes.

tvsonar:
Fixes #164


tvsonar:
I had a brief discussion with Sam from StyleCop, and he said that they experienced performance issues with `ConditionalWeakTable<SyntaxTree, bool>`, because the `GetOrCreateValue` is locking if the key is not in the table. So they moved to `ConditionalWeakTable<Compilation, ConcurrentDictionary<SyntaxTree, bool>>`, which seems like a reasonable thing to do, to push the locking issue up to the level where there will be the least amount of entries in the table. 

They use a slightly different approach then us, because all their analyzers use a compilation start action to grab the compilation first, then select the corresponding "`SyntaxTree, bool`" cache. On our side, every time an analyzer runs, we go to the main cache, and get the "`SyntaxTree, bool`" cache from it. This is a slight overhead, but it still won't use the locking, because the `GetOrCreateValue` first tries to get the value based on the key, and if not found, then locks and creates the value. 

In any case, without performance testing we can't judge the speed of this solution.

src/SonarLint.Extra/Rules/CommentRegularExpression.cs
```diff
@@ -66,21 +66,25 @@ public static DiagnosticDescriptor CreateDiagnosticDescriptor(string diagnosticI
...
                 {
-                    var comments = from t in c.Tree.GetCompilationUnitRoot().DescendantTrivia()
-                                   where IsComment(t)
-                                   select t;
-
-                    foreach (var comment in comments)
-                    {
-                        var text = comment.ToString();
-                        foreach (var rule in RuleInstances.Where(rule => Regex.IsMatch(text, rule.RegularExpression)))
+                    compilationStartContext.RegisterSyntaxTreeActionInNonGenerated(
```
duncanpMS: You could have added an extension method that internally called these first few lines of code. It's not important from an implementation point of view, but it would have made the code review simpler by reducing the code churn. In particular, it would have avoided having to indent large sections of code that otherwise seem to be unchanged (the GitHub web diff experience doesn't handle the diff very well).

src/SonarLint/Helpers/DiagnosticAnalyzerContextHelper.cs
```diff
@@ -133,6 +170,50 @@ private static bool HasGeneratedFileName(this SyntaxTree tree)
...
+        private static bool HasGeneratedCodeAttribute(this SyntaxTree tree)
+        {
+            var attributeNames = tree.GetRoot()
+                .DescendantNodesAndSelf()
+                .OfType<AttributeSyntax>()
+                .Select(attribute => attribute.Name.ToString());
+
+            return attributeNames.Any(attributeName =>
+                GeneratedCodeAttributes.Any(generatedCodeAttribute =>
+                    attributeName.EndsWith(generatedCodeAttribute, StringComparison.InvariantCultureIgnoreCase)));
```
duncanpMS: This check is case-insensitive, but the checks against the file name and auto-generated comments are not. It might be better to make them all case-insensitive.

duncanpMS:
LGTM

src/SonarLint.Extra/Rules/CommentRegularExpression.cs
tvsonar: Not a bad idea.

src/SonarLint/Helpers/DiagnosticAnalyzerContextHelper.cs
tvsonar: I changed the comment checking to lower case, the file name check was already done. But for the attribute checking I think it was a mistake to use case insensitive comparison. The attribute name must be exactly as we expect, so I modified the code accordingly.

PR #446, opened 2015-09-29
Issue #444 S2292 fails on properties with expression body

### LINKED ISSUE #444: S2292 fails on properties with expression body

tvsonar:
Fixes #444


duncanpMS:
(manually tested, rule does not now throw)

tvsonar:
@duncanpMS In VS, if you turn on to break on all exceptions, then you'll find that this rule was throwing an exception for expression bodied properties. There was no test failure, because we were not testing it for C# 6 features.

PR #450, opened 2015-09-30
Issue #229 Fix the codefix for S1940 (inverted boolean check)

### LINKED ISSUE #229: Fix the codefix for S1940 (inverted boolean check)
The current implementation of the codefix always removes the parentheses, but in some cases this can't be done.

Related to #148 

For example: `var a = !!(args.Length == 0)` will get fixed into `!args.Length != 0` which then doesn't compile.

With dynamics, it should be possible to get programs that still compile but whose meaning will change, as in:
```
dynamic foo = null;
var a = !(foo == 0) + 3;
```

that will become:
```
var a = foo != 0 + 3
```

which will be evaluated not as `(foo != 0) + 3` as originally intended, but as `foo != (0 + 3)`

tvsonar:
Fixes #229

### DIFF
--- a/src/Tests/SonarLint.UnitTest/TestCases/BooleanCheckInverted.cs
+++ b/src/Tests/SonarLint.UnitTest/TestCases/BooleanCheckInverted.cs
@@ -24,6 +24,11 @@ namespace Tests.Diagnostics
             {
             }
             b = (a >= 10);
+
+            var c = true && !(new int[0].Length == 0); // Noncompliant
+
+            int[] args = { };
+            var a = !!(args.Length == 0); // Noncompliant
         }
 
         public static bool operator ==     (BooleanCheckInverted a, BooleanCheckInverted b)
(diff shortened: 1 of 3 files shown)


src/Tests/SonarLint.UnitTest/TestCases/BooleanCheckInverted.cs
```diff
@@ -24,6 +24,8 @@ public void Test()
             {
             }
             b = (a >= 10);
+
+            var c = true && !(new int[0].Length == 0); // Noncompliant
         }
```
duncanpMS: You could add a check for one of the original error cases that doesn't seem to i.e.
            int[] args = { };
            var a = !!(args.Length == 0);

(the other *dynamic* issue listed in the bug gives a RuntimeBinderException).

duncanpMS:
LGTM

PR #462, opened 2015-10-05
Improving batch fixer to not use GUIDs

tvsonar:


tvsonar:
@duncanpMS I got rid of the Diagnostic-Guid pairs in the batch fixer by creating individual `SyntaxAnnotation`s instead. (Internally that creates a unique sequence id, so there is no need to create it ourselves)

duncanpMS:
LGTM

PR #530, opened 2015-10-19
Issue #474 End statements should not be used

### LINKED ISSUE #474: End statements should not be used
RSPEC: https://jira.sonarsource.com/browse/RSPEC-1147

tvsonar:
Fixes #474

### DIFF
--- a/src/SonarLint/Rules.Description/S1147.html
+++ b/src/SonarLint/Rules.Description/S1147.html
@@ -1,0 +1,26 @@
+<p>
+    <code>End</code> statements exit the control flow of the program in an unstructured way. This statement stops code
+    execution immediately without executing <code>Dispose</code> or <code>Finalize</code> methods, or executing
+    <code>Finally</code> blocks. Therefore, it should be avoided.
+</p>
+
+<h2>Noncompliant Code Example</h2>
+<pre>
+Module Module1
+    Sub Print(ByVal str As String)
+       Try
+            ...
+            End       ' Noncompliant
+        Finally
+            ' do something important here
+            ...
+        End Try
+    End Sub
+End Module
+</pre>
+
+<h2>See</h2>
+
+<ul>
+    <li><a href="http://cwe.mitre.org/data/definitions/382">MITRE, CWE-382</a> - J2EE Bad Practices: Use of System.exit()</li>
+</ul>
--- a/src/SonarLint/Rules/EndStatementUsage.cs
+++ b/src/SonarLint/Rules/EndStatementUsage.cs
@@ -1,0 +1,64 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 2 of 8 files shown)


src/SonarLint/Rules.Description/S1147.html
```diff
@@ -0,0 +1,24 @@
+<p>
+    Calling <code>ToString()</code> on an object should always return a string. Returning <code>null</code> instead contravenes the method's implicit contract.
```
dbolkensteyn: @tvsonar euh? seems to be the wrong description

src/SonarLint/Rules/EndStatementUsage.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    [DiagnosticAnalyzer(LanguageNames.VisualBasic)]
+    [SqaleConstantRemediation("30min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.SecurityFeatures)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags(Tag.Cwe, Tag.Pitfall)]
+    public class EndStatementUsage : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S1147";
+        internal const string Title = "\"End\" statements should not be used";
```
dbolkensteyn: @tvsonar why not also mention "stop" in the title as it also gets detected by this rule?

dbolkensteyn: @tvsonar I wouldn't merge this rule with the Stop one afterall. Currently the existing rule in the VB.NET plugin only detects the usage of the `End` statement, which is a kind of maintainability issue, whereas it looks like `Stop` is not expected to be used in production code as it that one indeed fails to free variables and stuff

dbolkensteyn: @tvsonar btw it looks like `Stop` doesn't interrupt the control flow, it sends a signal to the debugger:
```
Stop

Console.WriteLine("Hello, world!")
Console.ReadLine()
```

The `Stop` statement acts as a breakpoint. When clicking Resume on the debugger, the program just continues as if the `Stop` statement wasn't there at all. When launched outside of Visual Studio, the `Stop` statement leads to a "crash" prompt, which if you close it, simply continues to execute the program. That happens in both "Debug" and "Release" configuration. See https://msdn.microsoft.com/en-us/library/6acc062s.aspx

=> I don't expect to see any usage of the `Stop` statement in production code, it probably is a very bad replacement to breakpoints

My feeling is that `Stop` could deserve another rule, with a critical severity and a dedicated explanation mentioning that breakpoints are a better alternative.

tvsonar: The Stop statement suspends execution, but unlike End, it does not close any files or clear any variables, unless it is encountered in a compiled executable (.exe) file. In https://msdn.microsoft.com/en-us/library/0wt87xba.aspx

tvsonar: Created #551 for the Stop statement

PR #543, opened 2015-10-22
Exception thrown in property getter

### LINKED ISSUE #497: Exceptions should not be thrown from property getters
RSPEC: https://jira.sonarsource.com/browse/RSPEC-2372

tvsonar:
Fixes #497


tvsonar:
In its current form, this analyzer reports quite a few issues for C#. One option to report less issues is that we could filter out getters with single throw statements as their body. (This occures in test files)

dbolkensteyn:
LGTM

dbolkensteyn:
@tvsonar if this bothers you, I'd rather not enable this rule on tests. In production code, I don't expect to see lots of properties that just throw an exception

PR #545, opened 2015-10-22
Var decl

### LINKED ISSUE #486: Multiple variables should not be declared on the same line
RSPEC: https://jira.sonarsource.com/browse/RSPEC-1659

tvsonar:
Fixes #486


tvsonar:
@dbolkensteyn This reports a lot of issues in C#. If we only report on field declarations (and not on local variable declarations), we get 4 times less issues. I would keep both reportings though. What do you think?

We are providing a batch code fixer for this, so it's easy to fix all issues.

dbolkensteyn:
@tvsonar well this is a convention rule, that people may or may not choose to follow. perhaps the projects we analyze in ruling do not follow it, which simply means that the rule shouldn't be enabled (and it doesn't mean that the behavior of it should be changed). so we probably just don't want to enable this rule by default to not generate noise.

dbolkensteyn:
LGTM

PR #546, opened 2015-10-23
Issue #515 Improve codefixes to correctly find reported error locations

### LINKED ISSUE #515: Expression based code fixes should not crash when the expression is an argument
When a diagnostic is reported on an expression that expression can be an argument to a method at the same time. In the code fixes we use `FindNode`, which automatically takes the outermost complete match, but that means taking the argument syntax instead of the expression syntax. So in these cases the `getInnermostNodeForTie: true` should be specified.

tvsonar:
Fixes #515


dbolkensteyn:
LGTM, it seems like this is fixing some crashes (I could get the inverted boolean to crash without this commit - but not with it)

PR #554, opened 2015-10-29
S1694 should not report on classes that have derived abstract methods

### LINKED ISSUE #517: S1694 should not report on classes that have derived abstract methods
An abstract class deriving from an abstract class can still be abstract if not all the base abstract methods are overriden, so there are still abstract methods in it.

tvsonar:
Fixes #517


dbolkensteyn:
LGTM

PR #555, opened 2015-10-29
Issue #523 S2333 - False positive on generic base classes

### LINKED ISSUE #523: S2333 - False positive on generic base classes
Hello I get a false positive on generic abstract class with virtual methods.

Example code:

```
    internal abstract class BaseClass<T>
    {
        public virtual string Process(string input) => input; //Compliant
    }

    internal class SubClass : BaseClass<string>
    {
        public override string Process(string input)
        {
            return "Test";
        }
    }
```

tvsonar:
Fixes #523


dbolkensteyn:
LGTM

PR #556, opened 2015-10-29
Issue #461 S1125 should not report on nullable booleans

### LINKED ISSUE #461: S1125 should not report on nullable booleans
currently, the implementation (I use C#) will also trigger for nullable booleans. (ignore the bad use of Random. Just for demo purpose)

image

This will result in code that does not compile.

tvsonar:
Fixes #461

### DIFF
--- a/src/SonarLint/Rules/BooleanLiteralUnnecessary.cs
+++ b/src/SonarLint/Rules/BooleanLiteralUnnecessary.cs
@@ -65,8 +65,7 @@ namespace SonarLint.Rules
                 c =>
                 {
                     var binaryExpression = (BinaryExpressionSyntax)c.Node;
-
-                    if (CheckForBooleanConstantsReport(binaryExpression, true, c))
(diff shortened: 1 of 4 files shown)


src/SonarLint/Rules/BooleanLiteralUnnecessary.cs
```diff
@@ -66,7 +66,8 @@ public override void Initialize(AnalysisContext context)
                 {
                     var binaryExpression = (BinaryExpressionSyntax)c.Node;
 
-                    if (CheckForBooleanConstantsReport(binaryExpression, true, c))
+                    if (IsOnNullableBoolean(binaryExpression, c.SemanticModel) ||
```
dbolkensteyn: @tvsonar why not place the `IsOnNullableBoolean` call in `IsOnNullableBoolean` ? it seems that they are called together every single time (their 4 occurrences perfectly match)

dbolkensteyn:
LGTM

PR #563, opened 2015-11-02
Issue #480 Line continuations should not be used

### LINKED ISSUE #480: Line continuations should not be used
RSPEC: https://jira.sonarsource.com/browse/RSPEC-2354

tvsonar:
Fixes #480


src/SonarLint/Rules/LineContinuation.cs
```diff
@@ -0,0 +1,74 @@
...
+            context.RegisterSyntaxTreeActionInNonGenerated(
+                c =>
+                {
+                    var lineContinuations = c.Tree.GetRoot().DescendantTokens()
+                        .SelectMany(token => token.TrailingTrivia)
+                        .Where(trivia => trivia.IsKind(SyntaxKind.LineContinuationTrivia));
+
+                    foreach (var lineContinuation in lineContinuations)
+                    {
+                        c.ReportDiagnostic(Diagnostic.Create(Rule, lineContinuation.GetLocation()));
```
dbolkensteyn: the warning is quite hard to see - but I don't have a better idea... `c.Tree.GetLocation(lineContinuation.FullSpan))` is the same thing as `_` is a single trivia (it doesn't itself include leading or trailing trivias)

src/Tests/SonarLint.UnitTest/Verifier.cs
```diff
@@ -220,9 +221,39 @@ private static void Verify(string path, string assemblyName, DiagnosticAnalyzer
...
+                {
+                    return line.LineNumber + 1 + lineValue;
+                }
+                else if (sign.Value == "-")
+                {
+                    return line.LineNumber + 1 - lineValue;
+                }
+                else
+                {
+                    return 1 + lineValue;
```
dbolkensteyn: i don't think you want to do +1 in the case an absolute line number (and not relative one) is given.

dbolkensteyn:
LGTM

PR #566, opened 2015-11-03
Issue #487 Statements should be on separate lines

### LINKED ISSUE #487: Statements should be on separate lines
RSPEC: https://jira.sonarsource.com/browse/RSPEC-122

tvsonar:
Fixes #487

### DIFF
--- a/src/SonarLint/Rules/SingleStatementPerLine.cs
+++ b/src/SonarLint/Rules/SingleStatementPerLine.cs
@@ -1,0 +1,133 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using Microsoft.CodeAnalysis;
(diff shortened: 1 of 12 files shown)


src/SonarLint/Rules/SingleStatementPerLineBase.cs
```diff
@@ -0,0 +1,142 @@
...
+        {
+            var statements = tree.GetRoot()
+                    .DescendantNodesAndSelf()
+                    .OfType<TStatementSyntax>()
+                    .Where(st => !StatementShouldBeExcluded(st));
+
+            return statements;
+        }
+
+        protected abstract Func<TStatementSyntax, bool> StatementShouldBeExcluded { get; }
```
dbolkensteyn: @tvsonar This should be an abstract method, not a getter-only property (I think we do have a rule to detect this in VB.NET)

src/SonarLint/Rules/SingleStatementPerLineBase.cs
```diff
@@ -0,0 +1,142 @@
...
+        where TStatementSyntax : SyntaxNode
+    {
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterSyntaxTreeActionInNonGenerated(
+                c =>
+                {
+                    var statements = GetStatements(c.Tree);
+
+                    var statementsByLines = new Dictionary<int, List<TStatementSyntax>>();
```
dbolkensteyn: @tvsonar it would probably be nicier to introduce a `MultiValueDictionary`: http://stackoverflow.com/a/3850969/287740

src/SonarLint/Rules/SingleStatementPerLine.cs
```diff
@@ -0,0 +1,141 @@
...
+        [SqaleSubCharacteristic(SqaleSubCharacteristic.Readability)]
+        [Tags(Tag.Convention)]
+        public class SingleStatementPerLine : SingleStatementPerLineBase<StatementSyntax>
+        {
+            protected override Func<StatementSyntax, bool> StatementShouldBeExcluded
+            {
+                get
+                {
+                    return st =>
+                        st == null ||
```
dbolkensteyn: @tvsonar can this case occur? it probably shouldn't

dbolkensteyn:
LGTM

src/SonarLint/Rules/SingleStatementPerLineBase.cs
tvsonar: @dbolkensteyn Okay, it'll be implemented in #574, because there are multiple occurrences.

PR #567, opened 2015-11-04
Files should contain only one top-level class or interface each

### LINKED ISSUE #482: Files should contain only one top-level class or interface each
Rule: https://jira.sonarsource.com/browse/RSPEC-1996

tvsonar:
Fixes #482

### DIFF
--- a/src/SonarLint/Rules/SingleTopLevelTypePerFileBase.cs
+++ b/src/SonarLint/Rules/SingleTopLevelTypePerFileBase.cs
@@ -1,0 +1,98 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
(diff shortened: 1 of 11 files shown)


src/SonarLint/Rules/SingleTopLevelTypePerFileBase.cs
```diff
@@ -0,0 +1,98 @@
...
+        internal const string DiagnosticId = "S1996";
+        internal const string Title = "Files should contain only one top-level class or interface each";
+        internal const string Description =
+            "A file that grows too much tends to aggregate too many responsibilities and inevitably becomes harder to " +
+            "understand and therefore to maintain. This is doubly true for a file with multiple top-level classes and " +
+            "interfaces. It is strongly advised to divide the file into one top-level class or interface per file.";
+        internal const string MessageFormat = "Put this top-level {0} declaration in a dedicated source file.";
+        internal const string Category = Constants.SonarLint;
+        internal const Severity RuleSeverity = Severity.Major;
+        internal const bool IsActivatedByDefault = false;
```
dbolkensteyn: @tvsonar why can't this rule be enabled by default? 

it doesn't seem to be a very valuable one to me - from a functional point of view, the rule that checks the number of lines in a file does the same as this RSPEC's description, but probably in a more efficient way.

PR #570, opened 2015-11-04
Issue #529 "Exit" statements should not be used

### LINKED ISSUE #529: "Exit" statements should not be used
RSPEC:  https://jira.sonarsource.com/browse/RSPEC-3385

tvsonar:
Fixes #529

### DIFF
--- a/src/SonarLint/Rules/ExitStatementUsage.cs
+++ b/src/SonarLint/Rules/ExitStatementUsage.cs
@@ -1,0 +1,92 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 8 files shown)


src/SonarLint/Rules/ExitStatementUsage.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace SonarLint.Rules
+{
+    [DiagnosticAnalyzer(LanguageNames.VisualBasic)]
+    [SqaleConstantRemediation("5min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.InstructionReliability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags(Tag.BadPractice, Tag.BrainOverload)]
+    public class ExitStatementUsage : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S3385";
```
dbolkensteyn: @tvsonar i have the feeling that "Exit Do, Exit For, Exit Try, and Exit While" should be considered as the `break` statement usage - and maybe not enabled by default

"Exit Function, Exit Property, and Exit Sub" sounds like a whole different thing, that is definitely worth to check

=> it seems to me that it would be better to have 2 different rules

tvsonar: We discussed this in person, and agreed to keep it as it is.

PR #571, opened 2015-11-05
S1699 should not report on virtual methods called on other instances

### LINKED ISSUE #569: S1699 should not report on virtual methods called on other instances
Hi there
I discovered a bug whithin the rule S1699.
When you have a copy constructor which takes an instance of the class and calls a method that is virtual on this instance, it reports that this is not allowed. But calling these methods on an already initialized instance should be allowed.

Example:
protected abstract class A 
{
   public A(A otherInstanceOfA)
   {
      // Gets reported as violation of S1699
      otherInstanceOfA.DoSomething();
   }
    public virtual void DoSomething(){}
}

tvsonar:
Fixes #569

### DIFF
--- a/src/SonarLint/Rules/ConstructorOverridableCall.cs
+++ b/src/SonarLint/Rules/ConstructorOverridableCall.cs
@@ -56,57 +56,37 @@ namespace SonarLint.Rules
 
         public override void Initialize(AnalysisContext context)
         {
-            context.RegisterCodeBlockStartActionInNonGenerated<SyntaxKind>(
-                cbc =>
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c =>
                 {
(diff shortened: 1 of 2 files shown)


src/SonarLint/Rules/ConstructorOverridableCall.cs
```diff
@@ -59,8 +59,6 @@ public override void Initialize(AnalysisContext context)
             context.RegisterCodeBlockStartActionInNonGenerated<SyntaxKind>(
```
dbolkensteyn: @tvsonar is this still required? can't the rule be transformed into a normal `RegisterSyntaxNode` one?

dbolkensteyn:
LGTM

src/SonarLint/Rules/ConstructorOverridableCall.cs
tvsonar: No, I could get rid of it.

PR #577, opened 2015-11-09
Issue #519 Reduce the number of issues reported by S125

### LINKED ISSUE #519: Reduce the number of issues reported by S125
S125 complains too much in SonarLint 1.3.0:
- it fires on code-like content in plain text (if the whole line is not parseable as a statement/expression, then it should not be assumed to contain code)
- it should never fire on summary comments. Any code included in these is very likely provided by way of example, and is not commented out code
- it should only fire once for any uninterrupted block of comments, not once for each apparent block of code separated by either blank lines or text, or embedded comments
 
Here is an example where the rule files, which is both a summary comment, and contains no code
    /// <summary>
    /// The C++ access level for a member function, e.g. private
    /// </summary>

If I have a block of commented out code like this, with blank lines, I get 3 occurrences of S125 reported:

S125    //Debug.Assert(this.MemberTypeName != null == storage.HasFlag(StorageClass.Member));
            //
S125   //if (storage.HasFlag(StorageClass.Member))
            //{
            //    output |= this.MemberTypeName.DisplayOn(builder, s);
            //    builder.Append(CppNameBuilder.NameSeparator);
            //    // Not trailing space wanted
S125    //    output = false;
            //}

tvsonar:
Fixes #519

### DIFF
--- a/src/SonarLint/Rules/CommentedOutCode.cs
+++ b/src/SonarLint/Rules/CommentedOutCode.cs
@@ -69,44 +69,27 @@ namespace SonarLint.Rules
                         Action<IEnumerable<SyntaxTrivia>> check =
                             trivias =>
                             {
-                                var lastCommentedCodeLine = int.MinValue;
-
+                                var shouldReport = true;
                                 foreach (var trivia in trivias)
                                 {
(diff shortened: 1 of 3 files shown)


dbolkensteyn:
LGTM

PR #578, opened 2015-11-09
Issue #211 S1172 should not report on event handlers

### LINKED ISSUE #211: S1172 should not report on event handlers

tvsonar:
Fixes #211

### DIFF
--- a/src/SonarLint/Rules/MethodParameterUnused.cs
+++ b/src/SonarLint/Rules/MethodParameterUnused.cs
@@ -65,7 +65,8 @@ namespace SonarLint.Rules
                     var methodSymbol = cbc.OwningSymbol as IMethodSymbol;
                     if (methodDeclaration == null ||
                         methodSymbol == null ||
-                        !IsMethodCandidate(methodSymbol))
+                        !IsMethodCandidate(methodSymbol) ||
+                        IsMethodProbablyEventHandler(methodSymbol, cbc.SemanticModel.Compilation))
                     {
                         return;
                     }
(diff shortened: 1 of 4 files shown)


src/SonarLint/Rules/MethodParameterUnused.cs
```diff
@@ -124,5 +125,29 @@ private static bool IsMethodCandidate(IMethodSymbol methodSymbol)
...
+                methodSymbol.Parameters.Length != 2)
+            {
+                return false;
+            }
+
+            var eventArgs = methodSymbol.Parameters[1];
+            var eventArgsType = eventArgs.Type as INamedTypeSymbol;
+            if (eventArgsType == null)
+            {
+                return false;
```
dbolkensteyn: @tvsonar the conservative option (prefer false positives over false negatives) here would be to return `true` if we don't know exactly what the type is

src/SonarLint/Rules/MethodParameterUnused.cs
```diff
@@ -124,5 +125,29 @@ private static bool IsMethodCandidate(IMethodSymbol methodSymbol)
...
+            }
+
+            var sysEventArgs = compilation.GetTypeByMetadataName("System.EventArgs");
+            if (sysEventArgs == null)
+            {
+                return false;
```
dbolkensteyn: @tvsonar the conservative option (prefer false positives over false negatives) here would be to return `true` if we don't know exactly what the type is

dbolkensteyn:
LGTM

src/SonarLint/Rules/MethodParameterUnused.cs
tvsonar: Thanks, changed both of them.

PR #584, opened 2015-11-10
Issue #579 S1994 should check the body of the "for" for updates

### LINKED ISSUE #579: S1994 should check the body of the "for" for updates

tvsonar:
Fixes #579


tvsonar:
This is not a good fix, we should lower the severity of the RSPEC instead.

tvsonar:
Instead we went with #586

PR #585, opened 2015-11-10
Issue #406 S3240 should not report on any pattern that is not fixable

### LINKED ISSUE #406: S3240 should not report on any pattern that is not fixable
Related to #230 

The codefix provider was fixed to not provide a fix which results in compile time errors. This logic should be moved to the analyzer to not report at all on these cases.

tvsonar:
Fixes #406


dbolkensteyn:
LGTM

PR #609, opened 2015-11-12
Magic number

### LINKED ISSUE #598: Drop the magic number rule (S109)
It reports too many issues, and can't be enabled by default.

tvsonar:
Fixes #598


dbolkensteyn:
the `IEnumerable` reflection loading of parameters is probably dead code due to this and should also be dropped

dbolkensteyn:
LGTM

PR #615, opened 2015-11-13
Fix missing

### LINKED ISSUE #611: Fix missing analyzer dependencies warning with SonarLint NuGet
When using SonarLint 1.4.0 in Visual Studio 2015 Update 1, missing dependencies warnings are shown when Visual Studio is loaded with just C# or VB.NET projects, but not both.

warnings

The root cause is that SonarLint ships with just one assembly containing both the C# and VB.NET analyzers, and that assembly therefore depends on both the C# and VB.NET Roslyn compiler assemblies. However, Visual Studio will only load the C# Roslyn compiler assemblies (and not the VB.NET ones) when a C# project is opened.

The proper fix is to have 3 separate DLLs: a "common" one, one containing C# analyzers, and another one containing VB.NET analyzers. The NuGet package installation script should check the project's language to know which assemblies to add as analyzers.

The VSIX embed all 3 DLLs.

Thanks @jessehouwing @srivatsn for reporting this issue.

tvsonar:
Fixes #611


dbolkensteyn:
This PR is impossible to review due to the high number of changed files (592)

dbolkensteyn:
I'll just make a quick manual test on VS2015 Update 1 to see if it installs & runs smoothly.

dbolkensteyn:
I can't seem to get any SonarLint rule to show up in the ruleset editor, nor to be executed when writing code in a new Console App, after installing SonarLint 1.5.0's VSIX on a fresh VS 2015 Update 1 install.

dbolkensteyn:
By the way, I don't think this pull request should be moved into the "In Test & Review" column before you validated yourself that it properly fixes the original issue (#611 mentions VS 2015 Update 1)

tvsonar:
@dbolkensteyn There's a note on that issue saying that this issue has nothing to do with Update 1.

tvsonar:
@dbolkensteyn I just double checked this in VS2015 and Update1, it works in both versions both as a Nuget package and as a VSIX.

dbolkensteyn:
LGTM

PR #622, opened 2015-11-18
Issue #521 S3249 should not report on reference equality guard conditions in Equals

### LINKED ISSUE #521: S3249 should not report on reference equality guard conditions in Equals
S3249 fires for an Equals implementation that calls base.Equals where the nearest overridden implementation is that in object. The help states "no two Objects that use those base methods will ever be equal", but this is incorrect if the argument to Equals is the same object as the receiver. i.e. x.Equals(x), then the two objects are one and the same and are equal. 
Calling base.Equals and landing in object is the same as calling ReferenceEquals, and either can be a useful optimization as a guard clause at the top before performing more expensive value comparisons. I think if you believe this rule has value, then it should be more careful to differentiate between "base.Equals(x) && <class specific conditions)" (where calling base.Equals would be an error, since then only the same objects would ever be equal) and "base.Equals(x) || <class specific conditions>" where it isn't actually incorrect.
The rule could advise replacing the call with ReferenceEquals in the 2nd case, but not to remove it as although semantics should be preserved (if the Equals implementation is correct), removing it may have an adverse performance impact in some code.

Here's an example:

        public override bool Equals(object obj)
        {
            if (base.Equals(obj))              <------------
            {
                return true;
            }

            NameNode otherName = obj as NameNode;
            return otherName != null && otherName.Name == this.Name;
        }
Warning	S3249	Remove this "base" call.	

rather than removing the guard clause entirely, this should be rewritten as:

        public override bool Equals(object obj)
        {
            if (object.ReferenceEquals(obj, this))
            {
                return true;
            }
...

tvsonar:
Fixes #521

### DIFF
--- a/src/SonarLint.CSharp/Rules/GetHashCodeEqualsOverride.cs
+++ b/src/SonarLint.CSharp/Rules/GetHashCodeEqualsOverride.cs
@@ -18,6 +18,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
  */
 
+using System;
 using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
@@ -56,7 +57,9 @@ namespace SonarLint.Rules.CSharp
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Rule); } }
 
-        private static readonly string[] MethodNames = { "GetHashCode", "Equals" };
+        private const string EqualsName = "Equals";
+        private static readonly string[] MethodNames = { "GetHashCode", EqualsName };
+        private static readonly SyntaxNode TrueLiteral = SyntaxFactory.LiteralExpression(SyntaxKind.TrueLiteralExpression);
 
         public override void Initialize(AnalysisContext context)
         {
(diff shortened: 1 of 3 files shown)


src/Tests/SonarLint.UnitTest/TestCases/GetHashCodeEqualsOverride.cs
```diff
@@ -13,10 +13,36 @@ public GetHashCodeEqualsOverride(int x)
         }
         public override int GetHashCode()
         {
+            if (base.Equals(obj)) // Compliant
```
dbolkensteyn: @tvsonar this doesn't seem to make a lot of sense

src/SonarLint.CSharp/Rules/GetHashCodeEqualsOverride.cs
```diff
@@ -79,37 +82,98 @@ public override void Initialize(AnalysisContext context)
...
+            {
+                return false;
+            }
+
+            var argument = invocation.ArgumentList.Arguments.First().Expression;
+            var parameterSymbol = semanticModel.GetSymbolInfo(argument).Symbol as IParameterSymbol;
+            if (parameterSymbol == null ||
+                !parameterSymbol.ContainingSymbol.Equals(declaredMethod))
+            {
+                return false;
```
dbolkensteyn: this check might be a bit strong, it doesn't handle aliasing, as in:

```
MyClass o = obj as MyClass;
if (o == null)
{
  return null;
}

if (base.Equals(o))
{
  return true;
}

// ...
```

src/SonarLint.CSharp/Rules/GetHashCodeEqualsOverride.cs
```diff
@@ -79,37 +82,98 @@ public override void Initialize(AnalysisContext context)
...
+            }
+
+            return IfStatementWithSingleReturnTrue(ifStatement.Statement);
+        }
+
+        private static bool IfStatementWithSingleReturnTrue(StatementSyntax statement)
```
dbolkensteyn: @tvsonar looks like this name's method could be improved - it's unrelated to an if statement. - it checks whether the given statement is a return one, optionally unwrapping its containing block.

dbolkensteyn:
LGTM

PR #631, opened 2015-11-19
Issue #624 Load rule parameters from additional files

### LINKED ISSUE #624: Load rule parameters from additional files

tvsonar:
Fixes #624

### DIFF
--- a/src/SonarLint.VisualBasic/Rules/OptionalParameter.cs
+++ b/src/SonarLint.VisualBasic/Rules/OptionalParameter.cs
@@ -47,7 +47,9 @@ namespace SonarLint.Rules.VisualBasic
             parameter.Modifiers.First(m => m.IsKind(SyntaxKind.OptionalKeyword)).GetLocation();
 
         protected override IEnumerable<ParameterSyntax> GetParameters(MethodBaseSyntax method) =>
-            method.ParameterList.Parameters;
+            method.ParameterList == null
+            ? Enumerable.Empty<ParameterSyntax>()
+            : method.ParameterList.Parameters;
 
         protected override bool IsOptional(ParameterSyntax parameter) =>
             parameter.Modifiers.Any(m => m.IsKind(SyntaxKind.OptionalKeyword));
--- a/src/SonarLint/Common/AnalyzerAdditionalFile.cs
+++ b/src/SonarLint/Common/AnalyzerAdditionalFile.cs
@@ -1,0 +1,44 @@
+/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using Microsoft.CodeAnalysis.Text;
+using Microsoft.CodeAnalysis;
+using System.Threading;
+using System.IO;
+
+namespace SonarLint.Common
+{
+    public sealed class AnalyzerAdditionalFile : AdditionalText
+    {
+        private readonly string path;
+
+        public AnalyzerAdditionalFile(string path)
+        {
+            this.path = path;
+        }
+
+        public override string Path => path;
+
+        public override SourceText GetText(CancellationToken cancellationToken)
+        {
+            return SourceText.From(File.ReadAllText(path));
+        }
+    }
+}
--- a/src/SonarLint/Helpers/ParameterLoader.cs
+++ b/src/SonarLint/Helpers/ParameterLoader.cs
@@ -1,0 +1,151 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System;
+using System.Globalization;
+using Microsoft.CodeAnalysis;
+using SonarLint.Common;
+using System.Collections.Immutable;
+using System.Collections.Generic;
+using System.Linq;
+using System.Xml.Linq;
+using Microsoft.CodeAnalysis.Diagnostics;
+using System.Reflection;
+using System.IO;
+using System.Collections.Concurrent;
+
+namespace SonarLint.Helpers
+{
+    public static class ParameterLoader
+    {
+        public static readonly string ParameterConfigurationFileName = "SonarLint.xml";
+
+        public class RuleParameterValue
+        {
+            public string ParameterKey { get; set; }
+            public string ParameterValue { get; set; }
+        }
+        public class RuleParameterValues
+        {
+            public string RuleId { get; set; }
+            public List<RuleParameterValue> ParameterValues { get; set; } = new List<RuleParameterValue>();
+        }
+
+        //todo: this can become private when we remove the template rule
+        public static ImmutableList<RuleParameterValues> ParseParameters(XContainer xml)
+        {
+            var builder = ImmutableList.CreateBuilder<RuleParameterValues>();
+            foreach (var rule in xml.Descendants("Rule").Where(e => e.Elements("Parameters").Any()))
+            {
+                var analyzerId = rule.Elements("Key").Single().Value;
+
+                var parameterValues = rule
+                    .Elements("Parameters").Single()
+                    .Elements("Parameter")
+                    .Select(e => new RuleParameterValue
+                    {
+                        ParameterKey = e.Elements("Key").Single().Value,
+                        ParameterValue = e.Elements("Value").Single().Value
+                    });
+
+                var pvs = new RuleParameterValues
+                {
+                    RuleId = analyzerId
+                };
+                pvs.ParameterValues.AddRange(parameterValues);
+
+                builder.Add(pvs);
+            }
+
+            return builder.ToImmutable();
+        }
+
+        private readonly static ConcurrentDictionary<DiagnosticAnalyzer, byte> ProcessedAnalyzers = new ConcurrentDictionary<DiagnosticAnalyzer, byte>();
+
+        public static void SetParameterValues(DiagnosticAnalyzer parameteredAnalyzer,
+            AnalyzerOptions options)
+        {
+            if (ProcessedAnalyzers.ContainsKey(parameteredAnalyzer))
+            {
+                return;
+            }
+
+            var additionalFile = options.AdditionalFiles
+                .FirstOrDefault(f => ConfigurationFilePathMatchesExpected(f.Path));
+
+            if (additionalFile == null)
+            {
+                return;
+            }
+
+            var filePath = additionalFile.Path;
+            var xml = XDocument.Load(filePath);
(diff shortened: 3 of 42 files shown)


build/Version.props
```diff
@@ -1,6 +1,6 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <MainVersion>1.6.0</MainVersion>
+    <MainVersion>1.5.1</MainVersion>
```
dbolkensteyn: @tvsonar I don't think we're going to need a version change as part of this PR.

src/SonarLint.Vsix/Diagnostic.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>SonarLint</id>
-    <version>1.6.0</version>
+    <version>1.5.1</version>
```
dbolkensteyn: @tvsonar same here, not required

src/SonarLint/Common/AnalyzerAdditionalFile.cs
```diff
@@ -0,0 +1,44 @@
...
+        public AnalyzerAdditionalFile(string path)
+        {
+            this.path = path;
+        }
+
+        public override string Path => path;
+
+        public override SourceText GetText(CancellationToken cancellationToken)
+        {
+            return SourceText.From(File.OpenRead(path));
```
dbolkensteyn: @tvsonar in the doc of the `SourceText.From()` function, it says:

```
//     Reads from the beginning of the stream. Leaves the stream open.
```

Shouldn't we close & dispose of it then?

src/SonarLint/Helpers/ParameterLoader.cs
```diff
@@ -0,0 +1,171 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+using System.Reflection;
+using System.IO;
+using System.Security;
+
+namespace SonarLint.Helpers
+{
+    public static class ParameterLoader
+    {
+        public static readonly string CheckedFileName = "sonarlint.xml";
```
dbolkensteyn: @tvsonar why not SonarLint.xml ?

tvsonar: it doesn't really matter, we're using `StringComparison.InvariantCultureIgnoreCase` for the comparison.

src/SonarLint/Helpers/ParameterLoader.cs
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        private readonly static HashSet<DiagnosticAnalyzer> ProcessedAnalyzers = new HashSet<DiagnosticAnalyzer>();
+        private readonly static object lockObject = new object();
+
+        public static void SetParameterValues(DiagnosticAnalyzer parameteredAnalyzer,
+            AnalyzerOptions options)
+        {
+            var additionalFile = options.AdditionalFiles
+                .FirstOrDefault(f => f.Path.EndsWith(CheckedFileName, StringComparison.InvariantCultureIgnoreCase));
```
tvsonar: @dbolkensteyn should we check exact file match here? we are just checking the path ending with sonarlint.xml.

src/SonarLint/Helpers/ParameterLoader.cs
```diff
@@ -0,0 +1,171 @@
...
+
+            if (additionalFile == null || ProcessedAnalyzers.Contains(parameteredAnalyzer))
```
dbolkensteyn: @tvsonar This isn't thread-safe.

HashSet is not thread-safe: "Any instance members are not guaranteed to be thread safe" https://msdn.microsoft.com/en-us/library/bb359438(v=vs.110).aspx

This means that calling instance methods from several threads concurrently might blow up the internal state of the object. For example, calling `Contains()` from multiple threads concurrently might not work. Whether or not it will actually work only depends on how `Contains()` is implemented - but the doc says you cannot assume this to be safe => It's not safe.

Now, a simple fix would be to replace it by a concurrent collection, whose instance methods are thread-safe, such as `ConcurrentDictionary` as suggested in http://stackoverflow.com/questions/18922985/concurrent-hashsett-in-net-framework

I think that would be perfectly acceptable solution, however it wouldn't be sufficient to guarantee that parameters for a given analyzer are loaded at most once. Indeed:

1. 2 threads can call `SetParameterValues()` on the same analyzer instance
2. They both call `Contains()` which returns false
3. They both proceed with the loading of their parameters
4. They acquire the lock to write the result back, the last one to execute won't update the set as the analyzer is already present

So you'd actually have to make the whole reading of the parameters part of the critical section if you'd wanted to ensure that it gets executed at most once.

To sum up, multi-threading is super hard to get right, especially when we implement it by ourselves by introducing locks and explicitly acquiring them and so on.

My :+1: goes for removing the lock, and replacing the `HashSet` by a `ConcurrentDictionary`, and not care that the same parameters could potentially be set several times on the same analyzer.

src/SonarLint/Helpers/ParameterLoader.cs
```diff
@@ -0,0 +1,171 @@
...
+
+            }
+            catch (Exception exception)
+            {
+                if (exception is ArgumentNullException ||
+                    exception is SecurityException ||
+                    exception is FileNotFoundException ||
+                    exception is UriFormatException)
+                {
+                    // we expect these exceptions from the XML parsing.
```
dbolkensteyn: @tvsonar how can any of these exception be expected? If the additional file property is set, then the additional file should exist.

There is already some code above to verify that there is a SonarLint XML additional file, and quit the function if that's not the case.

src/SonarLint/Helpers/ParameterLoader.cs
```diff
@@ -0,0 +1,171 @@
...
+                    return;
+                }
+
+                throw;
+            }
+            finally
+            {
+                CacheProcessedAnalyzer(parameteredAnalyzer);
```
dbolkensteyn: @tvsonar It looks strange to cache from the finally

src/SonarQube.SonarLint.Descriptor/Descriptor.cs.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>SonarLint.Descriptor.CSharp</id>
-    <version>1.6.0</version>
+    <version>1.5.1</version>
```
dbolkensteyn: as above, unnecessary

src/SonarQube.SonarLint.Descriptor/Descriptor.vbnet.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>SonarLint.Descriptor.VisualBasic</id>
-    <version>1.6.0</version>
+    <version>1.5.1</version>
```
dbolkensteyn: as above, unnecessary

src/SonarQube.SonarLint.Runner/DiagnosticsRunner.cs
```diff
@@ -57,7 +60,10 @@ public IEnumerable<Diagnostic> GetDiagnostics(Compilation compilation)
 
             using (var tokenSource = new CancellationTokenSource())
             {
-                var compilationWithAnalyzer = modifiedCompilation.WithAnalyzers(diagnosticAnalyzers, cancellationToken: tokenSource.Token);
+                var compilationWithAnalyzer = modifiedCompilation.WithAnalyzers(
+                    diagnosticAnalyzers,
+                    new AnalyzerOptions(ImmutableArray.Create<AdditionalText>(new AnalyzerAdditionalFile(configuration.Path))),
```
dbolkensteyn: nice :)

src/SonarQube.SonarLint.Runner/Program.cs
```diff
@@ -34,12 +33,27 @@ public static class Program
     {
         public static int Main(string[] args)
         {
+            if (args.Length != 3)
+            {
+                Write("Expected parameters: ");
+                Write("[Input configuration path: should end with sonarlint.xml, otherwise rule parameter value will be ignored]");
```
dbolkensteyn: @tvsonar the ": should end with sonarlint.xml, otherwise rule parameter value will be ignored]" part can be removed - the fail fast behavior will make it explicit enough.

src/SonarQube.SonarLint.Runner/Program.cs
```diff
@@ -34,12 +33,27 @@ public static class Program
...
+                Write("[Output file path]");
+                Write("[AnalyzerLanguage: 'cs' for C#, 'vbnet' for VB.Net]");
+
+                return -1;
+            }
+
+            if (!args[0].EndsWith(ParameterLoader.CheckedFileName, StringComparison.InvariantCultureIgnoreCase))
```
dbolkensteyn: @tvsonar should be moved to `Configuration`, it's not the responsibility of the rule runner executable to know about this

src/SonarQube.SonarLint.Runner/Configuration.cs
```diff
@@ -29,66 +29,39 @@
...
+        public string Path { get; private set; }
 
         public bool IgnoreHeaderComments { get; }
         public IImmutableList<string> Files { get; }
         public IImmutableSet<string> AnalyzerIds { get; }
 
-        public Configuration(XContainer xml, AnalyzerLanguage language)
+        public Configuration(string path, AnalyzerLanguage language)
         {
+            Path = path;
```
dbolkensteyn: @tvsonar Here we should fail fast is the provided XML filename doesn't match `SonarLint.xml`

src/SonarQube.SonarLint.Runner/Runner.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>SonarLint.Runner</id>
-    <version>1.6.0</version>
+    <version>1.5.1</version>
```
dbolkensteyn: not necessary

src/Tests/SonarLint.SonarQube.Integration.UnitTest/SonarLint.SonarQube.Integration.UnitTest.csproj
```diff
@@ -132,7 +132,7 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="app.config" />
-    <None Include="TestResources\ConfigurationTest.xml">
+    <None Include="TestResources\SonarLint.xml">
```
dbolkensteyn: @tvsonar here SonarLint.xml is already properly cased :)

src/Tests/SonarLint.UnitTest/Verifier.cs
```diff
@@ -191,7 +192,10 @@ private static Document GetDocument(string filePath, string assemblyName, AdhocW
 
                 var compilationWithOptions = compilation.WithOptions(compilationOptions);
                 var compilationWithAnalyzer = compilationWithOptions
-                    .WithAnalyzers(ImmutableArray.Create(diagnosticAnalyzer), cancellationToken: tokenSource.Token);
+                    .WithAnalyzers(
+                        ImmutableArray.Create(diagnosticAnalyzer),
+                        new AnalyzerOptions(ImmutableArray.Create<AdditionalText>()),
```
dbolkensteyn: @tvsonar doesnt this have the exact same effect as before? what's the point of having more code to do the same thing?

src/SonarLint/Helpers/ParameterLoader.cs
tvsonar: @dbolkensteyn we could do the caching even before we calculated anything. The thing is, we don't care if the parsing failed or succeeded, we don't want to do it again.

dbolkensteyn:
See inline comments

src/SonarLint.VisualBasic/Rules/OptionalParameter.cs
```diff
@@ -47,7 +47,9 @@ public class OptionalParameter : OptionalParameterBase<SyntaxKind, MethodBaseSyn
             parameter.Modifiers.First(m => m.IsKind(SyntaxKind.OptionalKeyword)).GetLocation();
 
         protected override IEnumerable<ParameterSyntax> GetParameters(MethodBaseSyntax method) =>
-            method.ParameterList.Parameters;
+            method.ParameterList == null
+            ? Enumerable.Empty<ParameterSyntax>()
+            : method.ParameterList?.Parameters;
```
dbolkensteyn: `?.` doesn't seem to be wanted here, as `ParameterList` is already checked for not being null. Otherwise it seems like `GetParameters` is expected under some condition to return null.

src/SonarLint/Helpers/ParameterLoader.cs
```diff
@@ -0,0 +1,151 @@
...
+
+            return builder.ToImmutable();
+        }
+
+        private readonly static ConcurrentDictionary<DiagnosticAnalyzer, byte> ProcessedAnalyzers = new ConcurrentDictionary<DiagnosticAnalyzer, byte>();
+
+        public static void SetParameterValues(DiagnosticAnalyzer parameteredAnalyzer,
+            AnalyzerOptions options)
+        {
+            var additionalFile = options.AdditionalFiles
```
dbolkensteyn: It could make sense to not even run this code if the analyzer is already cached.

dbolkensteyn:
LGTM

Issue #632, opened 2015-11-20
S1172 Unused parameters for method syntax defined by delegate


pedromagueija:
There is a false positive reported by SonarLint when an parameter is not used in a method that is used as a delegate e.g.: an event handler.

```C#
namespace ConsoleApplication2
{
    using System;

    internal class Program
    {
        private static void Main(string[] args)
        {
            SomeClass instance= new SomeClass();
            instance.SomeEvent += OnSomeEvent;    
        }

        // arg1 is marked as unused (true, but necessary)
        private static void OnSomeEvent(string arg1, int arg2)
        {
            if (arg2 == 0)
            {
                Console.WriteLine("its zero");
            }
        }
    }

    class SomeClass
    {
        public event CustomDelegate SomeEvent;
    }

    delegate void CustomDelegate(string arg1, int arg2);
}
```

tvsonar:
@pedromagueija thanks for reporting this issue. I think in this case `arg1` is unused, and it's not necessary, because you could change your custom delegate to have only one parameter.

We have a mechanism to recognize event handlers, but that only works for delegates that conform to the [event handler definition guidelines](https://msdn.microsoft.com/library/ms229011(v=vs.100).aspx). (Two parameters, the first is `object`, the second one `EventArgs`). This was handled in #211.

Do you think we should exclude other patterns as well? What's the reason that you are not using the standard [`EventHandler<T>`](https://msdn.microsoft.com/en-us/library/db0etb8x(v=vs.110).aspx) delegate?

pedromagueija:
Hi @tvsonar,

I missed to mentioned that the event (and the delegate) in question are from a 3rd party API. I have no control over the source (proprietary API).

The example I posted is what I assume internally is done in that API.

Sorry about that.

Cheers.

tvsonar:
For private/internal methods we could do a bit more: we can check if the method in question is being assigned to an event anywhere in the class, and if so, then we could exclude it.

pedromagueija:
Is it possible to check if the method is being assigned to an event or delegate defined outside the assembly, and in that case exclude it from this check?

tvsonar:
Yes, we can check if the event is declared outside the assembly.

PR #635, opened 2015-11-20
Issue #60 "async" methods should not return "void"

### LINKED ISSUE #60: "async" methods should not return "void"
Replaces http://jira.sonarsource.com/browse/LINTVS-166

`async void` is "fire and forget" method, you can't await for the execution of such a method. Having `async void` methods will make your program work sometimes, and not in others, because of timing issues. For example:
```
class HttpPrinter
{
  private string content;
  public async Task PrintContent(string url)
  {
    CallNetwork(url);
    await Task.Delay(1000);
    Console.Write(content);
  }

  public async void CallNetwork(string url) //Noncompliant
  {
    var client = new HttpClient();
    var response = await client.GetAsync(url);
    content = await response.Content.ReadAsStringAsync();
  }
}
```
The above code works correctly if the web request finishes under one second, otherwise nothing is printed.

Furthermore, the caller of an `async void` method won't be able to catch exceptions thrown in the `async void` method. 

Use `async Task` instead of `async void`.

`async void` should only be used for event handlers, as they match the "fire and forget" nature of `async void`.

Implements http://jira.sonarsource.com/browse/RSPEC-3168

tvsonar:
Fixes #60

### DIFF
--- a/src/SonarLint.Extra/Rules.Description/S3168.html
+++ b/src/SonarLint.Extra/Rules.Description/S3168.html
@@ -1,0 +1,54 @@
+<p>
+    An async method with a void return type is a "fire and forget" method best reserved for event handlers because there's no way to wait for the method's execution to complete and respond accordingly. There's also no way to catch exceptions thrown from the method.
+</p>
+<p>
+    Having an async void method that is not an event handler could mean your program works some times and not others because of timing issues. Instead, async methods should return Task.
+</p>
+<p>
+    This rule raises an issue when non-event handler methods are both async and void.
+</p>
+
+<h2>Noncompliant Code Example</h2>
+<pre>
+class HttpPrinter
+{
+  private string content;
+
+  public async void CallNetwork(string url) //Noncompliant
+  {
+    var client = new HttpClient();
+    var response = await client.GetAsync(url);
+    content = await response.Content.ReadAsStringAsync();
+  }
+
+  public async Task PrintContent(string url)  // works correctly if web request finishes in under 1 second, otherwise content will be null
+  {
+    CallNetwork(url);
+    await Task.Delay(1000);
+    Console.Write(content);
+  }
+}
+</pre>
+
+<h2>Compliant Solution</h2>
+
+<pre>
+class HttpPrinter
+{
+  private string content;
+
+  public async Task CallNetwork(string url)
+  {
+    var client = new HttpClient();
+    var response = await client.GetAsync(url);
+    content = await response.Content.ReadAsStringAsync();
+  }
+
+  public async Task PrintContent(string url)
+  {
+    await CallNetwork(url); // &lt;----- call changed here. If await is not added warning CS4014 will be triggered
+    await Task.Delay(1000);
+    Console.Write(content);
+  }
+}
+</pre>
--- a/src/Tests/SonarLint.RulingTest/Expected/CSharp/S3168.json
+++ b/src/Tests/SonarLint.RulingTest/Expected/CSharp/S3168.json
@@ -1,0 +1,94 @@
+{
+  "Version": "0.1",
+  "ToolInfo": {
+    "ToolName": "SonarLint.RulingTest",
+    "FileVersion": "1.0.0"
+  },
+  "Issues": [
+    {
+      "RuleId": "S3168",
+      "FullMessage": "Return \"Task\" instead.",
+      "Locations": [
+        {
+          "AnalysisTarget": {
+            "Uri": "\\akka.net-dev_8b00423b29\\src\\benchmark\\PingPong\\Program.cs",
+            "Region": {
+              "StartLine": 47,
+              "EndLine": 47,
+              "StartColumn": 0,
+              "EndColumn": 2147483647
+            }
+          }
+        }
+      ]
+    },
+    {
+      "RuleId": "S3168",
+      "FullMessage": "Return \"Task\" instead.",
+      "Locations": [
+        {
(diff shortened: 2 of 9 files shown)


src/Tests/SonarLint.RulingTest/Expected/CSharp/S3168.json
```diff
@@ -0,0 +1,94 @@
...
+        }
+      ]
+    },
+    {
+      "RuleId": "S3168",
+      "FullMessage": "Return \"Task\" instead.",
+      "Locations": [
+        {
+          "AnalysisTarget": {
+            "Uri": "\\Mvc-dev_b245996949\\test\\Microsoft.AspNet.Mvc.Core.Test\\DefaultActionSelectorTests.cs",
```
tvsonar: This is in a test. MSTest ignores `async void` test methods (they don't show up in the Test Explorer). NUnit handles them, but this is/will be removed in upcoming versions. XUnit doesn't handle it. [1](https://msdn.microsoft.com/en-us/magazine/dn818493.aspx)

src/Tests/SonarLint.RulingTest/Expected/CSharp/S3168.json
```diff
@@ -0,0 +1,94 @@
...
+        }
+      ]
+    },
+    {
+      "RuleId": "S3168",
+      "FullMessage": "Return \"Task\" instead.",
+      "Locations": [
+        {
+          "AnalysisTarget": {
+            "Uri": "\\akka.net-dev_8b00423b29\\src\\core\\Akka.Remote.TestKit\\Conductor.cs",
```
tvsonar: This seems like an event handler, but with non-standard signature.

src/Tests/SonarLint.RulingTest/Expected/CSharp/S3168.json
```diff
@@ -0,0 +1,94 @@
...
+    "FileVersion": "1.0.0"
+  },
+  "Issues": [
+    {
+      "RuleId": "S3168",
+      "FullMessage": "Return \"Task\" instead.",
+      "Locations": [
+        {
+          "AnalysisTarget": {
+            "Uri": "\\akka.net-dev_8b00423b29\\src\\benchmark\\PingPong\\Program.cs",
```
tvsonar: This is a false positive. It will work because after the call, the main thread is blocked and is waiting for a user input to exit the application.

tvsonar:
There are false positives reported by the rule. What do you think @dbolkensteyn, should we change the severity to Major, and the tags to suspicious?

dbolkensteyn:
@tvsonar Let's discuss this - I don't expect to get a PR when you're not yourself convinced about the implementation

src/Tests/SonarLint.RulingTest/Expected/CSharp/S3168.json
tvsonar: we are in the process of creating a rule to recognize this, it'll be part of https://jira.sonarsource.com/browse/MMF-124

tvsonar:
@dbolkensteyn I don't think that we can do a lot better then this. We've got #632, which is related and adds an improvement to event handler recognition.

src/Tests/SonarLint.RulingTest/Expected/CSharp/S3168.json
dbolkensteyn: This issue shouldn't appear in the ruling results, see https://github.com/SonarSource/sonarlint-vs/issues/664

src/SonarLint.Extra/Rules.Description/S3168.html
```diff
@@ -0,0 +1,52 @@
...
+{
+  private string content;
+  public async Task PrintContent(string url)  // works correctly if web request finishes in under 1 second
+  {
+    CallNetwork(url);
+    await Task.Delay(1000);
+    Console.Write(content);
+  }
+
+  public async void CallNetwork(string url) //Noncompliant
```
dbolkensteyn: I'd write this method before the `PrintContent` to make the code more readable

src/SonarLint.Extra/Rules.Description/S3168.html
```diff
@@ -0,0 +1,52 @@
...
+{
+  private string content;
+  public async Task PrintContent(string url)
+  {
+    await CallNetwork(url); // &lt;----- call changed here. If await is not added warning CS4014 will be triggered
+    await Task.Delay(1000);
+    Console.Write(content);
+  }
+
+  public async Task CallNetwork(string url) //Compliant
```
dbolkensteyn: I think we decided to not write `Compliant` anymore under the `Compliant Solution` section, as it's compliant by definition

src/SonarLint.Extra/Rules.Description/S3168.html
```diff
@@ -0,0 +1,52 @@
...
+<p>
+    This rule raises an issue when non-event handler methods are both async and void.
+</p>
+
+<h2>Noncompliant Code Example</h2>
+<pre>
+class HttpPrinter
+{
+  private string content;
+  public async Task PrintContent(string url)  // works correctly if web request finishes in under 1 second
```
dbolkensteyn: imo it'd be nice to explain what happens when that's not the case - I fail to understand what's wrong with this code by reading this RSPEC description, and probably will have to Google if I'd really like to know

dbolkensteyn:
LGTM & tested

PR #636, opened 2015-11-20
Issue #590 Improve test recognition

### LINKED ISSUE #590: Improve test recognition, and S2372 should not report in test projects
Original:
We'll need this feature for the property getter shouldn't throw rule. The current `Compilation.IsTest()` is not enough, we should check for the most often used test libraries. They are already collected here: #65.

NOTE: property getters can throw in a test assembly, and not in a test class. So we'd just need to improve the recognition of test assemblies.

tvsonar:
Fixes #590

### DIFF
--- a/src/SonarLint/Helpers/ProjectTypeHelper.cs
+++ b/src/SonarLint/Helpers/ProjectTypeHelper.cs
@@ -19,13 +19,12 @@
  */
 
 using Microsoft.CodeAnalysis;
+using System.Linq;
 
 namespace SonarLint.Helpers
 {
     public static class ProjectTypeHelper
     {
-        public const string TestAssemblyNamePattern = "test";
-
         public static bool IsTest(this Compilation compilation)
         {
             var assemblyName = string.Empty;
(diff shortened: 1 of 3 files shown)


src/SonarLint/Helpers/ProjectTypeHelper.cs
```diff
@@ -19,21 +19,23 @@
...
-        public const string TestAssemblyNamePattern = "test";
-
         public static bool IsTest(this Compilation compilation)
         {
-            var assemblyName = string.Empty;
-            if (compilation.AssemblyName != null)
-            {
-                assemblyName = compilation.AssemblyName;
-            }
-            return assemblyName.ToLowerInvariant().Contains(TestAssemblyNamePattern);
```
dbolkensteyn: @tvsonar is there a good reason to remove the check on the assembly name?

src/SonarLint/Rules/PropertyGetterWithThrowBase.cs
```diff
@@ -57,6 +57,11 @@ public override void Initialize(AnalysisContext context)
                 GeneratedCodeRecognizer,
                 cbc =>
                 {
+                    if(cbc.SemanticModel.Compilation.IsTest())
```
dbolkensteyn: @tvsonar this seems like a functional change to a specific rule, which isn't obvious from reading the initial issue https://github.com/SonarSource/sonarlint-vs/issues/590 - could you update the issue to make it explicit from its title what is being changed?

dbolkensteyn:
LGTM

PR #637, opened 2015-11-23
Issue #500 Attribute, EventArgs, and Exception type names should end …

### LINKED ISSUE #500: Attribute, EventArgs, and Exception type names should end with the type being extended
Deriving from the following classes should also be visible in the name:

```
class MyAttribute : Attribute { }
class MyEventArgs : EventArgs { }
class MyException : Exception { }
```

Maybe there are some more examples I do not recall at the moment

tvsonar:
…with the type being extended

Fixes #500

### DIFF
--- a/src/SonarLint.CSharp/Rules/FrameworkTypeNaming.cs
+++ b/src/SonarLint.CSharp/Rules/FrameworkTypeNaming.cs
@@ -1,0 +1,109 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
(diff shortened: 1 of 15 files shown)


src/Tests/SonarLint.RulingTest/Expected/CSharp/S3376.json
```diff
@@ -0,0 +1,162 @@
...
+    "FileVersion": "1.0.0"
+  },
+  "Issues": [
+    {
+      "RuleId": "S3376",
+      "FullMessage": "Make this class name end with \"Exception\".",
+      "Locations": [
+        {
+          "AnalysisTarget": {
+            "Uri": "\\akka.net-dev_8b00423b29\\src\\core\\Akka.Remote.TestKit\\BarrierCoordinator.cs",
```
dbolkensteyn: Does this rule really matter in tests? This does seems like noise

src/SonarLint.CSharp/Rules/FrameworkTypeNaming.cs
```diff
@@ -0,0 +1,98 @@
...
+                        return;
+                    }
+
+                    var baseType = FrameworkTypes.FirstOrDefault(ft => symbol.BaseType.ToDisplayString().Equals(ft, System.StringComparison.InvariantCulture));
+                    if (baseType == null)
+                    {
+                        return;
+                    }
+
+                    var baseTypeName = baseType.Substring("System.".Length);
```
dbolkensteyn: this isn't great - it means we can't simply add any type to `FrameworkTypes` to expand this rule

dbolkensteyn:
LGTM, tested

src/SonarLint.CSharp/Rules/FrameworkTypeNaming.cs
```diff
@@ -0,0 +1,98 @@
...
+                {
+                    var classDeclaration = (ClassDeclarationSyntax)c.Node;
+                    var symbol = c.SemanticModel.GetDeclaredSymbol(classDeclaration);
+                    if (symbol == null ||
+                        symbol.BaseType == null)
+                    {
+                        return;
+                    }
+
+                    var baseType = FrameworkTypes.FirstOrDefault(ft => symbol.BaseType.ToDisplayString().Equals(ft, System.StringComparison.InvariantCulture));
```
dbolkensteyn: if you on purpose do not follow the type hierarchy here, then it's worth to add a unit test for it

```
class RuntimeException : Exception {}
class MyOwnErrror : RuntimeException {} // This isn't detected - was this choice made on purpose because maybe RuntimeException is coming from a 3rd party assembly and you're just following their convention, or was this use case simply forgotten?
```

PR #638, opened 2015-11-23
Issue #623 Replace rule S124 (regexp in comments) by 2 rules to detec…

### LINKED ISSUE #623: Replace rule S124 (regexp in comments) by 2 rules to detect FIXME and TODO comments
Rule template S124 will not work when analyzers are executed as part of MSBuild, and do not provide value out-of-the-box for end-users, as they have to first manually create instances of that rule.

In most cases, TODO and FIXME comments are the one we'd like to track.
  FIXME: https://jira.sonarsource.com/browse/RSPEC-1134
  TODO: https://jira.sonarsource.com/browse/RSPEC-1135

tvsonar:
…t FIXME and TODO comments

Fixes #623


src/SonarLint.CSharp/Rules/CommentRegularExpression.cs
```diff
@@ -24,86 +24,31 @@
...
-using SonarLint.Common.Sqale;
 using SonarLint.Helpers;
 
 namespace SonarLint.Rules.CSharp
 {
-    [DiagnosticAnalyzer(LanguageNames.CSharp)]
-    [NoSqaleRemediation]
-    [Rule(TemplateDiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
-    public class CommentRegularExpression : DiagnosticAnalyzer, IRuleTemplate<CommentRegularExpression.CommentRegularExpressionRule>
+    public abstract class CommentRegularExpressionBase : DiagnosticAnalyzer
```
tvsonar: This file should be renamed to match the class name.

dbolkensteyn:
LGTM, nice to see how much code this removes :)

However, IMO it'd be great to report on the exact line where the TODO or FIXME occurs, e.g. in:
```
            /*

            hehe

            TODO << This one should be non compliant ideally, so that double-clicking from the error list brings you directly to the right location

            todo

            */
```

tvsonar:
@dbolkensteyn Not a bad idea. I fixed the precise error location. As a consequence, we are now reporting two issues in you sample input.

PR #639, opened 2015-11-23
Issue #360 "ConfigureAwait(false)" should be used

### LINKED ISSUE #360: "ConfigureAwait(false)" should be used
When awaiting a `Task` we may specify how to continue execution after the `Task` has finished. We can either switch back to the original thread, or just use any thread available. Switching back to the original thread is usually required if the initial call happened on the UI thread. However in a library, we shouldn't depend on the thread which we are running on, thus we shouldn't force a context switching back to the original thread. Thus, we should call `Task.ConfigureAwait(false)` on the awaited task.

Also, as with `.ConfigureAwait(false)` we don't force switching back to the original thread, we reduce the risk of running into a deadlock. Because the original thread might be blocked, and waiting for the result of the asynchronous task which can in turn wait for a context switch to the blocked thread. (http://blog.ciber.no/2014/05/19/using-task-configureawaitfalse-to-prevent-deadlocks-in-async-code/)

https://jira.sonarsource.com/browse/RSPEC-3216

tvsonar:
Fixes #360

### DIFF
--- a/src/SonarLint.CSharp/Rules/TaskConfigureAwait.cs
+++ b/src/SonarLint.CSharp/Rules/TaskConfigureAwait.cs
@@ -1,0 +1,96 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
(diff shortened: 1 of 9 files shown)


src/SonarLint.Extra/Rules.Description/S3216.html
```diff
@@ -0,0 +1,20 @@
+<p>
+    After an <code>await</code>ed <code>Task</code> has executed, you can continue execution in the original, calling
+    thread or any arbitrary thread. Unless the rest of the code needs the context from which the <oode>Task</oode> was
+    spawned, <code>Task.ConfigureAwait(false)</code> should be used to keep execution in the <code>Task</code> thread
+    to avoid the need of context switching and the possibility of deadlocks.
+</p>
+<p>
+    This rule raises an issue when code in a class library <code>await</code>s a <code>Task</code> and continues
+    execution in the main thread.
```
dbolkensteyn: the previous paragraph talks about "original calling thread", not "main"

src/SonarLint.CSharp/Rules/TaskConfigureAwait.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c =>
+                {
+                    if (c.SemanticModel.Compilation.IsTest() ||
+                        c.SemanticModel.Compilation.Options.OutputKind != OutputKind.DynamicallyLinkedLibrary)
+                    {
+                        //this rule makes sense in libraries
```
dbolkensteyn: "this rule only makes sense" matches the code better

dbolkensteyn:
LGTM, tested

PR #640, opened 2015-11-23
Issue #177 Overflow checking should not be disabled for "Enumerable.Sum"

### LINKED ISSUE #177: Overflow checking should not be disabled for "Enumerable.Sum"
http://jira.sonarsource.com/browse/RSPEC-2291

tvsonar:
Fixes #177

### DIFF
--- a/src/SonarLint.CSharp/Rules/EnumerableSumInUnchecked.cs
+++ b/src/SonarLint.CSharp/Rules/EnumerableSumInUnchecked.cs
@@ -1,0 +1,153 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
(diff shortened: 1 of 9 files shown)


src/SonarLint.CSharp/Rules/EnumerableSumInUnchecked.cs
```diff
@@ -0,0 +1,153 @@
...
+                {
+                    return false;
+                }
+                returnType = nullableType.TypeArguments[0];
+            }
+
+            return AllowedTypes.Any(type => returnType.SpecialType == type);
+        }
+
+        private static readonly SpecialType[] AllowedTypes =
```
dbolkensteyn: seems more like the list of disallowed types, i.e. on which you can't call `Sum()`

dbolkensteyn:
LGTM, tested

PR #641, opened 2015-11-24
"base.Equals" should not be used to check for reference equality in "Equals" override

### LINKED ISSUE #621: "base.Equals" should not be used to check for reference equality in "Equals" override
RSPEC: https://jira.sonarsource.com/browse/RSPEC-3397

When overriding `Equals` you can add a guard condition to return fast like this:

    public override bool Equals(object obj)
    {
        if (base.Equals(obj))              <------------
        {
            return true;
        }

        NameNode otherName = obj as NameNode;
        return otherName != null && otherName.Name == this.Name;
    }

However, when the `base.Equals` is not `object.Equals` than this is a bug, because we are not checking the current classes other fields.

tvsonar:
Fixes #621

### DIFF
--- a/src/SonarLint.CSharp/Rules/GuardConditionOnEqualsOverride.cs
+++ b/src/SonarLint.CSharp/Rules/GuardConditionOnEqualsOverride.cs
@@ -19,7 +19,6 @@
  */
 
 using System.Collections.Immutable;
-using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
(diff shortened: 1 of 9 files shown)


src/SonarLint.CSharp/Rules/GetHashCodeEqualsOverride.cs
```diff
@@ -56,7 +56,9 @@ public class GetHashCodeEqualsOverride : DiagnosticAnalyzer
```
dbolkensteyn: @tvsonar in the rule title above, there are 2 spaces `Classes directly  extending \"Object\" should ` between `directly` and `extending`

dbolkensteyn:
LGTM - also see the comments on this particular commit https://github.com/tvsonar/sonarlint-vs/commit/a6d1b4f0ad1f7de1e8bbeabf6781835889086e3a

dbolkensteyn:
Be careful to re-run the tests after the merge conflict is solved, and before merging into the master.

tvsonar:
@dbolkensteyn How about the following for title:
"base.Equals" should not be used to check for reference equality in "Equals" if "base" is not "object"

dbolkensteyn:
LGTM, Tested

PR #643, opened 2015-11-25
Issue #362 Parallel collections should not be maintained

### LINKED ISSUE #362: Parallel collections should not be maintained
https://jira.sonarsource.com/browse/RSPEC-3221

http://codeblog.jonskeet.uk/2014/06/03/anti-pattern-parallel-collections/

tvsonar:
Fixes #362

### DIFF
--- a/src/SonarLint.CSharp/Rules/ParallelCollections.cs
+++ b/src/SonarLint.CSharp/Rules/ParallelCollections.cs
@@ -1,0 +1,194 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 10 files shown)


src/SonarLint.CSharp/Rules/ParallelCollections.cs
```diff
@@ -0,0 +1,194 @@
...
+        internal const string Title = "Parallel collections should not be maintained";
+        internal const string Description =
+            "Using parallel collections or arrays rather than classes to hold and process related pieces of data is " +
+            "an antipattern. Instead, define a type for the entity the arrays represent and use an array or " +
+            "collection of that type.";
+        internal const string MessageFormat =
+            "Create a type that holds an element from {0}, and merge these parallel collections indexed by the same value.";
+        internal const string Category = Constants.SonarLint;
+        internal const Severity RuleSeverity = Severity.Minor;
+        internal const bool IsActivatedByDefault = false;
```
tvsonar: The rule raises false positives, so we can't enable it by default.

dbolkensteyn: @tvsonar Why are we adding this rule to the product then?

tvsonar: @dbolkensteyn we add it to the product because it catches design issues. we disable it because it's noisy. 

Do you suggest to remove all disabled rules?

dbolkensteyn:

PR #648, opened 2015-11-25
Issue #627 Group rules by meaningful categories

### LINKED ISSUE #627: Group rules by meaningful categories
The categories commonly used are:
image

SonarLint and SonarQube is not good category name as they match the name of the ruleset. If one want to exclude the complete set, it can be nicely done using the ruleset view.  
The rules should be group into categories following the already established categories.

tvsonar:
Fixes #627


dbolkensteyn:
LGTM

PR #650, opened 2015-11-27
Issue #363 Simplify the LINQ call chain

### LINKED ISSUE #363: Simplify the LINQ call chain
Two cases are missing from the implementation of S2971:
* ToArray/ToList calls are unneccessary in the middle of a call chain
* Count() can be simplified to Count.

https://jira.sonarsource.com/browse/RSPEC-2971

tvsonar:
Fixes #363

### DIFF
--- a/src/SonarLint.CSharp/Rules/IndexOfCheckAgainstZero.cs
+++ b/src/SonarLint.CSharp/Rules/IndexOfCheckAgainstZero.cs
@@ -19,7 +19,6 @@
  */
 
 using System.Collections.Immutable;
-using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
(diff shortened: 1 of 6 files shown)


src/SonarLint.CSharp/Rules/IndexOfCheckAgainstZero.cs
```diff
@@ -112,7 +112,7 @@ private static bool IsIndexOfCall(ExpressionSyntax call, SemanticModel semanticM
 
         internal static bool DerivesOrImplements(INamedTypeSymbol type, ITypeSymbol[] possibleTypes)
```
dbolkensteyn: @tvsonar I guess it'd be better to move such helper methods to dedicated classes (it can be done progressively - doesn't have to be all at once) - it is strange to see IndexOfCheckAgainstZero being modified by this PR.

dbolkensteyn:
Tested

dbolkensteyn:
LGTM

PR #651, opened 2015-11-27
Issue #509 S1172 shouldn't report on extension methods' "this" parameter

### LINKED ISSUE #509: S1172 shouldn't report on extension methods' "this" parameter
SonarLint complains about 'this' parameters not being used in extension methods. 

For example it's quite common practice for HtmlHelper extension methods to not use the object instance as the helper method is only attached to the object as a convenience.

tvsonar:
Fixes #509


dbolkensteyn:
LGTM

PR #652, opened 2015-11-27
Issue #630 S101 should not trigger for static classes who's name ends…

### LINKED ISSUE #630: S101 should not trigger for static classes who's name ends in Extensions
The common naming pattern for Extension methods is to create a static class, have it start with the type it wants to extend followed by "Extensions", taking this one step further, a set of methods extending an interface thus starts with INameExtensions. Even StyleCop doesn't take offence on this construct.

Side note, the error message is horrible, as it takes the whole regex into the message and can't tell me what to do to if this naming issue occurs, I need to figure out what part of the expression causes the violation.

I vote to disable this rule by default and to let people use StyleCop instead for naming related items, it is far more advanced and there is no need to have two competing rulesets from two open source projects doing the exact same thing.

tvsonar:
… in Extensions

Fixes #630

### DIFF
--- a/src/SonarLint.CSharp/Rules/ClassName.cs
+++ b/src/SonarLint.CSharp/Rules/ClassName.cs
@@ -27,6 +27,7 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using SonarLint.Common;
 using SonarLint.Common.Sqale;
 using SonarLint.Helpers;
+using System.Globalization;
 
 namespace SonarLint.Rules.CSharp
 {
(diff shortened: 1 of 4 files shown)


src/SonarLint.CSharp/Rules/ClassName.cs
```diff
@@ -74,5 +82,10 @@ public override void Initialize(AnalysisContext context)
                 },
                 SyntaxKind.ClassDeclaration);
         }
+
+        private static bool IsSymbolExtensionClass(ISymbol symbol)
```
dbolkensteyn: @tvsonar IMO it's better to update the regex rather than add logic here. Indeed, currently I can call my class `foo__barExtensions` and this rule wouldn't pick it up - because it doesn't check *all* class names anymore.

tvsonar:
@dbolkensteyn We received complaints that it's hard to know what's wrong with the name if there's an issue. We include the regex in the issue message, but we can't really say which part of the regex is violated. If we add this exception to the regex too, then the issue message will become even more complex, and practically useless, and annoying. WDYT?

Another option would be to simplify the regex to the one used in Java (`^[A-Z][a-zA-Z0-9]*$`). And we could check this on the Extensions classes as well.

src/SonarLint.CSharp/Rules/ClassName.cs
```diff
@@ -55,7 +56,7 @@ public class ClassName : DiagnosticAnalyzer
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Rule); } }
 
-        private const string DefaultValueConvention = "^([A-HJ-Z][a-zA-Z0-9]+|I[a-z0-9][a-zA-Z0-9]*)$";
+        private const string DefaultValueConvention = "^([A-HJ-Z][a-zA-Z0-9]+|I[a-z0-9][a-zA-Z0-9]*|[A-Z][a-zA-Z0-9]+Extensions)$";
```
tvsonar: This needs to go into the RSPEC.

dbolkensteyn: @tvsonar feel free to update RSPEC

dbolkensteyn:
LGTM

PR #654, opened 2015-11-30
Method overloads with default parameter values should not overlap

### LINKED ISSUE #653: Method overloads with default parameter values should not overlap
Replaces http://jira.sonarsource.com/browse/LINTVS-180

```
void Print(string[] messages) {...}  // Noncompliant
void Print(string[] messages, string delimiter = "n") {...}
```

When you call `Print(new []\{"abc"\});` which method will be called? The first one, but it is not obvious. So we should avoid these overloads with default parameter values. 

Note there is this rule: https://msdn.microsoft.com/en-us/library/ms182135.aspx, CA1026, saying that optional parameters should be avoided altogether because of CLS (non-)compliance. But I wouldn't go this far as optional parameters are really useful if you use named parameters in the calls: http://codeblog.jonskeet.uk/2015/06/03/backwards-compatibility-is-still-hard/

R# ID: *MethodOverloadWithOptionalParameter*

This rule could be covered by the http://jira.sonarsource.com/browse/RSPEC-3220, which covers overloads with `params` argument.

Implements https://jira.sonarsource.com/browse/RSPEC-3427

tvsonar:
Fixes #653

### DIFF
--- a/src/SonarLint.CSharp/Rules/MethodOverloadOptionalParameter.cs
+++ b/src/SonarLint.CSharp/Rules/MethodOverloadOptionalParameter.cs
@@ -1,0 +1,180 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarLint.Common;
+using SonarLint.Common.Sqale;
+using SonarLint.Helpers;
+using System.Linq;
+using System.Collections.Generic;
+using SonarLint.Rules.Common;
+using System.IO;
+
+namespace SonarLint.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [SqaleConstantRemediation("10min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.Understandability)]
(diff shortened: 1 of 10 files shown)


src/SonarLint.CSharp/Rules/MethodOverloadOptionalParameter.cs
```diff
@@ -0,0 +1,181 @@
...
+                            continue;
+                        }
+
+                        var defaultCanBeUsed =
+                            IsMoreParameterAvailableInConflicting(hidingInfo) ||
+                            !MethodsUsingSameParameterNames(hidingInfo);
+
+                        var isOtherFile = syntax.SyntaxTree.FilePath != hidingMethodSyntax.SyntaxTree.FilePath;
+
+                        c.ReportDiagnostic(Diagnostic.Create(Rule, syntax.GetLocation(),
```
tvsonar: Should only report in non generated files

src/SonarLint.CSharp/Rules/MethodOverloadOptionalParameter.cs
```diff
@@ -0,0 +1,181 @@
...
+                {
+                    var methodSymbol = c.Symbol as IMethodSymbol;
+                    if (methodSymbol == null ||
+                        PublicMethodWithMultidimensionalArrayBase.IsMethodOverrideOrInterfaceImplementation(methodSymbol) ||
+                        !methodSymbol.Parameters.Any(p => p.IsOptional))
+                    {
+                        return;
+                    }
+
+                    var methods = methodSymbol.ContainingType.GetMembers()
```
dbolkensteyn: why not `GetMembers(methodSymbol.Name)`

src/SonarLint.CSharp/Rules/MethodOverloadOptionalParameter.cs
```diff
@@ -0,0 +1,181 @@
...
+                        .OfType<IMethodSymbol>();
+
+                    var hidingInfos = new List<ParameterHidingMethodInfo>();
+
+                    var matchingNamedMethods = methods
+                        .Where(m => m.Name == methodSymbol.Name)
```
dbolkensteyn: then this one could be dropped

src/SonarLint.CSharp/Rules/MethodOverloadOptionalParameter.cs
```diff
@@ -0,0 +1,181 @@
...
+        {
+            return hidingInfo.HiddenMethod.Parameters.IndexOf(hidingInfo.ParameterToReportOn) <
+                hidingInfo.HiddenMethod.Parameters.Count() - 1;
+        }
+
+        private static bool IsMethodHidingOriginal(IMethodSymbol candidateHidingMethod, IMethodSymbol method)
+        {
+            for (int i = 0; i < candidateHidingMethod.Parameters.Length; i++)
+            {
+                if (!candidateHidingMethod.Parameters[i].Type.Equals(method.Parameters[i].Type))
```
dbolkensteyn: iterating over parallel collections .. :)

dbolkensteyn:
LGTM, tested

PR #656, opened 2015-12-01
Method calls should not resolve ambiguously to overloads with "params"

### LINKED ISSUE #58: Method calls should not resolve ambiguously to overloads with "params"
Replaces http://jira.sonarsource.com/browse/LINTVS-99

Minor
```
public class MyClass  // Noncompliant; there are methods with overlapping definitions
{
  public void M(int i, int j, object k) { }
  public void M(params int[] ints) { }
}

new MyClass().M(1, 2, 3); //resolves to the params version, but did you mean that?
```

In this case `3` would also be assignable to `object k`, but the compiler chooses the other method.

On method overload resolution see:
http://stackoverflow.com/questions/5173339/how-does-the-method-overload-resolution-system-decide-which-method-to-call-when

R# ID: *PossiblyMistakenUseOfParamsMethod*

Implements http://jira.sonarsource.com/browse/RSPEC-3220

tvsonar:
Fixes #58

### DIFF
--- a/src/SonarLint.CSharp/Rules/ArrayCovariance.cs
+++ b/src/SonarLint.CSharp/Rules/ArrayCovariance.cs
@@ -136,52 +136,7 @@ namespace SonarLint.Rules.CSharp
                 SyntaxKind.CastExpression);
         }
 
-        // todo: this should come from the Roslyn API (https://github.com/dotnet/roslyn/issues/9)
-        internal class MethodParameterLookup
-        {
-            private readonly InvocationExpressionSyntax invocation;
-            private readonly IMethodSymbol methodSymbol;
-
-            public MethodParameterLookup(InvocationExpressionSyntax invocation, SemanticModel semanticModel)
-            {
-                this.invocation = invocation;
-                methodSymbol = semanticModel.GetSymbolInfo(invocation).Symbol as IMethodSymbol;
-            }
-
-            public IMethodSymbol MethodSymbol
-            {
-                get
-                {
-                    return methodSymbol;
-                }
-            }
-
-            public IParameterSymbol GetParameterSymbol(ArgumentSyntax argument)
-            {
-                if (!invocation.ArgumentList.Arguments.Contains(argument) ||
-                    methodSymbol == null)
-                {
-                    return null;
-                }
-
-                if (argument.NameColon != null)
-                {
-                    return methodSymbol.Parameters
-                        .FirstOrDefault(symbol => symbol.Name == argument.NameColon.Name.Identifier.ValueText);
-                }
-
-                var argumentIndex = invocation.ArgumentList.Arguments.IndexOf(argument);
-                var parameterIndex = argumentIndex;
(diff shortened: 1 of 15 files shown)


src/SonarLint.CSharp/Rules/ArrayCovariance.cs
```diff
@@ -156,30 +156,36 @@ public IMethodSymbol MethodSymbol
                 }
             }
 
-            public IParameterSymbol GetParameterSymbol(ArgumentSyntax argument)
+            public static IParameterSymbol GetParameterSymbol(ArgumentSyntax argument, ArgumentListSyntax argumentList, IMethodSymbol method)
```
dbolkensteyn: @tvsonar same stuff about helpers

src/SonarLint.CSharp/Rules/InvocationResolvesToOverrideWithParams.cs
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        private static bool ArgumentsMatchParameters(ArgumentListSyntax argumentList, List<INamedTypeSymbol> argumentTypes,
+            IMethodSymbol possibleOtherMethod)
+        {
+            var matchedParameters = new List<IParameterSymbol>();
+            for (int i = 0; i < argumentList.Arguments.Count; i++)
+            {
+                var argument = argumentList.Arguments[i];
+                var argumentType = argumentTypes[i];
```
dbolkensteyn: @tvsonar seems like you're iterating over 2 collections in parallel ;)

src/SonarLint.CSharp/Rules/InvocationResolvesToOverrideWithParams.cs
```diff
@@ -0,0 +1,175 @@
...
+                if (argumentType == null)
+                {
+                    if (!parameter.Type.IsReferenceType)
+                    {
+                        return false;
+                    }
+                }
+                else
+                {
+                    if (!IndexOfCheckAgainstZero.DerivesOrImplements(argumentType, new[] { parameter.Type }))
```
dbolkensteyn: @tvsonar a false negative, bit of a corner case: it'd probably be better to check for the presence of an implicit conversion as well, as in:

```
public static void Test(int foo, params object[] p)
{
    Console.WriteLine("test1");
}

public static void Test(double foo, object p1)
{
    Console.WriteLine("test2");
}

static void Main(string[] args)
{
    Test(42, null); // Should be Noncompliant
}
```

dbolkensteyn:
LGTM, tested

src/SonarLint.CSharp/Rules/InvocationResolvesToOverrideWithParams.cs
tvsonar: @dbolkensteyn yeah I know. It's bad design, but I probably won't fix it.

PR #657, opened 2015-12-01
Adding InternalsVisibleTo attributes to remove noise coming from S2339

tvsonar:


src/SonarLint.CSharp/Properties/AssemblyInfo.cs
```diff
@@ -19,7 +19,18 @@
...
+using System.Runtime.CompilerServices;
 
 [assembly: AssemblyTitle("SonarLint.CSharp")]
 [assembly: AssemblyProduct("SonarLint.CSharp")]
-[assembly: AssemblyDescription("")]
\ No newline at end of file
+[assembly: AssemblyDescription("")]
+
+#if !SIGNED
+[assembly: InternalsVisibleTo("SonarLint.RulingTest")]
```
tvsonar: This should be removed when we get rid of the template rule.

src/SonarLint.CSharp/Properties/AssemblyInfo.cs
```diff
@@ -19,7 +19,18 @@
...
+[assembly: InternalsVisibleTo("SonarLint.UnitTest")]
+[assembly: InternalsVisibleTo("SonarLint.Runner")]
```
tvsonar: This should also be removed. It shows bad design.

dbolkensteyn:
I'll +1 this PR if you can manage to get rid of conditional compilation.

Conditional compilation is evil because:
1. The CI build will run its tests only on one of the two branches
2. All our manual tests during the sprint will be done on that branch as well
3. But at the end of the sprint, we release something that's different and which has not been tested

So IMO we can't introduce conditional compilation without changing our CI build at the same time to test both branches.

dbolkensteyn:

tvsonar:
@dbolkensteyn This is a purely technical PR, so it doesn't have to show up in a sprints release notes. I've created a ticket for it nonetheless: https://jira.sonarsource.com/browse/SLVS-754

tvsonar:
I'm closing this PR, which is superseded by https://github.com/SonarSource-VisualStudio/sonaranalyzer-csharp/pull/899.

A solution based on non-conditional compilation would be ideal, but as it is today we are not testing signed assemblies automatically. So conditional compilation would result in the same inconvenience.

PR #741, opened 2015-12-31
reload SonarLint.xml settings when the file changes, and debugging improvements

kevinpdpath:
I wanted to be able to play with the settings in SonarLint.xml in my project, but changes to the file were not taking affect. This is because the setting had already been loaded into ProcessedAnalyzers, and the code does not reload them. 

I made changes to track the timestamp of SonarLint.xml so that it would reload the settings if the file has changed since it was loaded last. 

NOTE: Existing warnings in a code file do not appear to be refreshed by Visual Studio until the code file is modified, so to verify this you will need to modify SonarLint.xml and then modify the code file being analyzed.

I also added some debugging improvements in the process of working on this. The csproj changes allow you run the vsix project and load symbols so that breakpoints can be hit. The ToString() overrides make it easier to see the values of the parameters in debug windows.

### DIFF
--- a/src/SonarLint.Vsix/SonarLint.Vsix.csproj
+++ b/src/SonarLint.Vsix/SonarLint.Vsix.csproj
@@ -24,6 +24,7 @@
     <IncludeDebugSymbolsInLocalVSIXDeployment>false</IncludeDebugSymbolsInLocalVSIXDeployment>
     <CopyBuildOutputToOutputDirectory>false</CopyBuildOutputToOutputDirectory>
     <CopyOutputSymbolsToOutputDirectory>false</CopyOutputSymbolsToOutputDirectory>
+    <VSSDKTargetPlatformRegRootSuffix>Roslyn</VSSDKTargetPlatformRegRootSuffix>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <DebugSymbols>true</DebugSymbols>
@@ -45,6 +46,11 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <CodeAnalysisRuleSet>..\SonarLint.ruleset</CodeAnalysisRuleSet>
+  </PropertyGroup>
+  <PropertyGroup>
+    <StartAction>Program</StartAction>
+    <StartProgram>$(DevEnvDir)devenv.exe</StartProgram>
+    <StartArguments>/rootsuffix Roslyn</StartArguments>
   </PropertyGroup>
   <ItemGroup>
     <None Include="app.config" />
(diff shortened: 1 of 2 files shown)


src/SonarLint/Helpers/ParameterLoader.cs
```diff
@@ -96,6 +106,20 @@ public static ImmutableList<RuleParameterValues> ParseParameters(XContainer xml)
             }
 
             var filePath = additionalFile.Path;
+            DateTime lastFileUpdate = File.GetLastWriteTime(filePath);
```
dbolkensteyn: @kevinpdpath I don't like this unconditional IO call. This `SetParameterValues()` is called quite often from here: https://github.com/SonarSource/sonarlint-vs/blob/dfa56b11a37916e19fd734bcb5bbb73330cab3a6/src/SonarLint/Helpers/ParameteredDiagnosticAnalyzer.cs#L35

IO is slow by definition

Also, `GetLastWriteTimeUtc()` might be a slightly better option which wouldn't reload the same parameters in the unlikely use case that the locale gets changed while Visual Studio is running

src/SonarLint.Vsix/SonarLint.Vsix.csproj
```diff
@@ -24,6 +24,7 @@
     <IncludeDebugSymbolsInLocalVSIXDeployment>false</IncludeDebugSymbolsInLocalVSIXDeployment>
     <CopyBuildOutputToOutputDirectory>false</CopyBuildOutputToOutputDirectory>
     <CopyOutputSymbolsToOutputDirectory>false</CopyOutputSymbolsToOutputDirectory>
+    <VSSDKTargetPlatformRegRootSuffix>Roslyn</VSSDKTargetPlatformRegRootSuffix>
```
dbolkensteyn: @kevinpdpath this looks like a great change. I'll merge it, but with `Exp`.

It will simplify some of the explanations which currently are here: https://github.com/SonarSource/sonarlint-vs/blob/5ffc900f94ac9f08fe12efa98b20408a6e0b66af/its/README.md

dbolkensteyn: it also seems that we can set the VSIX project as the start-up project directly in the *.sln file: http://stackoverflow.com/questions/694730/why-is-set-as-startup-option-stored-in-the-suo-file-and-not-the-sln-file

dbolkensteyn: I've just opened this PR containing your change: https://github.com/SonarSource/sonarlint-vs/pull/742

dbolkensteyn:
As you say, this change is mainly useful to play around with the `SonarLint.xml` file, but won't be really usable as changes to that file are not picked up until you edit your code, which isn't ideal.

At the same time, we're working on a new version of SonarLint for Visual Studio, which will synchronize its quality profile (and rule parameters) with the SonarQube server.

So I don't want to risk to slow down the current SonarLint VS (see the comments relating to the usage of the `File.GetLastWrite()` method) for something which might not be that useful in the end.

I'm therefore closing ths PR. Feel free to re-open one if you find a nice way to reduce the IO operations.

I've merged your other change in https://github.com/SonarSource/sonarlint-vs/pull/742

PR #749, opened 2016-01-06
Fixes #714 S3220 fails with 'Sequence contains no element'

### LINKED ISSUE #714: S3220 fails with 'Sequence contains no element'
unnamed

While waiting for the fix to ship, users who are bothered by these messages can either disable the rule AD0001 to silently ignore all analyzer bugs, or disable this specific analyzer altogether.

The following code can be used to reproduce the issue:
```
public class MyClass
{
    public void Format(string a, params object[] b) { }
    public void Format() { } // The presence of this method causes the issue
}

static class Program
{
    static void Main()
    {
        MyClass myClass = new MyClass();
        myClass.Format("", null, null);
    }
}
```

dbolkensteyn:


its/expected/akka.net/Samples.Cluster.ConsistentHashRouting-{29A08A09-83F6-48D4-A9AE-B4AE314069C4}-S3220.json
```diff
@@ -12,6 +12,36 @@
 {
 "analysisTarget":  [
 {
+"uri":  "akka.net\\src\\examples\\Cluster\\Routing\\Samples.Cluster.ConsistentHashRouting\\BackendActor.cs",
```
dbolkensteyn: This is actually a false positive, see #750

PR #751, opened 2016-01-07
Fixes #715 S3240 (ConditionalSimplification) fails with NPE when both true & false types are null or actions

### LINKED ISSUE #715: S3240 (ConditionalSimplification) fails with NPE when both true & false types are null or actions
Somewhere in the middle of the Roslyn project analysis (at commit f47bfec42844d8f4f408fa313d09adab8cacd4d9), this analyzer fails with:

```
CSC : error AD0001: Analyzer 'SonarLint.Rules.CSharp.ConditionalSimplification' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'. [C:\projects\nemo-roslyn\roslyn\src\Compilers\Core\Portable\CodeAnalysis.csproj]
```

This code reproduces the issue:
```
if (false)
{
  return null;
}
else
{
  return (null);
}
```

or, more likely, with action literals, which also seem to have a `null` type:
```
Action myAction;
if (false)
{
  myAction = () => { };
}
else
{
  myAction = () => { Console.WriteLine(); };
}
```

dbolkensteyn:


src/Tests/SonarLint.UnitTest/TestCases/ConditionalSimplification.Fixed.cs
```diff
@@ -86,6 +86,13 @@ int Test(object a, object b, object y, bool condition)
             {
                 elem = new NonExistendType();
             }
+
+            elem = false ? null : (null);
+
+            Action myAction;
+            myAction = false ? () => { }
```
dbolkensteyn: this doesn't look nice, but probably the issue shouldn't have been triggered on this construct in the first place

Issue #739, opened 2015-12-29
Ruling IT: Enable AD0001 at Error level


dbolkensteyn:
See #664 for details, this little change had to be extracted into a separate ticket because it depends on the bogus analyzers to be fixed first, to avoid breaking the build.

PR #753, opened 2016-01-08
#739 Ruling IT: Enable AD0001 at Error level

dbolkensteyn:


its/SonarLint.Testing.ImportBefore.targets
```diff
@@ -21,7 +21,7 @@
 
       <!-- Don't fail the build -->
       <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
-      <WarningsAsErrors></WarningsAsErrors> <!-- TODO: Add AD0001 -->
+      <WarningsAsErrors></WarningsAsErrors>
```
dbolkensteyn: FYI adding `AD0001` here didn't work: The build of Akka.NET on its F# projects would fail with a message stating that the IDs here must be integers

It's anyway better IMO to place all rules and severities in the .ruleset, as this also enforces the rule to be enabled in the first place.

PR #754, opened 2016-01-11
SLVS-664 Don't block other builds while regression test is running

dbolkensteyn:


its/SonarLint.Testing.ImportBefore.targets
```diff
@@ -2,9 +2,15 @@
   <Target Name="OverrideRoslynCodeAnalysisPropertiesWithSonarLint"
           AfterTargets="ResolveCodeAnalysisRuleSet"
           BeforeTargets="CoreCompile">
-    <Error Text="Not running a SonarLint test build? Or forgot to build SonarLint first? Delete this file: $(MSBuildThisFileFullPath) to remove this error" Condition="!Exists('$(MSBuildStartupDirectory)\binaries\SonarLint.dll')" />
```
dbolkensteyn: Because of this error, all builds, even from the Visual Studio UI, will fail while the regression test is running... which is annoying

tvsonar:
It seems possible to get rid of the targets file copying to "MSBuild\14.0\Microsoft.Common.targets\ImportBefore", with registering before or after hooks for the solution (before.[Solution Name].sln.targets). But it seems quite messy, because you need to somehow get all the projects within the solution file to register the actual build task for all of them. And the only sample I found was [this](https://gist.github.com/3F/a77129e3978841241927), which seems to parse the solution file.

its/SonarLint.Testing.ImportBefore.targets
tvsonar: If we are running the regression test, and in the mean time we build any other project, this target will be picked up. So we are going to change the analyzers of the project and some properties. Changing the analyzers will change the number of warnings/errors reported, which is not good. The build fails if there is no "output" folder at the solution level, because we are trying to write to "$(MSBuildStartupDirectory)\output\" folder.

I think we either need to move this target out of the "MSBuild\14.0\Microsoft.Common.targets\" folder to not affect other builds, or leave the error message in place, and not run any other builds during regression testing.

its/SonarLint.Testing.ImportBefore.targets
```diff
@@ -2,9 +2,15 @@
...
 
-    <ItemGroup>
+    <PropertyGroup>
+      <SonarLintRegressionTestRunning Condition="Exists('$(MSBuildStartupDirectory)\binaries\SonarLint.dll')">true</SonarLintRegressionTestRunning>
+      <SonarLintRegressionTestRunning Condition="$(SonarLintRegressionTestRunning) == ''">false</SonarLintRegressionTestRunning>
+    </PropertyGroup>
+
+    <Warning Text="Not running a SonarLint test build? Or forgot to build SonarLint first? Delete this file: $(MSBuildThisFileFullPath) to remove this error" Condition="!$(SonarLintRegressionTestRunning)" />
```
tvsonar: Do we need this warning?

its/SonarLint.Testing.ImportBefore.targets
```diff
@@ -2,9 +2,15 @@
   <Target Name="OverrideRoslynCodeAnalysisPropertiesWithSonarLint"
           AfterTargets="ResolveCodeAnalysisRuleSet"
           BeforeTargets="CoreCompile">
-    <Error Text="Not running a SonarLint test build? Or forgot to build SonarLint first? Delete this file: $(MSBuildThisFileFullPath) to remove this error" Condition="!Exists('$(MSBuildStartupDirectory)\binaries\SonarLint.dll')" />
 
-    <ItemGroup>
+    <PropertyGroup>
+      <SonarLintRegressionTestRunning Condition="Exists('$(MSBuildStartupDirectory)\binaries\SonarLint.dll')">true</SonarLintRegressionTestRunning>
```
tvsonar: It's highly unlikely, but if there is a SonarLint.dll in a binaries folder in an arbitrary project, then we'll change the analyzers of the project and some properties. Changing the analyzers will change the number of warnings/errors reported, which is not good. The build will fail if there is no "output" folder at the solution level, ... but it would be an unlikely collision.

tvsonar:
SLVS-664 is already closed in Jira.

LGTM

its/SonarLint.Testing.ImportBefore.targets
dbolkensteyn: We don't strictly need this: It's unlikely that you'll get this warning in practice, unless you've interrupted the regression tests in the middle: This warning remind users in that case to remove this file to get rid of all its potential unwanted side effects (even though there should not be any).

PR #755, opened 2016-01-12
SLVS-709 Member override should do more than calling the base member

tvsonar:


src/SonarLint.CSharp/Rules/MemberOverrideCallsBaseMember.cs
```diff
@@ -0,0 +1,271 @@
...
+            return hasReturnValue
+                ? blockSyntax == null
+                    ? expressionBodySyntax?.Expression
+                    : (GetSingleStatement(blockSyntax) as ReturnStatementSyntax)?.Expression
+                : (GetSingleStatement(blockSyntax) as ExpressionStatementSyntax)?.Expression;
+        }
+
+        private static StatementSyntax GetSingleStatement(BlockSyntax block)
+        {
+            return block.Statements.Count == 1 ? block.Statements.First() : null;
```
dbolkensteyn: @tvsonar this can throw a NPE while coding

src/SonarLint.CSharp/Rules/MemberOverrideCallsBaseMember.cs
```diff
@@ -0,0 +1,271 @@
...
+            }
+
+            return true;
+        }
+
+        private static ExpressionSyntax GetExpressionToCheck(BlockSyntax blockSyntax, ArrowExpressionClauseSyntax expressionBodySyntax,
+            bool hasReturnValue)
+        {
+            return hasReturnValue
+                ? blockSyntax == null
```
dbolkensteyn: nested ternary is a bad idea

src/SonarLint.CSharp/Rules/MemberOverrideCallsBaseMember.cs
```diff
@@ -0,0 +1,271 @@
...
+                propertySyntax.ExpressionBody == null)
+            {
+                // no getter
+                return true;
+            }
+
+            var memberAccess = GetExpressionToCheck(getAccessor?.Body, propertySyntax.ExpressionBody, true)
+                as MemberAccessExpressionSyntax;
+
+            if (memberAccess == null ||
```
dbolkensteyn: this check could be removed with no functional impact, the property check done in `IsBaseProperty` is more robust

src/SonarLint.CSharp/Rules/MemberOverrideCallsBaseMember.cs
```diff
@@ -0,0 +1,271 @@
...
+
+            //check right:
+            var valueParameter = semanticModel.GetSymbolInfo(expressionToCheck.Right).Symbol as IParameterSymbol;
+            if (valueParameter == null || valueParameter.Name != "value" || !valueParameter.IsImplicitlyDeclared)
+            {
+                return false;
+            }
+
+            //check left:
+            var memberAccess = expressionToCheck.Left as MemberAccessExpressionSyntax;
```
dbolkensteyn: same thing here, this could be removed to simplify the code

src/Tests/SonarLint.UnitTest/TestCases/MemberOverrideCallsBaseMember.Fixed.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System;
```
dbolkensteyn: @tvsonar I think it's preferable to have a dedicated input file to test the code fix.

Indeed, it's super annoying to have to change the Fixed.cs file when we only want to fix a false positive in the rule

The corner cases that we need for the rule detection and the fixer are also different: The fixer will always for example care about how comments and spaces are transformed, whereas we typically don't care at all about these for the detection.

dbolkensteyn:
Also see https://jira.sonarsource.com/browse/SLVS-709

src/Tests/SonarLint.UnitTest/TestCases/MemberOverrideCallsBaseMember.Fixed.cs
tvsonar: That's true. However, I don't think we want to introduce one more test file for each analyzer, where we have a code fix. The ideal scenario is that we develop an analyzer, the test file contains all corner cases, we copy this file, remove the noncompliant code, and use the resulting file as the expected output of the codefix. And if either the analyzer or the codefix needs to be improved, then that will affect the other test file too.

src/SonarLint.CSharp/Rules/MemberOverrideCallsBaseMember.cs
tvsonar: We can't remove it, consider the following:

    public class MyDerived : MyBase
    {
        private MyBase instance;
        public override int MyProperty1 // Noncompliant
        {
            get
            {
                return base.MyProperty1;
            }
        }

        public override int MyProperty2 // Compliant
        {
            get
            {
                return instance.MyProperty2;
            }
        }
    }

dbolkensteyn: please update the test case with this

src/SonarLint.CSharp/Rules/MemberOverrideCallsBaseMember.cs
```diff
@@ -0,0 +1,303 @@
...
+            }
+
+            var methodSymbol = semanticModel.GetDeclaredSymbol(methodSyntax);
+
+            if (methodSymbol == null ||
+                !methodSymbol.IsOverride ||
+                methodSymbol.IsSealed ||
+                methodSymbol.OverriddenMethod == null ||
+                IgnoredMethodNames.Contains(methodSymbol.Name) ||
+                methodSymbol.Parameters.Any(p => p.HasExplicitDefaultValue))
```
dbolkensteyn: @tvsonar I would add this as well: `methodSymbol.OverriddenMethod.Parameters.Any(p => p.HasExplicitDefaultValue)`

it's not totally useless to want to make an optional parameter mandatory in a derived class, which is currently reported as redundant.

dbolkensteyn:
LGTM, but see my comment about the use case where the base method has an optional parameter, which the derived class makes mandatory.

PR #756, opened 2016-01-12
SLVS-655 Rework S2930 (DisposableNotDisposed) to remove all false positives

dbolkensteyn:


src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,273 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
-                analysisContext.RegisterSyntaxNodeAction(
-                    c => CollectDisposedSymbol((ConditionalAccessExpressionSyntax)c.Node, c.SemanticModel, disposeMethod,
-                    DisposableMemberInNonDisposableClass.IsNonStaticNonPublicDisposableField, ref fieldsDisposed),
-                    SyntaxKind.ConditionalAccessExpression);
-
-                analysisContext.RegisterCompilationEndAction(c =>
-                {
-                    var internallyInitializedFields = disposableFields.Intersect(fieldsAssigned);
-                    var nonDisposedFields = internallyInitializedFields.Except(fieldsDisposed);
+            "System.Net.Sockets.Socket",
```
dbolkensteyn: note sure that `Socket` should be part of the list because it has `Disconnect()` and `Shutdown()` after which dispose may or may not be required

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,273 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
-                {
-                    var internallyInitializedLocals = localsAssigned.Except(localsAssignedInUsing);
-                    var nonDisposedLocals = internallyInitializedLocals
-                        .Except(localsDisposed)
-                        .Except(localsReturned);
-
-                    ReportIssues(nonDisposedLocals, c.ReportDiagnostic);
-                });
-            });
+            "System.IO.FileStream",
```
dbolkensteyn: @tvsonar let me know if you see any other good candidate for these lists

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,273 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
-            if (TryGetLocallyConstructedSymbol(assignment, semanticModel,
-                DisposableMemberInNonDisposableClass.IsNonStaticNonPublicDisposableField, out fieldSymbol))
-            {
-                fieldsAssigned = fieldsAssigned.Add(fieldSymbol);
-            }
+            context.RegisterCompilationStartAction(
+                analysisContext =>
+                {
+                    var trackedNodesAndSymbols = new HashSet<Tuple<SyntaxNode, ISymbol>>();
+                    var excludedSymbols = new HashSet<ISymbol>();
```
dbolkensteyn: I expect this set to grow quite large, we could:
 1. restrict it to only private field and local symbols (not done yet)
 2. even better perhaps, restrict to only contain tracked symbols - but this will require the order of the sub-visitors to be deterministic - and I'm not sure if this is the case

dbolkensteyn: It's also unclear to me if we should pay attention to threading in this state: `RegisterCompilationStartAction` is designed to have a state, but does it have to protected by a lock or will all the sub-visitors run from the same thread?

I've had a quick look at the analyzers of the Roslyn project itself, but I can only find examples with a read only state: I can't find any where a sub-visitor updates it.

its/expected/Nancy/Nancy.Demo.Hosting.Aspnet-{E127FED3-01C0-41BA-BF83-D8DCDD827D6A}-S2930.json
```diff
@@ -0,0 +1,39 @@
...
+"fileVersion":  "1.1.0"
+},
+"issues":  [
+{
+"ruleId":  "S2930",
+"locations":  [
+{
+"analysisTarget":  [
+{
+"uri":  "Nancy\\src\\Nancy.Demo.Hosting.Aspnet\\PngSerializer.cs",
```
dbolkensteyn: this is a real issue! woohoo :)

dbolkensteyn:
the build on appveyor is failing because of the unit test run out of memory - which is super strange.

it's caused by the additional assembly that gets loaded in the verified - it seems that the `AdhocWorkspace` does get probably disposed, do you have an idea @tvsonar ?

src/Tests/SonarLint.UnitTest/Verifier.cs
```diff
@@ -36,11 +36,18 @@
...
 using Microsoft.CodeAnalysis.Text;
 using System.Text.RegularExpressions;
+using System.Net;
+using System;
 
 namespace SonarLint.UnitTest
 {
     public static class Verifier
     {
+        private static readonly MetadataReference systemAssembly = MetadataReference.CreateFromFile(typeof(object).Assembly.Location);
```
dbolkensteyn: this is clearly just a workaround... somehow references must be kept on these preventing garbage collection - but I wasn't able to quickly figure out what's going on exactly

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
dbolkensteyn: i'm removing it

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
dbolkensteyn: I've updated the PR to only exclude private fields & locals

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -40,14 +41,6 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
     {
         internal const string DiagnosticId = "S2930";
         internal const string Title = "\"IDisposables\" should be disposed";
-        internal const string Description =
```
tvsonar: It looks nice in VS to have a description. We should add a summary of the RSPEC. Or just use the first two paragraph of RSPEC as they are.

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
             new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category,
                 RuleSeverity.ToDiagnosticSeverity(), IsActivatedByDefault,
-                helpLinkUri: DiagnosticId.GetHelpLink(),
-                description: Description);
+                helpLinkUri: DiagnosticId.GetHelpLink());
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Rule); } }
 
-        public override void Initialize(AnalysisContext context)
+        private readonly IImmutableSet<string> TrackedTypes = new string[]
```
tvsonar: This can be made `static`. Also, `new string[]` can be simplified to `new []`.

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
+            "System.Net.Sockets.TcpListener",
+            "System.Net.Sockets.UdpClient",
 
-                    ReportIssues(nonDisposedFields, diagnostic => c.ReportDiagnosticIfNonGenerated(diagnostic, c.Compilation));
-                });
-            });
-        }
+            "System.Drawing.Image",
+            "System.Drawing.Bitmap"
+        }.ToImmutableHashSet();
```
tvsonar: Do we gain anything with this `ToImmutableHashSet` call? In the other classes where we store a list of method/class names we tend not to use it. The `ImmutableHashSet` of strings could simply be stored in a `string[]`.

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
 
-        internal static IMethodSymbol GetDisposeMethod(Compilation compilation)
+        private readonly IImmutableSet<string> DisposeMethods = new string[]
```
tvsonar: same as above

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
-                .SingleOrDefault();
-        }
+            "Dispose",
+            "Close"
+        }.ToImmutableHashSet();
 
-        private static void CollectDisposedSymbol<T>(InvocationExpressionSyntax invocation, SemanticModel semanticModel,
-            IMethodSymbol disposeMethod, Predicate<T> isSymbolRelevant, ref ImmutableHashSet<T> fieldsDisposed)
-             where T : class, ISymbol
+        private readonly IImmutableSet<string> FactoryMethods = new string[]
```
tvsonar: same as above

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
+                                    c.ReportDiagnosticIfNonGenerated(Diagnostic.Create(Rule, trackedNodeAndSymbol.Item1.GetLocation(), trackedNodeAndSymbol.Item2.Name), c.Compilation);
+                                }
+                            }
+                        });
+                });
         }
 
-        private static void CollectDisposableFields(FieldDeclarationSyntax field, SemanticModel semanticModel,
-            ref ImmutableHashSet<IFieldSymbol> disposableFields, ref ImmutableHashSet<IFieldSymbol> fieldsAssigned)
+        private void TrackInitializedLocalsAndPrivateFields(CompilationStartAnalysisContext context, ISet<Tuple<SyntaxNode, ISymbol>> trackedNodesAndSymbols)
```
tvsonar: Could be `static`.

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
-                    continue;
-                }
+                    VariableDeclarationSyntax declaration;
+                    if (c.Node.IsKind(SyntaxKind.LocalDeclarationStatement))
+                    {
+                        declaration = ((LocalDeclarationStatementSyntax)c.Node).Declaration;
+                    }
+                    else if (c.Node.IsKind(SyntaxKind.FieldDeclaration))
+                    {
+                        var n = (FieldDeclarationSyntax)c.Node;
```
tvsonar: instead of `n`, let's use `fieldDeclaration`

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
-            return methodSymbol != null &&
-                   (methodSymbol.Equals(disposeMethod) ||
-                   methodSymbol.Equals(methodSymbol.ContainingType.FindImplementationForInterfaceMember(disposeMethod)));
+            return !(e.Parent is ExpressionSyntax) ||
+                (parentAsAssignment != null && object.ReferenceEquals(e, parentAsAssignment.Right));
         }
 
-        private static bool TryGetLocallyConstructedSymbol<T>(AssignmentExpressionSyntax assignment, SemanticModel semanticModel,
-            Predicate<T> isSymbolRelevant, out T localSymbol, bool doCheckNonLocalDisposables = false) where T : class, ISymbol
+        private bool IsInstantiation(ExpressionSyntax e, SemanticModel sm)
```
tvsonar: Parameter names could be changed to `expression` and `semanticModel`.

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
+            ITypeSymbol type = sm.GetTypeInfo(e).Type;
+            if (type == null || !TrackedTypes.Contains(type.ToDisplayString()))
             {
                 return false;
             }
 
-            localSymbol = semanticModel.GetDeclaredSymbol(declarator) as ILocalSymbol;
-            return IsDisposableLocalSymbol(localSymbol);
+            var constructor = sm.GetSymbolInfo(e).Symbol as IMethodSymbol;
+            return constructor != null && !constructor.Parameters.Any(p => DisposableMemberInNonDisposableClass.ImplementsIDisposable(p.Type));
```
tvsonar: `ImplementsIDisposable` is used in 3 different classes, we could put it into a helper class.

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
         }
 
-        private static bool MightUseNonLocalDisposable(ObjectCreationExpressionSyntax objectCreation,
-            SemanticModel semanticModel)
+        private bool IsFactoryMethodInvocation(ExpressionSyntax e, SemanticModel sm)
```
tvsonar: parameter names as above.

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
         {
-            return MightUseNonLocalDisposableInArguments(objectCreation, semanticModel) ||
-                   MightUseNonLocalDisposableInInitializer(objectCreation, semanticModel);
-        }
-
-        private static bool MightUseNonLocalDisposableInInitializer(ObjectCreationExpressionSyntax objectCreation,
-            SemanticModel semanticModel)
-        {
-            if (objectCreation.Initializer == null)
+            var n = e as InvocationExpressionSyntax;
```
tvsonar: variable name as above

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
-                    })
-                .Any(expression => HasDisposableNotLocalIdentifier(expression, semanticModel));
-        }
-
-        private static bool MightUseNonLocalDisposableInArguments(ObjectCreationExpressionSyntax objectCreation,
-            SemanticModel semanticModel)
-        {
-            if (objectCreation.ArgumentList == null)
+            var methodSymbol = sm.GetSymbolInfo(n).Symbol as IMethodSymbol;
+            if (methodSymbol == null)
```
tvsonar: We don't have to check the names of non-static methods, so we could filter them with ` || !methodSymbol.IsStatic`

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
-            {
-                return false;
-            }
-
-            var methodSymbol = semanticModel.GetSymbolInfo(conditionalAccess.WhenNotNull).Symbol as IMethodSymbol;
-            return methodSymbol != null &&
-                   (methodSymbol.Equals(disposeMethod) ||
-                   methodSymbol.Equals(methodSymbol.ContainingType.FindImplementationForInterfaceMember(disposeMethod)));
+            return !(e.Parent is ExpressionSyntax) ||
+                (parentAsAssignment != null && object.ReferenceEquals(e, parentAsAssignment.Right));
```
tvsonar: For me `==` is easier to read than `object.ReferenceEquals`.

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
+                        }
+                    }
+                },
+                SyntaxKind.IdentifierName,
+                SyntaxKind.SimpleMemberAccessExpression);
         }
 
-        private static bool TryGetDisposedSymbol<T>(ConditionalAccessExpressionSyntax conditionalAccess, SemanticModel semanticModel,
-            IMethodSymbol disposeMethod, Predicate<T> isSymbolRelevant, out T localSymbol) where T : class, ISymbol
+        private bool IsStandaloneExpression(ExpressionSyntax e)
```
tvsonar: single letter parameter name

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
         {
-            localSymbol = null;
+            var parentAsAssignment = e.Parent as AssignmentExpressionSyntax;
```
tvsonar: this cast is not needed if `e.Parent` is not `ExpressionSyntax`

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
-                if (variableDeclaratorSyntax.Initializer == null ||
-                    !(variableDeclaratorSyntax.Initializer.Value is ObjectCreationExpressionSyntax))
+        private void TrackAssignmentsToLocalsAndPrivateFields(CompilationStartAnalysisContext context, ISet<Tuple<SyntaxNode, ISymbol>> trackedNodesAndSymbols)
+        {
+            context.RegisterSyntaxNodeAction(
+                c =>
                 {
-                    return;
-                }
+                    var n = c.Node as AssignmentExpressionSyntax;
```
tvsonar: Why not cast here?

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
-        {
-            ILocalSymbol localSymbol;
-            if (TryGetLocallyConstructedInitializedLocalSymbol(variableDeclarator, semanticModel, out localSymbol))
-            {
-                localsAssigned = localsAssigned.Add(localSymbol);
-            }
-        }
+                    if (c.Node.IsKind(SyntaxKind.InvocationExpression))
+                    {
+                        var n = c.Node as InvocationExpressionSyntax;
```
tvsonar: Why not cast here?

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
-            {
-                localsAssigned = localsAssigned.Add(localSymbol);
-            }
-        }
+                        name = memberAccessNode?.Name;
+                        expression = memberAccessNode?.Expression;
+                    }
+                    else if (c.Node.IsKind(SyntaxKind.ConditionalAccessExpression))
+                    {
+                        var n = c.Node as ConditionalAccessExpressionSyntax;
```
tvsonar: Why not cast here?

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
-            {
-                return false;
-            }
+            context.RegisterSyntaxNodeAction(
+                c =>
+                {
+                    ExpressionSyntax expression;
+                    if (c.Node.IsKind(SyntaxKind.IdentifierName))
+                    {
+                        expression = c.Node as IdentifierNameSyntax;
```
tvsonar: Why not cast here?

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
+                    }
+                    else if (c.Node.IsKind(SyntaxKind.SimpleMemberAccessExpression))
+                    {
 
-            localSymbol = semanticModel.GetSymbolInfo(memberAccess.Expression).Symbol as T;
-            if (!isSymbolRelevant(localSymbol))
-            {
-                return false;
-            }
+                        var n = c.Node as MemberAccessExpressionSyntax;
```
tvsonar: Why not cast here?

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -56,383 +49,268 @@ public class DisposableNotDisposed : DiagnosticAnalyzer
...
-            IFieldSymbol fieldSymbol;
-            if (TryGetLocallyConstructedSymbol(assignment, semanticModel,
-                DisposableMemberInNonDisposableClass.IsNonStaticNonPublicDisposableField, out fieldSymbol))
-            {
-                fieldsAssigned = fieldsAssigned.Add(fieldSymbol);
-            }
+            context.RegisterCompilationStartAction(
+                analysisContext =>
+                {
+                    var trackedNodesAndSymbols = new HashSet<Tuple<SyntaxNode, ISymbol>>();
```
tvsonar: Intsead of the `Tuple` I would create a nested class, with appropriate property names. `Item1` and `Item2` doesn't mean anything a few lines below.

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
tvsonar: `HttpClient` is also `IDisposable`

tvsonar:
LGTM

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
dbolkensteyn: in fact I also had in mind whole other namespaces, perhaps WinForms etc

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
dbolkensteyn: i'd ask the question the other way around: Do we gain anything by switching to an array of string?As only `Contains()` is called on this, a `Set` does seem like a good reasonable default choice.

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
dbolkensteyn: I won't do that, or at least not now: it's already complicated enough to write a good RSPEC and then keep that in sync with the HTML description & implementation - look at the number of outdated rules on the language team wallboard

Plus at the end you don't even see the Noncompliant & Compliant examples, which is perhaps the most valuable part of the description.

IMO it's best for users to learn where they can find the real description on SonarLint's website.

=> I'm not convinced about the ROI of these descriptions

PR #760, opened 2016-01-13
SLVS-405 S1854 should not report inside "try" blocks

tvsonar:


src/SonarLint.CSharp/Rules/DeadStores.cs
```diff
@@ -183,14 +183,32 @@ public override void Initialize(AnalysisContext context)
...
+                var tryStatement = currentNode.Parent as TryStatementSyntax;
+                if (tryStatement != null && tryStatement.Block == currentNode)
+                {
+                    return;
+                }
+
+                currentNode = currentNode.Parent;
+            }
+
+            context.ReportDiagnostic(Diagnostic.Create(Rule, location, name));
```
dbolkensteyn: it would be better here to do `node.GetLocation()` if that's possible, and not pass both a node and a location

dbolkensteyn:
LGTM

PR #761, opened 2016-01-13
SLVS-734 Do not ship *.pdb files in the released VSIX

tvsonar:


dbolkensteyn:
@tvsonar let's squash & force push only when we want to merge. But while reviewing of the PR is happening, it's best to keep all commits for tracability (I think that the AppVeyor build on this one failed initially because the VSIX shrunk)

dbolkensteyn:
LGTM

PR #763, opened 2016-01-14
SLVS-712 Redundant "ToString()" call

tvsonar:


its/expected/akka.net/Akka.Remote.Tests-{95E921AB-1F5A-4AF9-B7FE-284E317A97F4}-S1858.json
```diff
@@ -0,0 +1,41 @@
...
+"issues":  [
+{
+"ruleId":  "S1858",
+"locations":  [
+{
+"analysisTarget":  [
+{
+"uri":  "akka.net\\src\\core\\Akka.Remote.Tests\\RemoteDaemonSpec.cs",
+"region":  {
+"startLine":  71,
```
tvsonar: Real issue: `ToString` called on a string.

its/expected/akka.net/Akka-{5DEDDF90-37F0-48D3-A0B0-A5CBD8A7E377}-S1858.json
```diff
@@ -0,0 +1,103 @@
...
+},
+"issues":  [
+{
+"ruleId":  "S1858",
+"locations":  [
+{
+"analysisTarget":  [
+{
+"uri":  "akka.net\\src\\core\\Akka\\Actor\\ActorBase.cs",
+"region":  {
```
tvsonar: Real issue: `ToString` called in a concatenation. `return "Failure: " + Cause.ToString();`

dbolkensteyn: @tvsonar don't forget to send that to the sprint's highlight thread

src/SonarLint.CSharp/Rules/RedundantToStringCall.cs
```diff
@@ -0,0 +1,230 @@
...
+                    }
+                },
+                SyntaxKind.AddAssignmentExpression);
+        }
+
+        private static void CheckSides(SyntaxNodeAnalysisContext c, ExpressionSyntax sideA, ExpressionSyntax sideB, int checkedSideIndex)
+        {
+            Location location;
+            IMethodSymbol methodSymbol;
+            if (IsCandidateToStringCall(sideA, c.SemanticModel, out location, out methodSymbol) &&
```
dbolkensteyn: @tvsonar `CheckSides()` and `IsCandidateToStringCall()` are both not very descriptive of what exactly they do: You must look at the code if you want to know.

I also think that a complex method should not call another complex one, because, to know what `CheckSides()` does, I also need to look at `IsCandidateToStringCall()`.

dbolkensteyn:
@tvsonar Several methods do not comply with the method complexity rule, and the overall code seems unnecessarily complex.

Once you've managed to implement the rule and are happy with the functional results, it is really important to spend some more time to refactor the code to make it as clear and simple as you can

src/SonarLint.CSharp/Rules/RedundantToStringCall.cs
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        private static bool IsCandidateToStringCall(ExpressionSyntax expression, SemanticModel semanticModel,
+            out Location location, out IMethodSymbol methodSymbol)
+        {
+            location = null;
+            methodSymbol = null;
+            var invocation = expression as InvocationExpressionSyntax;
+            if (invocation == null ||
+                invocation.ArgumentList.CloseParenToken.IsMissing)
```
dbolkensteyn: what's the intended purpose of this check? Visual Studio inserts the closing parenthesis immediatly for me when I type the opening one

src/SonarLint.CSharp/Rules/RedundantToStringCall.cs
```diff
@@ -0,0 +1,230 @@
...
+                methodSymbol.Name != "ToString" ||
+                methodSymbol.Parameters.Any())
+            {
+                return false;
+            }
+
+            location = Location.Create(invocation.SyntaxTree,
+                TextSpan.FromBounds(
+                    memberAccess.OperatorToken.SpanStart,
+                    invocation.ArgumentList.CloseParenToken.IsMissing
```
dbolkensteyn: can this ever evaluate to true? it seems this was checked already at line 201

dbolkensteyn: why not `invocation.Span.End` ?

dbolkensteyn:
there is a false positive in this case:

```
public class MyType
{
    public static string operator +(string a, MyType b)
    {
        return "mwahaha";
    }
}

public class A
{
    public static void Main()
    {
        var t = new MyType();

        var c = "foo";
        c += t.ToString(); // Should be compliant - removing it changes the behavior

        Console.WriteLine(c);
        Console.ReadLine();
    }
}
```

dbolkensteyn:
I've slightly edited the RSPEC of this one

src/SonarLint.CSharp/Rules/RedundantToStringCall.cs
tvsonar: when you have something like this in your code: `.ToString(`, but the `)` is missing, then we don't want to report, because we don't really know at that state which overload will be called.

VS automatically adds the `()` but if you erase it you can still get to the mentioned state. (I got there somehow, and the fading at that point is misleading.)

src/SonarLint.CSharp/Rules/RedundantToStringCall.cs
```diff
@@ -0,0 +1,254 @@
...
+                description: Description,
+                customTags: ideVisibility.ToCustomTags());
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Rule); } }
+
+        private const string additionOperatorName = "op_Addition";
+
+        public override void Initialize(AnalysisContext context)
+        {
+            RegisterActionForInvocations(context);
```
dbolkensteyn: that's still a too technical name for me => it doesn't say anything about what this rule is doing or how it is working

src/SonarLint.CSharp/Rules/RedundantToStringCall.cs
```diff
@@ -0,0 +1,254 @@
...
+                c =>
+                {
+                    var assignment = (AssignmentExpressionSyntax)c.Node;
+                    var operation = c.SemanticModel.GetSymbolInfo(assignment).Symbol as IMethodSymbol;
+                    if (!IsOperationAddOnString(operation))
+                    {
+                        return;
+                    }
+
+                    CheckExpressionForRemovableToStringCall(c, assignment.Right, assignment.Left, 1);
```
dbolkensteyn: this 1 definitely looks magic :)

src/SonarLint.CSharp/Rules/RedundantToStringCall.cs
```diff
@@ -0,0 +1,254 @@
...
+                return;
+            }
+
+            var subExpressionType = context.SemanticModel.GetTypeInfo(subExpression).Type;
+            if (subExpressionType == null)
+            {
+                return;
+            }
+
+            var stringParameterIndex = (checkedSideIndex + 1) % 2;
```
dbolkensteyn: seems too complicated

dbolkensteyn:
LGTM - but I still find this rule overly complex for what's it's doing - it should be possible to come up with a better looking implementation

PR #766, opened 2016-01-15
SLVS-730 Do not report issues on partially generated files using old versions of Windows Form Designer

dbolkensteyn:


tvsonar:
LGTM

PR #768, opened 2016-01-15
SLVS-632 S1172 (unused parameters) has false positives on delegate methods & non-private methods

dbolkensteyn:


its/expected/Nancy/Nancy-{34576216-0DCA-4B0F-A0DC-9075E75A676F}-S1172.json
```diff
@@ -12,134 +12,10 @@
...
-"ruleId":  "S1172",
-"locations":  [
-{
-"analysisTarget":  [
-{
 "uri":  "Nancy\\src\\Nancy\\Diagnostics\\DiagnosticsHook.cs",
 "region":  {
 "startLine":  258,
-"startColumn":  152,
+"startColumn":  128,
```
dbolkensteyn: this change of issue location to include the parameter type doesn't hurt IMO, and makes the implementation easier (no need to find an  IParameterSymbol 's identifier node)

its/expected/Nancy/Nancy.Demo.CustomModule-{F56F3983-0C34-4AEC-B418-A8AFFA63F1C4}-S1172.json
```diff
@@ -1,41 +0,0 @@
-{
```
dbolkensteyn: lots of FP are being removed

dbolkensteyn:
with this new approach, the rule becomes a bit slower during a SQ analysis, but is more responsive (it no longer depends on the size of the project):
```
SonarLint.Rules.CSharp.MethodParameterUnused                            5.528
```

whereas before this PR it was:
```
SonarLint.Rules.CSharp.MethodParameterUnused                              1.44
```

I need to find the time spent before the last commit, to know how fast the compilation event based solution was.

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,64 +60,141 @@ public class MethodParameterUnused : DiagnosticAnalyzer
...
+        private static IImmutableList<IParameterSymbol> GetUnusedParameters(IMethodSymbol methodSymbol, Compilation compilation)
+        {
+            var usedParameters = new HashSet<IParameterSymbol>();
 
-                    cbc.RegisterSyntaxNodeAction(
-                        c =>
+            var bodies = methodSymbol.DeclaringSyntaxReferences
+                .Select(r => r.GetSyntax())
+                .Where(n => n.IsKind(SyntaxKind.MethodDeclaration) || n.IsKind(SyntaxKind.ConstructorDeclaration))
+                .SelectMany(
```
dbolkensteyn: @tvsonar could this be slow?

dbolkensteyn: do we have a way to profile an analyzer BTW?

dbolkensteyn:
Note to self: Send update to 1.7 highlights

dbolkensteyn:
Note to self: Run the comparison tool

its/expected/Nancy/Nancy-{34576216-0DCA-4B0F-A0DC-9075E75A676F}-S1172.json
tvsonar: I don't mind this change, it makes the codefix very simple.
Note however that R# is only fading out the name of the parameter.

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,64 +60,141 @@ public class MethodParameterUnused : DiagnosticAnalyzer
...
+            if (!unusedParameters.Any())
+            {
+                return;
+            }
+
+            if (!IsUsedAsEventHandlerFunctionOrAction(methodSymbol, c.Compilation))
+            {
+                foreach (var unusedParameter in unusedParameters)
+                {
+                    foreach (var unusedParameterDeclaration in unusedParameter.DeclaringSyntaxReferences.Select(r => r.GetSyntax()))
```
tvsonar: To only report on the parameter name, we would need to modify this single location to `((ParameterSyntax)r.GetSyntax()).Identifier`

its/expected/Nancy/Nancy.Demo.CustomModule-{F56F3983-0C34-4AEC-B418-A8AFFA63F1C4}-S1172.json
tvsonar: We should also check public methods of private classes. R# does that. There's an effective visibility calculation method at https://github.com/SonarSource/sonarlint-vs/pull/767/files#diff-d58ee1cdb05597214f0cec6356cf8974R422

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
tvsonar: You could give a try to the CPU profiling in the VS/Diagnostic Tools.

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,64 +60,141 @@ public class MethodParameterUnused : DiagnosticAnalyzer
...
+                    {
+                        c.ReportDiagnosticIfNonGenerated(Diagnostic.Create(Rule, unusedParameterDeclaration.GetLocation(), unusedParameter.Name));
+                    }
+                }
+            }
+        }
+
+        private static bool MethodCanBeSafelyChanged(IMethodSymbol methodSymbol, Compilation compilation)
+        {
+            return methodSymbol.DeclaredAccessibility == Accessibility.Private &&
```
tvsonar: We could check the effective visibility here.

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,64 +60,141 @@ public class MethodParameterUnused : DiagnosticAnalyzer
...
+                    builder.Add((IParameterSymbol)symbol);
+                }
+            }
+
+            return builder.ToImmutable();
+        }
+
+        private static bool IsUsedAsEventHandlerFunctionOrAction(IMethodSymbol methodSymbol, Compilation compilation)
+        {
+            return methodSymbol
```
tvsonar: I tend to only put the LINQ calls to new line or chains that are too long.

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,64 +60,141 @@ public class MethodParameterUnused : DiagnosticAnalyzer
...
+                .DeclaringSyntaxReferences
+                .Select(r => r.GetSyntax())
+                .Any(n => IsUsedAsEventHandlerFunctionOrAction(n, methodSymbol, compilation.GetSemanticModel(n.SyntaxTree)));
+        }
+
+        private static bool IsUsedAsEventHandlerFunctionOrAction(SyntaxNode classDeclaration, IMethodSymbol methodSymbol, SemanticModel semanticModel)
+        {
+            return classDeclaration
+                .DescendantNodes()
+                .Any(n => n is ExpressionSyntax && IsUsedAsEventHandlerFunctionOrAction((ExpressionSyntax)n, methodSymbol, semanticModel));
```
tvsonar: double casting here, use `OfType<ExpressionSyntax>()` instead

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,64 +60,141 @@ public class MethodParameterUnused : DiagnosticAnalyzer
...
+        private static bool IsUsedAsEventHandlerFunctionOrAction(IMethodSymbol methodSymbol, Compilation compilation)
+        {
+            return methodSymbol
+                .ContainingType
+                .DeclaringSyntaxReferences
+                .Select(r => r.GetSyntax())
+                .Any(n => IsUsedAsEventHandlerFunctionOrAction(n, methodSymbol, compilation.GetSemanticModel(n.SyntaxTree)));
+        }
+
+        private static bool IsUsedAsEventHandlerFunctionOrAction(SyntaxNode classDeclaration, IMethodSymbol methodSymbol, SemanticModel semanticModel)
```
tvsonar: The previous method properly calculates if the method symbol `IsUsedAsEventHandlerFunctionOrAction`. But this method does slightly different stuff. I would change the order of the `classDeclaration` and the `methodSymbol`, and would rename `classDeclaration` to `container`, as we are trying to find inside of it a call.

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,64 +60,141 @@ public class MethodParameterUnused : DiagnosticAnalyzer
...
+        {
+            return classDeclaration
+                .DescendantNodes()
+                .Any(n => n is ExpressionSyntax && IsUsedAsEventHandlerFunctionOrAction((ExpressionSyntax)n, methodSymbol, semanticModel));
+        }
+
+        private static bool IsUsedAsEventHandlerFunctionOrAction(ExpressionSyntax expression, IMethodSymbol methodSymbol, SemanticModel semanticModel)
```
tvsonar: I'd change the order of the parameters here too.

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,64 +60,141 @@ public class MethodParameterUnused : DiagnosticAnalyzer
...
+        {
+            return !expression.IsKind(SyntaxKind.InvocationExpression) &&
+                IsStandaloneExpression(expression) &&
+                methodSymbol == semanticModel.GetSymbolInfo(expression).Symbol;
+        }
+
+        private static bool IsStandaloneExpression(ExpressionSyntax expression)
+        {
+            var parentAsAssignment = expression.Parent as AssignmentExpressionSyntax;
```
tvsonar: there's no need for this cast if `!(expression.Parent is ExpressionSyntax)`

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,64 +60,141 @@ public class MethodParameterUnused : DiagnosticAnalyzer
...
+        }
+
+        private static bool IsUsedAsEventHandlerFunctionOrAction(ExpressionSyntax expression, IMethodSymbol methodSymbol, SemanticModel semanticModel)
+        {
+            return !expression.IsKind(SyntaxKind.InvocationExpression) &&
+                IsStandaloneExpression(expression) &&
+                methodSymbol == semanticModel.GetSymbolInfo(expression).Symbol;
+        }
+
+        private static bool IsStandaloneExpression(ExpressionSyntax expression)
```
tvsonar: this logic is duplicated in two classes already, and chances are we could reuse it in other places as well. BTW, you didn't refactor it in the other place.

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,64 +60,141 @@ public class MethodParameterUnused : DiagnosticAnalyzer
...
+                return;
+            }
+
+            var unusedParameters = GetUnusedParameters(methodSymbol, c.Compilation);
+            if (!unusedParameters.Any())
+            {
+                return;
+            }
+
+            if (!IsUsedAsEventHandlerFunctionOrAction(methodSymbol, c.Compilation))
```
tvsonar: we can reduce the nesting by moving this condition to the previous `if`.

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,64 +60,141 @@ public class MethodParameterUnused : DiagnosticAnalyzer
...
+            var unusedParameter = methodSymbol.Parameters.Except(usedParameters);
+            if (methodSymbol.IsExtensionMethod)
+            {
+                unusedParameter = unusedParameter.Except(new[] { methodSymbol.Parameters.First() });
+            }
+
+            return unusedParameter.Except(usedParameters).ToImmutableArray();
+        }
+
+        private static IImmutableSet<IParameterSymbol> GetUsedParameters(ImmutableArray<IParameterSymbol> parameters, SyntaxNode body, SemanticModel semanticModel)
```
tvsonar: This method can be simplified to a single LINQ chain:

    return body.DescendantNodes()
       .Where(n => n.IsKind(SyntaxKind.IdentifierName))
       .Select(identierName => semanticModel.GetSymbolInfo(identierName).Symbol as IParameterSymbol)
       .Where(symbol => symbol != null && parameters.Contains(symbol))
       .ToImmutableHashSet();

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,64 +60,141 @@ public class MethodParameterUnused : DiagnosticAnalyzer
...
 
-                    cbc.RegisterSyntaxNodeAction(
-                        c =>
+            var bodies = methodSymbol.DeclaringSyntaxReferences
+                .Select(r => r.GetSyntax())
+                .Where(n => n.IsKind(SyntaxKind.MethodDeclaration) || n.IsKind(SyntaxKind.ConstructorDeclaration))
+                .SelectMany(
+                    n =>
+                    {
+                        if (n.IsKind(SyntaxKind.MethodDeclaration))
```
tvsonar: R# also reports on operator declarations.

tvsonar:
We might want to exclude `static void Main(string[] args)`, where `args` is not used. Although `args` can be removed, this is what VS generates by default, so the rule will trigger issues on an empty default project. (R# reports on it)

tvsonar:
False positive on generic methods:
```
private class MyType
{
    public MyType()
    {
        Action<string> x = WriteLine;
        Action<string> y = WriteLine<int>;
    }

    private static void WriteLine<T>(string format) { } // Rule reports here 
    private static void WriteLine(string format) { }             
}
```

You can get the underlying symbol from a generic one by using the `OriginalDefinition` property. You'll need something like this `methodSymbol == semanticModel.GetSymbolInfo(expression).Symbol?.OriginalDefinition;` in `IsUsedAsEventHandlerFunctionOrAction`.

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,64 +60,141 @@ public class MethodParameterUnused : DiagnosticAnalyzer
...
+            return classDeclaration
+                .DescendantNodes()
+                .Any(n => n is ExpressionSyntax && IsUsedAsEventHandlerFunctionOrAction((ExpressionSyntax)n, methodSymbol, semanticModel));
+        }
+
+        private static bool IsUsedAsEventHandlerFunctionOrAction(ExpressionSyntax expression, IMethodSymbol methodSymbol, SemanticModel semanticModel)
+        {
+            return !expression.IsKind(SyntaxKind.InvocationExpression) &&
+                IsStandaloneExpression(expression) &&
+                methodSymbol == semanticModel.GetSymbolInfo(expression).Symbol;
```
tvsonar: `==` calls reference equality here because we are working on interfaces. It's probably fine, because we want to check reference equality, but it's best to use `Equals`, which is overriden in `Symbol` and has double dispatch. See [link](https://github.com/dotnet/roslyn/wiki/FAQ#how-do-i-tell-if-symbols-are-the-same).

its/expected/Nancy/Nancy.Demo.CustomModule-{F56F3983-0C34-4AEC-B418-A8AFFA63F1C4}-S1172.json
dbolkensteyn: we can't make pull requests depend on one another, that's a huge anti-pattern - let's enter a new ticket to fix it in the next sprint

dbolkensteyn: ticket created: https://jira.sonarsource.com/browse/SLVS-762

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
dbolkensteyn: as discussed above, ticket created: https://jira.sonarsource.com/browse/SLVS-762 - we're not going to add any dependencies between pull requests

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
dbolkensteyn: in fact the method name sucks, and having 2 overloads differening only with the `SyntaxNode` vs `ExpressionSyntax` is super dangerous and misleading

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
dbolkensteyn: such operators must be public, even if they are within a private class, so that's going to be part of https://jira.sonarsource.com/browse/SLVS-762

dbolkensteyn:
@tvsonar Did you run the comparison tool on this rule to notice the difference already? If not, could you take care of that please?

tvsonar:
I ran the tool. R# finds more issue because of the `public` method in a `private` class issue. But that'll be handled.

LGTM

PR #769, opened 2016-01-19
SLVS-655 Rework DisposableNotDisposed to not be based on compilation events & fix FPs/FNs

dbolkensteyn:


dbolkensteyn:
This rule's execution time went up from 2.2 seconds to 4.3, but is now way more responsive (as in real-time) on large projects.

tvsonar:
This PR fixes FP&FNs, so it would also be good be place to add `HttpClient` to the `TrackedTypes`.

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -91,89 +91,107 @@ private class NodeAndSymbol
...
         {
-            context.RegisterCompilationStartAction(
-                analysisContext =>
+            context.RegisterSymbolAction(
+                c =>
                 {
-                    var trackedNodesAndSymbols = new HashSet<NodeAndSymbol>();
-                    var excludedSymbols = new HashSet<ISymbol>();
+                    var namedType = (INamedTypeSymbol)c.Symbol;
+                    if (namedType.TypeKind != TypeKind.Class && namedType.TypeKind != TypeKind.Struct || namedType.ContainingType != null)
```
tvsonar: Parentheses should be added to explicitly mark the precedence of `&&` and `||`. Also, we could swap the two sides of the `||`

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -91,89 +91,107 @@ private class NodeAndSymbol
...
+                    {
+                        return;
+                    }
 
-                    TrackInitializedLocalsAndPrivateFields(analysisContext, trackedNodesAndSymbols);
-                    TrackAssignmentsToLocalsAndPrivateFields(analysisContext, trackedNodesAndSymbols);
+                    var typesDeclarationAndCompilation =
```
tvsonar: misleading variable name, it contains `SyntaxNode`-`SemanticModel` pairs, and not `Compilation`s

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -182,29 +200,32 @@ private static bool IsLocalOrPrivateField(ISymbol symbol)
...
-            context.RegisterSyntaxNodeAction(
-                c =>
-                {
+            var incovationsAndConditionalAccesses = typeDeclaration
+                .DescendantNodes()
+                .Where(n => n.IsKind(SyntaxKind.InvocationExpression) || n.IsKind(SyntaxKind.ConditionalAccessExpression));
+
+            foreach (SyntaxNode incovationOrConditionalAccess in incovationsAndConditionalAccesses)
+            {
                     SimpleNameSyntax name;
```
tvsonar: too long indentation

src/SonarLint.CSharp/Rules/DisposableNotDisposed.cs
```diff
@@ -219,55 +240,54 @@ private static void ExcludeDisposedAndClosedLocalsAndPrivateFields(CompilationSt
...
+                        excludedSymbols.Add(referencedSymbol);
                     }
-                },
-                SyntaxKind.IdentifierName,
-                SyntaxKind.SimpleMemberAccessExpression);
+                }
+            }
         }
 
         private static bool IsStandaloneExpression(ExpressionSyntax expression)
```
tvsonar: The body if this method still bothers me.

tvsonar:
We have false negatives on arrays, but I don't think we can do a lot about it:
```
private static IDisposable[] field_fs3 = new[] { new FileStream(@"c:\foo.txt", FileMode.Open) };
```
We could check if inside the class there's at least one `field_fs3[X].Dispose()` or `.Close()` with any index or if the `field_fs3` or any `field_fs3[X]` is passed around.

tvsonar:
LGTM

PR #771, opened 2016-01-19
SLVS-459 S2934 should not produce false-positives with chained type c…

tvsonar:
…onstraints

### DIFF
--- a/src/SonarLint.CSharp/Rules/GenericReadonlyFieldPropertyAssignment.cs
+++ b/src/SonarLint.CSharp/Rules/GenericReadonlyFieldPropertyAssignment.cs
@@ -121,17 +121,8 @@ namespace SonarLint.Rules.CSharp
             }
 
             var fieldSymbol = semanticModel.GetSymbolInfo(memberAccess.Expression).Symbol as IFieldSymbol;
-            if (fieldSymbol == null ||
-                !fieldSymbol.IsReadOnly ||
-                !RelevantFieldType(fieldSymbol.Type))
-            {
-                return;
-            }
-
-            var constructorSymbol = semanticModel.GetEnclosingSymbol(expression.SpanStart) as IMethodSymbol;
-            if (constructorSymbol != null &&
-                constructorSymbol.MethodKind == MethodKind.Constructor &&
-                constructorSymbol.ContainingType.Equals(fieldSymbol.ContainingType))
+            if (!IsFieldReadonlyAndPossiblyValueType(fieldSymbol) ||
+                IsInsideConstructorDeclaration(expression, fieldSymbol.ContainingType, semanticModel))
             {
                 return;
             }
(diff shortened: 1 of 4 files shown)


src/SonarLint.CSharp/Rules/GenericReadonlyFieldPropertyAssignment.cs
```diff
@@ -108,47 +108,74 @@ public override void Initialize(AnalysisContext context)
...
-            if (memberAccess == null)
+            var propertySymbol = semanticModel.GetSymbolInfo(expression).Symbol as IPropertySymbol;
+            if (propertySymbol == null)
             {
                 return;
             }
 
-            var propertySymbol = semanticModel.GetSymbolInfo(expression).Symbol as IPropertySymbol;
-            if (propertySymbol == null)
+            var memberAccess = expression as MemberAccessExpressionSyntax;
```
dbolkensteyn: looks like this check should be made first for performances? why did you reorder?

src/SonarLint.CSharp/Rules/GenericReadonlyFieldPropertyAssignment.cs
```diff
@@ -108,47 +108,74 @@ public override void Initialize(AnalysisContext context)
...
-                   !typeParameterSymbol.ConstraintTypes.OfType<IErrorTypeSymbol>().Any() &&
-                   !typeParameterSymbol.ConstraintTypes.Any(typeSymbol =>
-                       typeSymbol.IsReferenceType &&
-                       typeSymbol.TypeKind == TypeKind.Class);
+            if (type.TypeKind == TypeKind.Interface)
+            {
+                return true;
+            }
+
+            return GenericParameterMightBeValueType(type as ITypeParameterSymbol);
```
dbolkensteyn: @tvsonar what happens on invalid input where cycles are created (in the IDE)? with a bit of luck our rule doesn't get called in such a case

dbolkensteyn:
LGTM

src/SonarLint.CSharp/Rules/GenericReadonlyFieldPropertyAssignment.cs
tvsonar: I just moved the declaration closer to the usage. I haven't thought about the performance consequences. I'll revert this to first check the syntax, and then the semantics.

src/SonarLint.CSharp/Rules/GenericReadonlyFieldPropertyAssignment.cs
tvsonar: It seems that in case of invalid circular references, Roslyn chooses one of the type parameters and reports on it, and at the same time it excludes from the constraints on other type parameters. Which means in our case that our rule still reports. As an example, you can check this:
```
public class MyClass<T1, T2> where T1 : T2, IPoint where T2 : T1, IPoint
{
    readonly T1 point;  // this could be a struct

    public void Method()
    {
        point.X++; //Noncompliant;
    }
}
```

PR #773, opened 2016-01-21
SLVS-756 Disable all "controversial" rules by default

dbolkensteyn:


src/SonarLint.CSharp/Rules/InsecureEncryptionAlgorithm.cs
```diff
@@ -45,7 +45,7 @@ public class InsecureEncryptionAlgorithm : DiagnosticAnalyzer
         internal const string MessageFormat = "Use the recommended AES (Advanced Encryption Standard) instead.";
         internal const string Category = SonarLint.Common.Category.Security;
         internal const Severity RuleSeverity = Severity.Critical;
-        internal const bool IsActivatedByDefault = false;
+        internal const bool IsActivatedByDefault = true;
```
dbolkensteyn: this one actually is being enabled by default

PR #774, opened 2016-01-25
SLVS-761 Rules should not throw exceptions on invalid inputs

tvsonar:


src/SonarLint.CSharp/Rules/RedundantArgument.cs
```diff
@@ -115,6 +115,19 @@ public ArgumentParameterMapping(ArgumentSyntax argument, IParameterSymbol parame
                 Argument = argument;
                 Parameter = parameter;
             }
+
+            internal static ArgumentParameterMapping GetParameterMapping(ArgumentSyntax argument, MethodParameterLookup methodParameterLookup)
```
tvsonar: As this method and the class is shared by many analyzers, we could put it inside the `MethodParameterLookup` class.

src/SonarLint.CSharp/Helpers/MethodParameterLookup.cs
```diff
@@ -44,18 +45,22 @@ public IMethodSymbol MethodSymbol
             }
         }
 
-        public static IParameterSymbol GetParameterSymbol(ArgumentSyntax argument, ArgumentListSyntax argumentList, IMethodSymbol method)
+        public static bool TryGetParameterSymbol(ArgumentSyntax argument, ArgumentListSyntax argumentList,
+            IMethodSymbol method, out IParameterSymbol parameter)
         {
+            parameter = null;
             if (!argumentList.Arguments.Contains(argument) ||
```
dbolkensteyn: @tvsonar minor but it seems that this should be the last check, as the ones on the method symbol are going to be super fast

src/SonarLint.CSharp/Rules/RedundantArgument.cs
dbolkensteyn: then let's do so?

dbolkensteyn:
@tvsonar I think the `GetParameterMapping()` method should be removed. Could you also update the Jira ticket to make it more user friendly? (i.e. document the patterns that are fixed by this ticket and PR)

dbolkensteyn:
LGTM

PR #775, opened 2016-01-26
SLVS-771 S2486 (empty catch blocks)

tvsonar:


src/SonarLint.CSharp/Rules/CatchEmpty.cs
```diff
@@ -33,15 +35,15 @@ namespace SonarLint.Rules.CSharp
...
-        internal const string Title = "Exceptions should not be ignored";
+        internal const string Title = "Generic exceptions should not be ignored";
         internal const string Description =
             "When exceptions occur, it is usually a bad idea to simply ignore them. Instead, it " +
             "is better to handle them properly, or at least to log them.";
-        internal const string MessageFormat = "Handle the exception, rather than swallow it with an empty statement.";
+        internal const string MessageFormat = "Handle the exception or explain in a comment why it can be ignored.";
         internal const string Category = SonarLint.Common.Category.Reliability;
         internal const Severity RuleSeverity = Severity.Major;
         internal const bool IsActivatedByDefault = false;
```
dbolkensteyn: @tvsonar shouldn't this rule be enabled by default?

src/SonarLint.CSharp/Rules/CatchEmpty.cs
```diff
@@ -61,12 +63,52 @@ public override void Initialize(AnalysisContext context)
                 {
                     var catchClause = (CatchClauseSyntax)c.Node;
 
-                    if (!catchClause.Block.Statements.Any())
+                    if (!HasStatements(catchClause) &&
+                        !HasComments(catchClause) &&
+                        !c.SemanticModel.Compilation.IsTest() &&
```
dbolkensteyn: @tvsonar this should ideally be tested too

tvsonar:
@dbolkensteyn What do you mean by seeing `Noncompliant`? You mean in the test file? I think it's good to use the same test file for both test and non-test assembly checks because this way we can be sure that the input file for the test assembly check contains issues.

dbolkensteyn:
LGTM. @tvsonar yes I now agree that it's OK to test on the same file - I was initially confused by the test method name, thinking that it was a normal `Verify()` that would expect an issue for every single `Noncompliant`

PR #776, opened 2016-01-26
SLVS-776 Make SonarLint compatible with VS 2015 Update 2

tvsonar:


tvsonar:
This change was tested locally on VS 2015 Update 1, and on the VM with VS 2015 Update 2 CTP (14.0.24922.00). On both machines the parametered rules work as expected, and not throwing type load exceptions any more.

At the same time a PR was sent to the Roslyn team to fix the backward compatibility issues: roslyn/pull/8179. If this gets merged, then there's no need to apply the change to our code base. However, there's no harm in adding it.

dbolkensteyn:
@tvsonar as discussed, you could attempt to upgrade the Roslyn version to the beta of 1.2.0 and run the regression tests to verify that there is no other behavioral change?

tvsonar:
One of the required nuget packages is not available yet on nuget.org, so the nuget restore fails. We could specify additional nuget feeds for the restore, but it seems too much work considering that we won't merge that given commit to the master from this PR.

dbolkensteyn:
@tvsonar Did you create an issue on the Roslyn project regarding the handling of new lines? If so, can you please mention it here and in the Jira ticket to reference it later on? Thanks

dbolkensteyn:
LGTM, but can you remove the Roslyn upgrade commit & the SARIF normalizer one from this PR?

tvsonar:
@dbolkensteyn yes, about the new lines, here's the Roslyn issue: https://github.com/dotnet/roslyn/issues/8200

PR #778, opened 2016-01-28
SLVS-769 Rework S3217 (potential invalid cast in foreach)

tvsonar:


src/SonarLint.Extra/Rules.Description/S3217.html
```diff
@@ -11,46 +11,61 @@
```
dbolkensteyn: @tvsonar the reasoning at the top of this file should be updated, it's incorrect - the rule no longer detects usages of old non-generic collections

dbolkensteyn: the `New C# code should use generics ` part is not at all part of the scope of this rule and should be removed from the description.

It should be clear that this rule only runs when generic collections are used and the old implicit cast feature is abused.

src/SonarLint.Extra/Rules.Description/S3217.html
```diff
@@ -11,46 +11,61 @@
...
+  {
+    var fruitBasket = new List<Fruit>();
+    fruitBasket.Add(new Orange());
+    fruitBasket.Add(new Orange());
+    // fruitBasket.Add(new Apple());  // uncommenting this line will make both foreach below throw an InvalidCastException
 
-// can be written in an equivalent but shorter way:
+    foreach (Fruit fruit in fruitBasket)
+    {
+      var orange = (Orange)fruit; // This "explicit" conversion is hidden within the foreach loop
```
dbolkensteyn: @tvsonar I've appended `below` at the end of this comment line to make it explicit that they both are equivalent

dbolkensteyn:
@tvsonar I let you update the rule's description to remove all possible confusion: this rule doesn't apply to old pre-generic collections

dbolkensteyn:
LGTM

PR #781, opened 2016-01-29
SLVS-695 S1006 - Optional parameters should have the same default val…

tvsonar:
…ue in overrides

### DIFF
--- a/src/SonarLint.CSharp/Rules/MethodOverrideChangedDefaultValue.cs
+++ b/src/SonarLint.CSharp/Rules/MethodOverrideChangedDefaultValue.cs
@@ -1,0 +1,108 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015-2016 SonarSource SA
+ * mailto:contact@sonarsource.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarLint.Common;
+using SonarLint.Common.Sqale;
+using SonarLint.Helpers;
+using System.Linq;
+
+namespace SonarLint.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [SqaleConstantRemediation("2min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.DataReliability)]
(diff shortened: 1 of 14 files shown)


src/Tests/SonarLint.UnitTest/TestCases/MethodOverrideChangedDefaultValue.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Tests.Diagnostics
+{
+    public interface IMyInterface
+    {
+        void Write(int i, int j = 5);
+    }
+
+    public class Base : IMyInterface
+    {
+        public virtual void Write(int i, int j = 0) // Compliant, interfaces are ignored
```
tvsonar: @dbolkensteyn Should we include interfaces too?

src/SonarLint.Extra/Rules.Description/S1006.html
```diff
@@ -0,0 +1,64 @@
...
+    }
+}
+
+public class Program
+{
+    public static void Main()
+    {
+        var derived = new Derived();
+        derived.Write(); // writes 5
+        ((Base)derived).Write(); // writes 42; was that expected
```
dbolkensteyn: @tvsonar this seems to unrealistic to be really convincing IMO. could you update this to pass `derived` to a method that takes a `Base` type and calls `.Write()` on it? That way, we show a real, quite tricky bug in action.

src/Tests/SonarLint.UnitTest/TestCases/MethodOverrideChangedDefaultValue.cs
dbolkensteyn: what do you think @tvsonar ? why do you want to ignore them?

src/SonarLint.CSharp/Rules/MethodOverrideChangedDefaultValue.cs
```diff
@@ -0,0 +1,120 @@
...
+                            continue;
+                        }
+                    }
+                },
+                SyntaxKind.MethodDeclaration);
+        }
+
+        private static bool DefaultParameterValueExists(IMethodSymbol methodSymbol)
+        {
+            return methodSymbol.OverriddenMethod.Parameters.Any(p => p.HasExplicitDefaultValue) ||
```
dbolkensteyn: @tvsonar did you mean `&&`  here instead of `||` ? it doesn't make sense to run this rule when only the override method has optional parameters

src/SonarLint.CSharp/Rules/MethodOverrideChangedDefaultValue.cs
```diff
@@ -0,0 +1,120 @@
...
+                methodSymbol.Parameters.Any(p => p.HasExplicitDefaultValue);
```
dbolkensteyn: @tvsonar I could remove this line without breaking the unit test

src/SonarLint.CSharp/Rules/MethodOverrideChangedDefaultValue.cs
```diff
@@ -0,0 +1,120 @@
...
+                        return;
+                    }
+
+                    for (int i = 0; i < methodSymbol.Parameters.Length; i++)
+                    {
+                        var derivedParameter = methodSymbol.Parameters[i];
+                        var overriddenParameter = methodSymbol.OverriddenMethod.Parameters[i];
+
+                        var parameterSyntax = derivedParameter.DeclaringSyntaxReferences
+                            .FirstOrDefault()?.GetSyntax() as ParameterSyntax;
```
dbolkensteyn: @tvsonar can you get rid of the `GetSyntax()` call here? It's not ideal to have it, and it raises the question of why you use `FirstOrDefault()`

dbolkensteyn:
Reopening mainly to discuss whether the rule should apply to interface implementations or not.

src/SonarLint.CSharp/Rules/MethodOverrideChangedDefaultValue.cs
tvsonar: This whole method is just to return early from the analyzer if there's nothing to do there.

dbolkensteyn:
nice change to the code fix, LGTM

PR #782, opened 2016-02-01
SLVS-666 S3450 - Parameters with [DefaultParameterValue] attributes s…

tvsonar:
…hould also be [Optional]

### DIFF
--- a/src/SonarLint.CSharp/Rules/MethodParameterMissingOptional.cs
+++ b/src/SonarLint.CSharp/Rules/MethodParameterMissingOptional.cs
@@ -1,0 +1,118 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015-2016 SonarSource SA
+ * mailto:contact@sonarsource.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarLint.Common;
+using SonarLint.Common.Sqale;
+using SonarLint.Helpers;
+using System.Collections.Generic;
+
+namespace SonarLint.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [SqaleConstantRemediation("2min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.Understandability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags(Tag.Pitfall)]
+    public class MethodParameterMissingOptional : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S3450";
+        internal const string Title = "Parameters with \"[DefaultParameterValue]\" attributes should also be marked with \"[Optional]\"";
+        internal const string Description =
+            "There is no point in providing a default value for a parameter if callers are required to provide a value for it anyway. Thus, " +
+            "\"[DefaultParameterValue]\" should always be used in conjunction with \"[Optional]\".";
+        internal const string MessageFormat = "Add the \"[Optional]\" attribute to this parameter.";
+        internal const string Category = SonarLint.Common.Category.Maintainability;
+        internal const Severity RuleSeverity = Severity.Major;
+        internal const bool IsActivatedByDefault = true;
+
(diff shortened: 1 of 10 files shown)


src/SonarLint.CSharp/Rules/MethodParameterMissingOptional.cs
```diff
@@ -0,0 +1,99 @@
...
+            "\"[DefaultParameterValue]\" should always be used in conjunction with \"[Optional]\".";
+        internal const string MessageFormat = "Add the \"[Optional]\" attribute to this parameter.";
+        internal const string Category = SonarLint.Common.Category.Maintainability;
+        internal const Severity RuleSeverity = Severity.Major;
+        internal const bool IsActivatedByDefault = true;
+        private const IdeVisibility ideVisibility = IdeVisibility.Hidden;
+
+        internal static readonly DiagnosticDescriptor Rule =
+            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category,
+                RuleSeverity.ToDiagnosticSeverity(ideVisibility), IsActivatedByDefault,
```
tvsonar: Faded and underlined code looks weird. Should we modify the severity calculation to report info level issues for major and faded (redundant) code? This way the issue shows up faded in the text editor with no underline, and it is still displayed in the error window, under the messages category. WDYT @dbolkensteyn?

src/SonarLint.Extra/Rules.Description/S3450.html
```diff
@@ -0,0 +1,20 @@
...
+<pre>
+public void MyMethod([DefaultParameterValue(5)] int j) //Noncompliant, useless
+{
+  Console.WriteLine(j);
+}
+</pre>
+
+<h2>Compliant Solution</h2>
+<pre>
+public void MyMethod([DefaultParameterValue(5)][Optional] int j)
```
dbolkensteyn: @tvsonar why not suggest the "normal" syntax of `int j = 5` first here and mention the C#/.NET version requirement?

src/SonarLint.CSharp/Rules/MethodParameterMissingOptionalCodeFixProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+using SonarLint.Common;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Formatting;
+
+namespace SonarLint.Rules.CSharp
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public class MethodParameterMissingOptionalCodeFixProvider : CodeFixProvider
+    {
+        private const string Title = "Add missing \"[Optional]\" attribute";
```
dbolkensteyn: @tvsonar The proper fix is to use `= [the default value]`, not to add the `[Optional]` attribute.

src/SonarLint.CSharp/Rules/MethodParameterMissingOptional.cs
dbolkensteyn: @tvsonar no I think it's fine like that - another major + faded rule is for example the unused parameter one, where it's also valuable to underline the faulty location as it might reveal a bug.

src/SonarLint.CSharp/Rules/MethodParameterMissingOptional.cs
```diff
@@ -0,0 +1,99 @@
...
+using SonarLint.Common.Sqale;
+using SonarLint.Helpers;
+
+namespace SonarLint.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [SqaleConstantRemediation("2min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.Understandability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags(Tag.ApiDesign)]
```
dbolkensteyn: @tvsonar Why do you think this is an API design issue? It seems more like a pitfall to me: this rule points out that you did something half way which is strange

src/SonarLint.CSharp/Rules/MethodParameterMissingOptional.cs
```diff
@@ -0,0 +1,99 @@
...
+                    }
+
+                    var attributes = parameterSymbol.GetAttributes();
+                    var defaultParameterValueAttribute = attributes
+                        .FirstOrDefault(attr => attr.AttributeClass.ToDisplayString() == DefaultParameterValueAttributeName);
+
+                    if (defaultParameterValueAttribute != null &&
+                        !attributes.Any(attr => attr.AttributeClass.ToDisplayString() == OptionalAttributeName))
+                    {
+                        var syntax = defaultParameterValueAttribute.ApplicationSyntaxReference.GetSyntax();
```
dbolkensteyn: what is `ApplicationSyntaxReference` @tvsonar ? I don't see any documentation attached to its declaration

don't you think it would be better to resolve the attribute symbols rather than the parameter one? We don't really need the parameter symbol here anyway

src/SonarLint.CSharp/Rules/MethodParameterMissingOptional.cs
tvsonar: Ohh yes, I think that's annoying too. I think people coming from a R# background are not used to seeing this bloat of effects on the code. But having these issues as info level has the benefit of still seeing the problem in the Error Window.

src/SonarLint.Extra/Rules.Description/S3450.html
tvsonar: Yep, that's a good idea. But we should still keep this variant too, because there must be a reason for using the `DefaultParameterValue` in the first place, for example that the code is (C#3 or before) legacy code.

src/SonarLint.CSharp/Rules/MethodParameterMissingOptionalCodeFixProvider.cs
tvsonar: This only works if C# 4 or upper is targeted. It's probably possible to get this info from the compilation, but I'll have to check.

dbolkensteyn:
LGTM

PR #783, opened 2016-02-01
SLVS-671 S3447 - [Optional] should not be used with ref or out

tvsonar:


src/SonarLint.CSharp/Rules/OptionalRefOutParameter.cs
```diff
@@ -0,0 +1,93 @@
...
+        internal const string DiagnosticId = "S3447";
+        internal const string Title = "\"[Optional]\" should not be used on \"ref\" or \"out\" parameters";
+        internal const string Description =
+            "The use of \"ref\" or \"out\" in combination with \"[Optional]\" is both confusing and contradictory. \"[Optional]\" indicates " +
+            "that the parameter doesn't have to be provided, while out and ref mean that the parameter will be used to return data to the " +
+            "caller. Thus, making it \"[Optional]\" to provide the parameter in which you will be passing back the method results doesn't " +
+            "make sense.";
+        internal const string MessageFormat = "Remove the \"[Optional]\" attribute.";
+        internal const string Category = SonarLint.Common.Category.Maintainability;
+        internal const Severity RuleSeverity = Severity.Critical;
```
dbolkensteyn: @tvsonar why did you make this rule critical? In the end, it seems to me that the `[Optional]` attribute is only going to be unused: The compilation fails if you don't pass the parameter, so there isn't any strange behavior.

So I don't see why it's of a higher importance than this rule for example: https://jira.sonarsource.com/browse/RSPEC-3450

src/SonarLint.CSharp/Rules/OptionalRefOutParameter.cs
```diff
@@ -0,0 +1,93 @@
...
+using SonarLint.Helpers;
+using System.Linq;
+
+namespace SonarLint.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [SqaleConstantRemediation("2min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.Understandability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags(Tag.ApiDesign)]
```
dbolkensteyn: @tvsonar Why api-design? This seems more like a pitfall: something is only half-done, which doesn't make sense. This rule applies also to non-API methods.

src/SonarLint.CSharp/Rules/OptionalRefOutParameter.cs
```diff
@@ -0,0 +1,93 @@
...
+                    var parameterSymbol = c.SemanticModel.GetDeclaredSymbol(parameter);
+                    if (parameterSymbol == null ||
+                        parameterSymbol.RefKind == RefKind.None)
+                    {
+                        return;
+                    }
+
+                    var optionalAttribute = parameterSymbol.GetAttributes()
+                        .FirstOrDefault(attr => attr.AttributeClass.ToDisplayString() == OptionalAttributeName)
+                        ?.ApplicationSyntaxReference
```
dbolkensteyn: @tvsonar what is `ApplicationSyntaxReference`? I'd rather resolve the attributes than the parameter here to avoid this stange `ApplicationSyntaxReference` property and the `GetSyntax()` call

src/SonarLint.CSharp/Rules/OptionalRefOutParameterCodeFixProvider.cs
```diff
@@ -0,0 +1,67 @@
...
+using Microsoft.CodeAnalysis.CodeActions;
+using SonarLint.Common;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace SonarLint.Rules.CSharp
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public class OptionalRefOutParameterCodeFixProvider : CodeFixProvider
+    {
+        internal const string Title = "Remove attribute";
```
dbolkensteyn: @tvsonar "Remove unused [Optional] attribute" might be better, especially if there are multiple attributes on the line

src/SonarLint.CSharp/Rules/OptionalRefOutParameter.cs
```diff
@@ -0,0 +1,93 @@
...
+    public class OptionalRefOutParameter : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S3447";
+        internal const string Title = "\"[Optional]\" should not be used on \"ref\" or \"out\" parameters";
+        internal const string Description =
+            "The use of \"ref\" or \"out\" in combination with \"[Optional]\" is both confusing and contradictory. \"[Optional]\" indicates " +
+            "that the parameter doesn't have to be provided, while out and ref mean that the parameter will be used to return data to the " +
+            "caller. Thus, making it \"[Optional]\" to provide the parameter in which you will be passing back the method results doesn't " +
+            "make sense.";
+        internal const string MessageFormat = "Remove the \"[Optional]\" attribute.";
```
dbolkensteyn: @tvsonar It's a bit of a surprise to see this message in Visual Studio: It would be better here to mention why one should do that, to remove the surprise.

src/SonarLint.CSharp/Rules/OptionalRefOutParameter.cs
```diff
@@ -0,0 +1,93 @@
...
+        internal const string Category = SonarLint.Common.Category.Maintainability;
```
dbolkensteyn: Understandbility sounds better to me, as in https://jira.sonarsource.com/browse/RSPEC-3450

src/SonarLint.CSharp/Rules/OptionalRefOutParameter.cs
tvsonar: @dbolkensteyn that's true, it's just redundant code. The reason why it can still be critical is that if you use the C# 4 syntax to define optional parameters, and you also set the parameter `ref` or `out`, then you get a syntax error (which we could say is a blocker).

Methods that fall under RSPEC-3450 can still be invoked. There the attribute is just simply redundant.

src/SonarLint.CSharp/Rules/OptionalRefOutParameter.cs
tvsonar: I haven't seen this either before, and it seems that the XML comment is only applied on the [overrides](http://source.roslyn.codeplex.com/#Microsoft.CodeAnalysis.CSharp/Symbols/Attributes/AttributeData.cs,6ee6902fa504ec9b). It is analogous to the `DeclaringSyntaxReferences` on symbol declarations.

We can do the attribute resolution, but it adds a couple of lines because we'll need to process lists of `AttributeList`s. While this `GetAttributes()` does exactly what we needed.

src/SonarLint.CSharp/Rules/OptionalRefOutParameterCodeFixProvider.cs
tvsonar: I don't mind, but the attribute will be already underlined, and I don't think there's going to be many attribute removals in a single line for the line based lightbulb tooltip.

dbolkensteyn:
LGTM except for the typo

PR #786, opened 2016-02-02
SLVS-688 Issue message of S107 should properly mention the declaratio…

tvsonar:
…n type

### DIFF
--- a/src/SonarLint.CSharp/Rules/TooManyParameters.cs
+++ b/src/SonarLint.CSharp/Rules/TooManyParameters.cs
@@ -26,6 +26,9 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using SonarLint.Common;
 using SonarLint.Common.Sqale;
 using SonarLint.Helpers;
+using System.Linq;
+using System;
+using System.Collections.Generic;
 
 namespace SonarLint.Rules.CSharp
 {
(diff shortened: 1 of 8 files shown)


src/SonarLint.CSharp/Rules/TooManyParameters.cs
```diff
@@ -76,24 +77,21 @@ public override void Initialize(WrappingAnalysisContext context)
...
+                case SyntaxKind.MethodDeclaration:
+                    return "Method";
+                case SyntaxKind.DelegateDeclaration:
+                case SyntaxKind.AnonymousMethodExpression:
+                    return "Delegate";
+                case SyntaxKind.ParenthesizedLambdaExpression:
+                case SyntaxKind.SimpleLambdaExpression:
+                    return "Lambda";
+                default:
+                    return "Declaration";
```
dbolkensteyn: @tvsonar is this reachable? why not throw if it's unxpected so that we can easily spot it and fix it properly?

dbolkensteyn:
LGTM - but I'm half convinced about updating disabled rules.

For example the issue message still mentions `, which is greater than the {0} authorized.` whereas in Visual Studio there is no way to authorize anything, so there is a bigger issue to be solved: but it's not worth the effort on disabled rules for the time being

tvsonar:
@dbolkensteyn yes, I know we should be focusing on enabled rules. But this doesn't mean that we shouldn't fix known bugs.

PR #787, opened 2016-02-04
SLVS-783 S2737 should not report on exception rethrow, where there's …

tvsonar:
…a comment

### DIFF
--- a/src/SonarLint.CSharp/Rules/CatchRethrow.cs
+++ b/src/SonarLint.CSharp/Rules/CatchRethrow.cs
@@ -26,6 +26,7 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using SonarLint.Common;
 using SonarLint.Common.Sqale;
 using SonarLint.Helpers;
+using System.Linq;
 
 namespace SonarLint.Rules.CSharp
 {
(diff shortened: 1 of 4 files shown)


dbolkensteyn:
I don't see why we would want to allow comments here. R# reports even if there are comments. I can't think of any good reason to simply rethrow from the last case. Most likely the actual error handling logic has yet to be added (or the developer changed his mind), and in both cases that's some real technical debt which should be tracked.

Could you try to find the original cases back in Roslyn that lead to the creation of this ticket and summarize them in https://jira.sonarsource.com/browse/SLVS-783 description?

Also, if you'd still like to make this change, you'll have to update the non compliant code example in https://jira.sonarsource.com/browse/RSPEC-2737 , because the `// Noncompliant` itself will make it compliant. Therefore, a user copy & pasting that noncompliant in Visual Studio won't see any issue. Moreover, the exception should also be documented there as such.

BTW I've created the following ticket to handle some false negatives better: https://jira.sonarsource.com/browse/SLVS-793

tvsonar:
@dbolkensteyn I couldn't find it, so I'll just close this PR and the corresponding ticket. (This shows that we shouldn't create tickets without code examples or exact pointers in the issue description.)

PR #788, opened 2016-02-04
SLVS-784 S3236 should not report on cases when the passed value is a …

tvsonar:
…parameter with caller info attribute


dbolkensteyn:
LGTM

PR #790, opened 2016-02-05
SLVS-779 S3251 - Implementations should be provided for "partial" met…

tvsonar:
…hods

### DIFF
--- a/src/SonarLint.CSharp/Rules/PartialMethodNoImplementation.cs
+++ b/src/SonarLint.CSharp/Rules/PartialMethodNoImplementation.cs
@@ -1,0 +1,119 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015-2016 SonarSource SA
+ * mailto:contact@sonarsource.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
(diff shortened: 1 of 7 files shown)


src/SonarLint.CSharp/Rules/PartialMethodNoImplementation.cs
```diff
@@ -0,0 +1,119 @@
...
+                {
+                    var invocation = (InvocationExpressionSyntax)c.Node;
+                    var methodSymbol = c.SemanticModel.GetSymbolInfo(invocation).Symbol as IMethodSymbol;
+                    if (methodSymbol == null)
+                    {
+                        return;
+                    }
+
+                    // from the method symbol it's not possible to tell if it's a partial method or not.
+                    // https://github.com/dotnet/roslyn/issues/48
```
dbolkensteyn: looks like we're waiting for a ticket that's in a "Closed" state to get resolved to improve the code below

src/SonarLint.CSharp/Rules/PartialMethodNoImplementation.cs
```diff
@@ -0,0 +1,119 @@
...
+                        .OfType<MethodDeclarationSyntax>()
+                        .Where(method => method.Body == null && method.Modifiers.Any(m => m.IsKind(SyntaxKind.PartialKeyword)));
+
+                    if (methodSymbol.PartialImplementationPart == null &&
+                        partialDeclarations.Any())
+                    {
+                        var statement = invocation.Parent as StatementSyntax;
+                        if (statement == null)
+                        {
+                            return;
```
dbolkensteyn: @tvsonar can this case happen? wouldn'it it also be noncompliant if it could?

dbolkensteyn:
LGTM

src/SonarLint.CSharp/Rules/PartialMethodNoImplementation.cs
tvsonar: Added a comment/question to that thread, we'll see if the Roslyn team provides an easier/faster solution to this issue.

src/SonarLint.CSharp/Rules/PartialMethodNoImplementation.cs
tvsonar: It can only happen in invalid code. In that case we don't want to report on it. As an example:
```
partial void M();
public void Mm()
{
    var x = M(); // this is invalid. M() is void return.
}
```

PR #792, opened 2016-02-05
SLVS-703 S3532 - "default" clause of a "switch" should not be empty

tvsonar:


tvsonar:
Note to self: the wording of the RSPEC has been updated since, before merging it to master update the description and the HTML.

dbolkensteyn:
@tvsonar can you disable this rule in sonarqube?

dbolkensteyn:
also there is no need to check whether the `default: break;` contains comments: even with a comment, it's completely useless and should be highlighted as such (also R#'s behavior)

dbolkensteyn:
I'd also remove the SQALE remediation cost on this rule - I don't believe there is real technical debt associated to this rule. it will be up to the developer to decide whether or not he prefers to keep the default block.

PR #793, opened 2016-02-05
SLVS-782 S1449 reports false positives on "string.Compare" with "Stri…

tvsonar:
…ngComparison"


src/SonarLint.CSharp/Rules/StringOperationWithoutCulture.cs
```diff
@@ -94,33 +94,19 @@ public override void Initialize(AnalysisContext context)
...
-                    if (cultureInfoType == null || compareOptionsType == null)
-                    {
-                        return;
-                    }
-
                     if (calledMethod.ContainingType.SpecialType == SpecialType.System_String &&
                         calledMethod.Name == CompareMethodName &&
                         !calledMethod.Parameters
-                            .Any(param => param.Type.Equals(cultureInfoType) || param.Type.Equals(compareOptionsType)))
+                            .Any(param => StringCultureSpecifierNames.Contains(param.Type.ToDisplayString())))
```
dbolkensteyn: @tvsonar could you update https://jira.sonarsource.com/browse/SLVS-782 to explain a bit why the previous code wasn't working? at first, both version would seem to be equivalent. This way we won't forget if we have to find this information back in 6 months from now.

dbolkensteyn:
LGTM, but please explain a bit what's happening as I have myself no clue why you had to change the code here.

tvsonar:
Added the explanation to the Jira ticket.

PR #794, opened 2016-02-08
SLVS-673 S3444 - Interfaces with colliding, inherited members should …

tvsonar:
…explicitly redefine interface members

### DIFF
--- a/src/SonarLint.CSharp/Rules/InheritedCollidingInterfaceMembers.cs
+++ b/src/SonarLint.CSharp/Rules/InheritedCollidingInterfaceMembers.cs
@@ -1,0 +1,210 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015-2016 SonarSource SA
+ * mailto:contact@sonarsource.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarLint.Common;
+using SonarLint.Common.Sqale;
+using SonarLint.Helpers;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace SonarLint.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [SqaleConstantRemediation("5min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.Understandability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags(Tag.Design)]
+    public class InheritedCollidingInterfaceMembers : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S3444";
+        internal const string Title = "Interfaces with colliding, inherited members should explicitly redefine interface members";
+        internal const string Description =
+            "When an interface \"IDerived\" inherits from two interfaces \"IBase1\" and \"IBase2\" that both define a member " +
+            "\"SomeProperty\", calling \"IDerived.SomeProperty\" will result in the compiler error \"CS0229 Ambiguity between " +
+            "'IBase1.SomeProperty' and 'IBase2.SomeProperty'\". Every caller will be forced to cast instances of \"IDerived\" " +
+            "to \"IBase1\" or \"IBase2\" to resolve the ambiguity and to be able to access \"SomeProperty\". Instead, it is " +
+            "better to resolve the ambiguity on the definition of \"IDerived\" either by: renaming one of the \"SomeProperty\" " +
+            "in \"IBase1\" or \"IBase2\" to remove the collision or by also defining a new \"SomeProperty\" member on " +
+            "\"IDerived\". Use the latter only if all \"SomeProperty\" are meant to hold the same value.";
+        internal const string MessageFormat = "Rename or add member{1} {0} to this interface to resolve ambiguities.";
+        internal const string Category = SonarLint.Common.Category.Design;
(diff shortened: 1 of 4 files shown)


src/SonarLint.Extra/Rules.Description/S3444.html
```diff
@@ -0,0 +1,72 @@
+<p>
+    When an interface implements two other interfaces that have members with the same signature, confusion is almost guaranteed; when access
+    is made through an implementing class, which one was intended? Further, cast the implementing class to the interface it implemented, and
```
dbolkensteyn: @tvsonar I've tried to read paragraph 3 times, and I can't manage to understand it. It's very dense, and requires you to imagine the code.

It might be much better to use some code to explain what the issue is - rather than on a lot of text.

dbolkensteyn: I've updated the RSPEC: https://jira.sonarsource.com/browse/RSPEC-3444

src/Tests/SonarLint.UnitTest/TestCases/InheritedCollidingInterfaceMembers.cs
```diff
@@ -0,0 +1,82 @@
...
+
+    public interface IMyInterfaceGeneric1
+    {
+        unsafe string Method<T1>(T1[,] parameter, int* pointer);
+    }
+    public interface IMyInterfaceGeneric2
+    {
+        unsafe int Method<T2>(T2[,] parameter, int* pointer);
+    }
+    public interface IMyInterfaceGenericCommon1 : IMyInterfaceGeneric1, IMyInterfaceGeneric2 // Non-compliant, but not recognized
```
dbolkensteyn: @tvsonar R# detects this case, can we now also detect it with the newly discovered Roslyn APIs perhaps?

src/SonarLint.CSharp/Rules/InheritedCollidingInterfaceMembers.cs
```diff
@@ -0,0 +1,204 @@
...
+                        var pluralize = collidingMembers.Count > 1 ? "s" : string.Empty;
+
+                        c.ReportDiagnostic(Diagnostic.Create(Rule, interfaceDeclaration.Identifier.GetLocation(),
+                            membersText, pluralize));
+                    }
+                },
+                SyntaxKind.InterfaceDeclaration);
+        }
+
+        private const int MaxMemberDisplayCount = 2;
```
dbolkensteyn: @tvsonar i'd expect consts & fields to be declared before any method

src/Tests/SonarLint.UnitTest/TestCases/InheritedCollidingInterfaceMembers.cs
dbolkensteyn: the compiler warning CS0108 reports that `new` should be used if you add a method `unsafe int Method<T2>(T2[,] parameter, int* pointer);`

it would be interesting to see how that one is implemented to see if we can do something similar:
https://github.com/dotnet/roslyn/blob/8eb34a73f658fd9017f53763152570d59cbb0bde/src/Compilers/CSharp/Portable/Symbols/Source/SourceMemberContainerSymbol_ImplementationChecks.cs#L540

src/SonarLint.CSharp/Rules/InheritedCollidingInterfaceMembers.cs
tvsonar: @dbolkensteyn I moved it further up, but even in this case it will be below the `SupportedDiagnostics` property. And there are many contradicting cases already: https://github.com/SonarSource-VisualStudio/sonarlint-roslyn-analyzer/blob/master/src/SonarLint.CSharp/Metrics/Metrics.cs#L119

src/Tests/SonarLint.UnitTest/TestCases/InheritedCollidingInterfaceMembers.cs
tvsonar: @dbolkensteyn Here is what I found:

For whatever reason `syntaxNode.GetDiagnostics` always returns an empty collection of diagnostics. `semanticModel.GetDiagnostics(syntaxNode.Span)` on the other hand returns what we expected, and we could filter for the ID.

However modifying the derived interface would not simply solve our problem. Consider the following code
```C#
public interface IBase1<TX>
{
    string MethodGen<T1>(T1[,] parameter, TX[] arr); // MethodA
    void M();
}
public interface IBase2<TY>
{
    int MethodGen<T2>(T2[,] parameter, TY[] arr); // MethodB
}
public interface IDerived<T> : IBase1<T>, IBase2<T> // Noncompliant
{
}
```
Adding method `M()` to `IDerived` would result in `CS0108`, so this is not a good indication of colliding methods. Instead we should add method `M` to `IBase2` and check if that generates a new `CS0111` ("type already defines a member with same name and parameter types"). If so then we have a colliding member.

However adding a generic method, such as `MethodGen`, from `IBase1` to `IBase2` is not as simple as a non-generic one, because the type parameter names need to be adjusted. In general we don't know which one should be mapped to which one, but because we are processing a derived interface, there are already some type argument substitutions.

So it seems easier to add two possibly colliding methods to `IDerived` and check if one of them generates a `CS0111` and both the `CS0108`. In this case `TX` and `TY` would already be mapped to `T` in the semantic model, so we can do that change on the inserted methods. (We would need to get all identifiers from the method declaration, and check if any of them is mapped to `TX` or not. If so, then change it to `T`.)

At this point we have possibly open generic methods, but generic type parameters would be defined on the method level, so we don't have to care about their names.

Here is the result of what we should be checking for (on a bit more complex example, where there's a generic container class as well):
```C#
public class Container1<TT>
{
    public interface IBase1<TX>
    {
        string MethodGen<T1>(T1[,] parameter, IEnumerable<TX> en, TT[] arr) where T1: class; // MethodA
    }
}
public class Container2<TTT>
{
    public interface IBase2<TY>
    {
        int MethodGen<T2>(T2[,] parameter, IEnumerable<TY> en, TTT[] arr) where T2 : struct; // MethodB
    }
}

public interface IDerived<TC, T> : Container1<TC>.IBase1<T>, Container2<TC>.IBase2<T> // Noncompliant
{
    // TC is mapped to TT, T is mapped to TX, we can drop the type constraint
    string MethodGen<T1>(T1[,] parameter, IEnumerable<T> en, TC[] arr);    // check for CS0108
    // TC is mapped to TTT, T is mapped to TY
    int MethodGen<T2>(T2[,] parameter, IEnumerable<T> en, TC[] arr);       // check for CS0108, CS0111
}
``` 

So to conclude, as far as I see it, I think this improvement could be done by using the `ReplaceSyntaxTree()`, but I'm not convinced about the value considering the required effort. I would probably just create a ticket, and implement this check later. What do you think?

dbolkensteyn:
LGTM @tvsonar . As discussed please do create an issue on Roslyn. Instead of create an issue on our repo, you can reference this PR. In any case, we'll also need a Jira ticket to not forget to come back to this in the future.

dbolkensteyn:
BTW I am wondering if this alternative idea could work: Instead of trying to edit interfaces, checking if a call to `IDerived.M()` results in a compiler error about ambiguous methods might be closer to want we initially wanted. WDYT @tvsonar ?

tvsonar:
@dbolkensteyn the ticket was already created (https://jira.sonarsource.com/browse/SLVS-814), I added your idea of generating a call.

PR #795, opened 2016-02-08
SLVS-678 S3441 - Redundant property names should be omitted in anonym…

tvsonar:
…ous classes

### DIFF
--- a/src/SonarLint.CSharp/Rules/RedundantPropertyNamesInAnonymousClass.cs
+++ b/src/SonarLint.CSharp/Rules/RedundantPropertyNamesInAnonymousClass.cs
@@ -1,0 +1,94 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015-2016 SonarSource SA
+ * mailto:contact@sonarsource.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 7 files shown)


src/SonarLint.Extra/Rules.Description/S3441.html
```diff
@@ -0,0 +1,40 @@
...
+
+var anon = new
+{
+  X = X, //Noncompliant, the new object would have the same property without the "X =" part.
+  Y = Y  //Noncompliant
+};
+
+var anon2 = new
+{
+  X = X, // Ignored, condensing this would make the code harder to read
```
dbolkensteyn: @tvsonar I don't see a good reason to ignore this case, R# does detect it, and it is redundant.

This rule is also a findings one that simply highlights everything that's redundant. Developers will decide whether or not to change their code depending on their own preferences. I personally even find the first case more readable and probably would never remove the redundencies found by this rule.

src/SonarLint.CSharp/Rules/RedundantPropertyNamesInAnonymousClass.cs
```diff
@@ -0,0 +1,97 @@
...
+using SonarLint.Helpers;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace SonarLint.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [SqaleConstantRemediation("2min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.Readability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
```
dbolkensteyn: @tvsonar as this is a finding rule, it shouldn't be enabled in SQ

dbolkensteyn:
LGTM but see remark about this rule being a findings one

PR #796, opened 2016-02-09
SLVS-347 S1698 "==" should not be used when "Equals" is overridden

tvsonar:


dbolkensteyn:
@tvsonar Please review this commit which removes the use of the compilation end event, making the rule much more responsible in the IDE on large projects, and also enabling solution wide analysis: https://github.com/dbolkensteyn/sonarlint-vs/commit/916df59a624d5d4724a18bc96e8fd6d6544a6bcf

dbolkensteyn:
On AppVeyor, the CPU time required to run this rule in batch increased from:
```
SonarLint.Rules.CSharp.ReferenceEqualityCheckWhenEqualsExists            0.059
```
to
```
SonarLint.Rules.CSharp.ReferenceEqualityCheckWhenEqualsExists              6.3
```

but the IDE, the responsivness seems really good & fast.

PR #800, opened 2016-02-11
Avoid merge conflicts by using wildcards in *.csproj

dbolkensteyn:
Merge conflicts are super annoying, for many reasons:
1. AppVeyor refuses to build PR which have merge conflicts
2. When you want to merge and there's a conflict, you need to manually resolve it

Most merge conflicts in SLVS mostly occur in *.csproj files, because each newly added source files gets explicitly listed in them.

Moreover, those automated edits of *.csproj files also pollute the diff of new rules pull requests.


tvsonar:
@dbolkensteyn It looks good to me. But we should do this change in the SonarLint.VisualBasic.csproj as well.

PR #801, opened 2016-02-12
SLVS-702 S3257 - Redundancies in declarations

tvsonar:


src/SonarLint.CSharp/Rules/RedundantDeclaration.cs
```diff
@@ -0,0 +1,405 @@
...
+using Microsoft.CodeAnalysis.Text;
+using System.Linq;
+using System;
+
+namespace SonarLint.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [SqaleConstantRemediation("1min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.Readability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
```
dbolkensteyn: @tvsonar this seems to be a findings rule that shouldn't be enabled in SonarQube

src/SonarLint.CSharp/Rules/RedundantDeclarationCodeFixProvider.cs
```diff
@@ -0,0 +1,229 @@
...
+using System;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Formatting;
+
+namespace SonarLint.Rules.CSharp
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public class RedundantDeclarationCodeFixProvider : CodeFixProvider
+    {
+        internal const string TitleRedundantArraySize = "Remove redundant array size";
```
dbolkensteyn: @tvsonar when I choose to fix this particular (array size) in the whole solution, I expect only array size issues to get resolved. AFAIR this is achievable using the "Title" scope or something like that.

src/SonarLint.CSharp/Rules/RedundantDeclaration.cs
```diff
@@ -0,0 +1,405 @@
...
+            {
+                return;
+            }
+
+            var newParameterList = SyntaxFactory.ParameterList(
+                SyntaxFactory.SeparatedList(lambda.ParameterList.Parameters.Select(p => SyntaxFactory.Parameter(p.Identifier))));
+            var newLambda = lambda.WithParameterList(newParameterList);
+
+            SemanticModel newSemanticModel;
+            newLambda = ChangeSyntaxElement(lambda, newLambda, c.SemanticModel, out newSemanticModel);
```
dbolkensteyn: nice trick!

dbolkensteyn:
LGTM - but see the remark about the findings nature of this rule.

BTW, it might be a good idea in the next sprint to start with higher value rules even if they seem more complex, because unfortunately this one seemed to have been quite a lot of work (whereas I guess it seemed easy at first :p), and in the end it won't benefit SonarQube, nor generate nice issues we can hope to blog on

tvsonar:
Finding rules also provide a good value in SonarLint. The fading is not obtrusive at all, so the developer experience is good, and we can provide code fixes for them, which helps removing these issue with one click in the entire solution.

PR #804, opened 2016-02-16
SLVS-801 S3265 - Non-flags enums should not be used in bitwise operat…

tvsonar:
…ions

### DIFF
--- a/src/SonarLint.CSharp/Rules/NonFlagsEnumInBitwiseOperationCodeFixProvider.cs
+++ b/src/SonarLint.CSharp/Rules/NonFlagsEnumInBitwiseOperationCodeFixProvider.cs
@@ -1,0 +1,95 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015-2016 SonarSource SA
+ * mailto:contact@sonarsource.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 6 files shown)


src/SonarLint.CSharp/Rules/NonFlagsEnumInBitwiseOperationCodeFixProvider.cs
```diff
@@ -0,0 +1,85 @@
...
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            var diagnostic = context.Diagnostics.First();
+            var diagnosticSpan = diagnostic.Location.SourceSpan;
+            var node = root.FindNode(diagnosticSpan, getInnermostNodeForTie: true);
+
+            var semanticModel = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+            var operation = semanticModel.GetSymbolInfo(node).Symbol as IMethodSymbol;
+            var enumDeclaration = operation?.ReturnType?.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax(context.CancellationToken)
```
dbolkensteyn: @tvsonar the code fix is being offered also when the enum is a different file in the same project, whereas that is not working. would it be possible to filter that case out? (perhaps by checking that both `enumDeclaration` and `node` have the same `SyntaxTree`?)

dbolkensteyn:
LGTM

src/SonarLint.CSharp/Rules/NonFlagsEnumInBitwiseOperationCodeFixProvider.cs
tvsonar: @dbolkensteyn I fixed this issue. Now the solution is modified and not just a single document.

PR #805, opened 2016-02-16
SLVS-805 S3451 - "[DefaultValue]" should not be used with "[Optional]…

tvsonar:
…" parameters

### DIFF
--- a/src/SonarLint.Extra/Rules.Description/S3451.html
+++ b/src/SonarLint.Extra/Rules.Description/S3451.html
@@ -1,0 +1,37 @@
+<p>
+    The use of <code>[DefaultValue]</code> with <code>[Optional]</code> has no more effect than <code>[Optional]</code> alone. That's because 
+    <code>[DefaultValue]</code> doesn't actually do anything; it merely indicates the intent for the value. More than likely, 
+    <code>[DefaultValue]</code> was used in confusion instead of <code>[DefaultParameterValue]</code>.
+</p>
+
+<h2>Noncompliant Code Example</h2>
+<pre>
+class MyClass
+{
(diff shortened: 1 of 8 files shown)


src/SonarLint.Extra/Rules.Description/S3451.html
```diff
@@ -0,0 +1,37 @@
+<p>
+    The use of <code>[DefaultValue]</code> with <code>[Optional]</code> has no more effect than <code>[Optional]</code> alone. That's because 
+    <code>[DefaultValue]</code> doesn't actually do anything; it merely indicates the intent for the value. More than likely,
```
dbolkensteyn: @tvsonar The following part of the description sounds wrong and seems to actually be useless: `That's because <code>[DefaultValue]</code> doesn't actually do anything; it merely indicates the intent for the value. `

Could you remove it?

dbolkensteyn:
@tvsonar I don't think this rule should report when there already is a `[DefaultParameterValue]` attribute along with the `[Optional]` one (even if R# does). The codefix will also generate invalid code in such cases. I've updated rule's title accordingly.

src/SonarLint.Extra/Rules.Description/S3451.html
tvsonar: @dbolkensteyn I think this statement is correct. MSDN says "A DefaultValueAttribute will not cause a member to be automatically initialized with the attribute's value. You must set the initial value in your code". The attribute is only there to help visual designers or code generators specify a default value.

dbolkensteyn:
LGTM

It's best to not squash commits during the reviewing to make it easier to see the changes

PR #791, opened 2016-02-05
S134 Limit nesting depth of control flow statements in functions

p3pijn:
I have made an attempt to implement the rule _Limit nesting depth of control flow statements in functions_ (https://jira.sonarsource.com/browse/RSPEC-134).
I previously suggested this rule in issue #684 in the old repository.

This is my first PR in this repo. Feedback is really appreciated!

Issue #684, opened 2015-12-04
Limit nesting depth of control flow statements in functions


p3pijn:
I would like to suggest a new rule which limits the nesting depth of control flow statements to a certain value.

Many software quality standards describe a maximum nesting depth of control flow statements per function. An example of this the IfSQ level 2 standard for maintainable source code (http://www.ifsq.org/level-2.html), which prescribes a maximum nesting depth of 4 (see: http://www.ifsq.org/indicator-SP-2.html).
Another well known set of coding guidelines that limit the nesting depth to 1 control structure are the Object Calisthenics by Jeff Bay (http://www.cs.helsinki.fi/u/luontola/tdd-2009/ext/ObjectCalisthenics.pdf).

Although this rule is slightly related to limiting the cyclomatic complexity of functions to a certain value, being able to limit the nesting depth at the function level still provides a valuable addition. For example, when the nesting depth of control flow statements of a function exceeds 4 the function is considered to be hard to understand, while the cyclomatic complexity of the function might still fall within the acceptable range of 5 to 10.

I would suggest a default value of 4 for the maximum nesting depth at the function level (in accordance to the IfSQ level 2 standard). Whereas users should be able to change the value according to their needs.

tvsonar:
Thanks @p3pijn. I just did a searhc in our rule repository, and found the following rule, which covers what you propose: https://jira.sonarsource.com/browse/RSPEC-134. So we can implement it relatively quickly.

PR #807, opened 2016-02-17
S134 Limit nesting depth of control flow statements in functions (new attempt)

p3pijn:
Continued from #791... (I messed up the merge)

> I have made an attempt to implement the rule Limit nesting depth of control flow statements in functions (https://jira.sonarsource.com/browse/RSPEC-134).
> I previously suggested this rule in issue #684 in the old repository.

> I have enhanced this PR according to the feedback.

Once I have solved my issue with running the automated regression test I can update this PR with the expected json files as well. Help on the issue is greatly appreciated:
> The `JavaScriptSerializer` is complaining about unrecognized escape sequences in the generated json files. I've manually analyzed the json files and all `\` and `"` characters have indeed not been escaped. 
> Any clue why the content in my json files might not be properly escaped?


p3pijn:
I've been able to succesfully run the regression test after upgrading VS2015 to Update 1.
The regression test is updated with the expected results for this new rule.

I'm looking forward to your feedback.

tvsonar:
Cool. I'll have a look this week.

src/SonarLint.Extra/Rules.Description/S134.html
```diff
@@ -0,0 +1,5 @@
+﻿<p>
```
tvsonar: Rule descriptions should be exactly the same as the ones specified in RSPEC. I've customized the original one for C#: https://jira.sonarsource.com/browse/RSPEC-3534

src/SonarLint.CSharp/Rules/FunctionNestingDepth.cs
```diff
@@ -0,0 +1,155 @@
...
+using SonarLint.Common.Sqale;
+using SonarLint.Helpers;
+using System;
+using System.Collections.Generic;
+
+namespace SonarLint.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [SqaleConstantRemediation("10min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.UnitTestability)]
```
tvsonar: This should be `LogicChangeability` as specified in RSPEC (https://jira.sonarsource.com/browse/RSPEC-134)

src/SonarLint.CSharp/Rules/FunctionNestingDepth.cs
```diff
@@ -0,0 +1,155 @@
...
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags(Tag.BrainOverload)]
+    public class FunctionNestingDepth : ParameteredDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S134";
+        internal const string Title = "Control flow statements \"if\", \"for\", \"foreach\", \"do\", \"while\", \"switch\" and \"try\" should not be nested too deeply";
+        internal const string Description =
+           "The control flow statement nesting depth of a function should not exceed a defined threshold. "
```
tvsonar: This short description is not specified in RSPEC, I tend to use the first one or two paragraphs of the description. In this case this could be: `Nested "if", "switch", "for", "foreach", "while", "do", and "try" statements are key ingredients for making what's known as "Spaghetti code". Such code is hard to read, refactor and therefore maintain.`

src/SonarLint.CSharp/Rules/FunctionNestingDepth.cs
```diff
@@ -0,0 +1,155 @@
...
+            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category,
+                RuleSeverity.ToDiagnosticSeverity(), IsActivatedByDefault,
+                helpLinkUri: DiagnosticId.GetHelpLink(),
+                description: Description);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Rule); } }
+
+        private const int DefaultValueMaximum = 3;
+
+        [RuleParameter("maximumControlFlowNestingDepthThreshold", PropertyType.Integer,
```
tvsonar: Metadata should match the one specified in RSPEC: `max`, `Maximum allowed control flow statement nesting depth.`

src/SonarLint.CSharp/Rules/FunctionNestingDepth.cs
```diff
@@ -0,0 +1,155 @@
...
+    [Tags(Tag.BrainOverload)]
+    public class FunctionNestingDepth : ParameteredDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S134";
+        internal const string Title = "Control flow statements \"if\", \"for\", \"foreach\", \"do\", \"while\", \"switch\" and \"try\" should not be nested too deeply";
+        internal const string Description =
+           "The control flow statement nesting depth of a function should not exceed a defined threshold. "
+            + "Nested control statements are a key ingredient for making what's known as Spaghetti code. "
+            + "Such code is hard to read, refactor and therefore maintain.";
+        internal const string MessageFormat = "The control flow statement nesting depth of this method is {1} which is greater than {0} authorized.";
```
tvsonar: RSPEC specifies `Refactor this code to not nest more than {X} control flow statements.` as the message.

src/SonarLint.CSharp/Rules/FunctionNestingDepth.cs
```diff
@@ -0,0 +1,155 @@
...
+            SyntaxKind.ForStatement,
+            SyntaxKind.ForEachStatement,
+            SyntaxKind.DoStatement,
+            SyntaxKind.WhileStatement,
+            SyntaxKind.SwitchStatement,
+            SyntaxKind.TryStatement
+        };
+
+        private static Func<SyntaxNode, bool> IsNestingIncreasingKind => node => NestingIncreasingKinds.Contains(node.Kind());
+        private static Func<SyntaxNode, bool> IsPartOfElseClause => node => node.Parent != null && node.Parent.IsKind(SyntaxKind.ElseClause);
```
tvsonar: This can be specified as a function, it is not being passed around.

src/SonarLint.CSharp/Rules/FunctionNestingDepth.cs
```diff
@@ -0,0 +1,155 @@
...
+            SyntaxKind.IfStatement,
+            SyntaxKind.ForStatement,
+            SyntaxKind.ForEachStatement,
+            SyntaxKind.DoStatement,
+            SyntaxKind.WhileStatement,
+            SyntaxKind.SwitchStatement,
+            SyntaxKind.TryStatement
+        };
+
+        private static Func<SyntaxNode, bool> IsNestingIncreasingKind => node => NestingIncreasingKinds.Contains(node.Kind());
```
tvsonar: This can be specified as a function.

src/SonarLint.CSharp/Rules/FunctionNestingDepth.cs
```diff
@@ -0,0 +1,155 @@
...
+
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c => CheckFunctionNestingDepth<AccessorDeclarationSyntax>(c, a => a.Keyword.GetLocation(), b => b.Body),
+                SyntaxKind.GetAccessorDeclaration,
+                SyntaxKind.SetAccessorDeclaration,
+                SyntaxKind.AddAccessorDeclaration,
+                SyntaxKind.RemoveAccessorDeclaration);
+        }
+
+        private void CheckFunctionNestingDepth<TSyntax>(SyntaxNodeAnalysisContext c, Func<TSyntax, Location> location, Func<TSyntax, BlockSyntax> functionBody)
```
tvsonar: It's not the best to report on the function, because it can be huge, and then we might be far from the real location of the issue. I think what is proposed in the rule (and implemented for example in Java) is better, namely that the issue is reported on the violating control flow element. 

Reporting the issue on the method gives the feeling that this is a metric calculated for each method. But reporting it exactly on all violating control flow constructs make the rule more actionable, like refactor this code here.

Note also, that we should report issues on violating constructs that are on exactly the nesting level `Maximum + 1`, so reporting on level 6 is wrong, if we already reported on level 4. Check the noncompliant example at https://jira.sonarsource.com/browse/RSPEC-3534.

src/SonarLint.CSharp/Rules/FunctionNestingDepth.cs
```diff
@@ -0,0 +1,155 @@
...
+                return 0;
+            }
+
+            var controlFlowStatements = body.DescendantNodes().Where(IsNestingIncreasingKind);
+            if (!controlFlowStatements.Any())
+            {
+                return 0;
+            }
+
+            var statementNestingDepths = controlFlowStatements.Select(statement => GetNestingDepthForControlStatement(statement, body));
```
tvsonar: The `DescendantNodes()` call traverses the tree downwards, and then the `GetNestingDepthForControlStatement` goes up. It's not ideal. Considering that we should report on the control flow statements and not the method, this can be avoided. What comes to mind for the solution is the `CSharpSyntaxWalker`. It's not used anywhere in SonarLint yet, but you can check it out [here](https://joshvarty.wordpress.com/2014/07/26/learn-roslyn-now-part-4-csharpsyntaxwalker/). Basically you would be doing a tree traversal, override the control flow element visits, such as `VisitForStatement`, increase a counter before, decrease the counter after the `base.VisitForStatement(node);`. If you reached the limit, then you'd report on the syntax node, and would not visit any further element below the current one.

This raises a question: Do lambda expressions reset the nesting counter or not? So in the following code what is the nesting level at point `A`? Is it 2 or is it reset to 1? What do you think @dbolkensteyn? 
```C#
if (true) // Nesting level 1
{
  Func<int> f = () => 
  {
    if (true)  // A
    {
    }
  };
}
```

src/SonarLint.CSharp/Rules/FunctionNestingDepth.cs
```diff
@@ -0,0 +1,155 @@
...
+            SyntaxKind.TryStatement
+        };
+
+        private static Func<SyntaxNode, bool> IsNestingIncreasingKind => node => NestingIncreasingKinds.Contains(node.Kind());
+        private static Func<SyntaxNode, bool> IsPartOfElseClause => node => node.Parent != null && node.Parent.IsKind(SyntaxKind.ElseClause);
+
+        public override void Initialize(WrappingAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c => CheckFunctionNestingDepth<MethodDeclarationSyntax>(c, m => m.Identifier.GetLocation(), b => b.Body),
```
tvsonar: Checking only the `Body` is not sufficient, we need to check the `ExpressionBody` too, where an expression might embed a lambda, with nested control flow statements:
```C#
public int M1() => (new Func<int>(() => 
{ 
if (true)
  if(true)
    return 0;
return 42; 
}))();
```

its/expected/akka.net/Akka-{5DEDDF90-37F0-48D3-A0B0-A5CBD8A7E377}-S134.json
```diff
@@ -0,0 +1,490 @@
...
+"isSuppressedInSource":  "False"
+}
+},
+{
+"ruleId":  "S134",
+"locations":  [
+{
+"analysisTarget":  [
+{
+"uri":  "akka.net\\src\\core\\Akka\\IO\\InetAddressDnsResolver.cs",
```
tvsonar: I'll need to go through the issues found by the rule. I just did a random pass now, and I had a hard time finding what the issue was with this given method. I could hardly find the `if` inside the `catch`. Reporting on the exact issue location would help a lot.

src/SonarLint.CSharp/Rules/FunctionNestingDepth.cs
p3pijn: I did indeed consider the `CSharpSyntaxWalker` but decided not to use it since it would introduce a whole new mechanism for tree traversal just for this rule. However, an implementation using the `CSharpSyntaxWalker` is much closer to the current Java implementation of this rule ([NestedIfStatementsCheck.java](https://github.com/SonarSource/sonar-java/blob/a7371a09a447714290de977ad02c7e97de559703/java-checks/src/main/java/org/sonar/java/checks/NestedIfStatementsCheck.java)).

tvsonar:
@p3pijn I'm hoping that you also see the value in reporting on the exact issue location and not just on the method name. And as for the rule metadata (title, SCALE mapping, description, ...) we try to match character-to-character what's in RSPEC.

src/SonarLint.CSharp/Rules/FunctionNestingDepth.cs
tvsonar: Yep, we didn't use it before. Maybe we could have. In this case, `CSharpSyntaxWalker` seems like a perfect fit.

tvsonar:
@p3pijn If you want to check the differences between the actual and expected integration test results, you should move to the `actual` folder, and do a `git diff --cached`.

p3pijn:
Ok, @tvsonar I think I have processed all feedback.
Reporting on the exact issue location feels indeed a lot better.
What would be the best place to host the `NestingDepthWalker` class according to your standards?

tvsonar:
We don't have any such standards. I think a nested class would be perfect for this.

tvsonar:
Thanks @p3pijn for the changes. They seem good, I squashed your commits, applied some minor modifications and merged this PR into master. Here's the link to the exact [commit](https://github.com/SonarSource-VisualStudio/sonaranalyzer-csharp-vbnet/commit/1ddad96767a1d4347b34d1988f0f3b905488390d).

PR #808, opened 2016-02-17
SLVS-802 S3453 - Classes should not have only "private" constructors

tvsonar:


src/SonarLint.CSharp/Rules/SealedClassWithOnlyPrivateConstructor.cs
```diff
@@ -0,0 +1,139 @@
...
+    [Tags(Tag.Design)]
+    public class SealedClassWithOnlyPrivateConstructor : DiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S3453";
+        internal const string Title = "\"sealed\" classes should not have only \"private\" constructors";
+        internal const string Description =
+            "A \"sealed\" class can't be extended, and if it has only \"private\" constructors, it can't be instantiated " +
+            "either. In short, it seems to be pointless code. In such cases, either a constructor should be made \"public\", " +
+            "or the \"sealed\" modifier should be removed.";
+        internal const string MessageFormat = "Un-seal this class by making {0} \"public\" or by adding a \"public\" constructor.";
```
dbolkensteyn: adding a public constructor won't "un-seal" the class: only removing the `sealed` modifier will. the choice of words here should be improved.

dbolkensteyn:
Let's discuss this rule tomorrow

PR #810, opened 2016-02-17
SLVS-811 SymbolHelper.TryGetOverriddenOrInterfaceMember can return tr…

tvsonar:
…ue and provide a null overridden symbol


dbolkensteyn:
LGTM

PR #812, opened 2016-02-19
SLVS-812 S1698 should not report on types that override Equals but de…

tvsonar:
…fines it as ReferenceEquals


tvsonar:
@dbolkensteyn I couldn't come up with an improved issue message. The issue message should be improved to handle the case when it's okay to use `==`. We don't know the functionality defined in the `Equals`, but if it's only doing a reference equality check, then it's perfectly fine to use `==`. 

My problem is that it's too long to reason about the raised issue. The issue message could be something like this: Consider using "Equals" if value comparison was intended. "TypeA" overrides "object.Equals", so it's implementation might not simply be a reference equality check.

But at the same time, we have the interface option as well:
Consider using "Equals" if value comparison was intended. "ITypeA" is implemented by "TypeA" that overrides "object.Equals", so it's implementation might not simply be a reference equality check.

This is way too long for an issue message. Do you have any idea on this?

dbolkensteyn:
LGTM

tvsonar:
@dbolkensteyn I guess this also means that you didn't have a better idea for the issue message.

PR #813, opened 2016-02-22
SLVS-683 Unifying type checking

tvsonar:


src/SonarLint/Helpers/KnownType.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+                System_Single,
+                System_Double
+            });
+        public static readonly ISet<KnownType> NonIntegralNumbers = new HashSet<KnownType>(new[]
+        {
+                System_Single,
+                System_Double,
+                System_Decimal
+            });
```
dbolkensteyn: strange indentation

src/SonarLint/Helpers/TypeHelper.cs
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        public static bool IsClass(this ITypeSymbol self)
+        {
+            return self != null && self.TypeKind == TypeKind.Class;
+        }
+
+        public static bool IsClassOrStruct(this ITypeSymbol self)
+        {
+            return self.Is(TypeKind.Struct) || self.IsClass();
```
dbolkensteyn: @tvsonar this method doesn't seem to handle the `null` case, unlike the 2 methods above

src/SonarLint/Helpers/KnownType.cs
```diff
@@ -0,0 +1,147 @@
...
+        public static readonly KnownType System_Char = new KnownType(SpecialType.System_Char, "char");
+        public static readonly KnownType System_Byte = new KnownType(SpecialType.System_Byte, "byte");
+        public static readonly KnownType System_SByte = new KnownType(SpecialType.System_SByte, "sbyte");
+
+        public static readonly ISet<KnownType> FloatingPointNumbers = new HashSet<KnownType>(new[]
+        {
+                System_Single,
+                System_Double
+            });
+        public static readonly ISet<KnownType> NonIntegralNumbers = new HashSet<KnownType>(new[]
```
dbolkensteyn: why do you need both `NonIntegralNumbers` and `FloatingPointNumbers ` ?

the only reference I see to `FloatingPointNumbers` is from `EqualityOnFloatingPoint`, and that rule seem to also be applicable to `Single` types which is also an IEEE floating point type

dbolkensteyn:
LGTM

dbolkensteyn:
Note: As discussed the ticket SLVS-683 should probably be deleted: this is a non-fonctional change

src/SonarLint/Helpers/KnownType.cs
dbolkensteyn: if that change makes sense, it should probably be part of https://github.com/SonarSource-VisualStudio/sonaranalyzer-csharp-vbnet/pull/817 and not of this PR

PR #815, opened 2016-02-23
SLVS-681 S2333 - "unsafe" context is redundant

tvsonar:
Note: there's also https://jira.sonarsource.com/browse/SLVS-774 to enable this rule by default.

### DIFF
--- a/src/SonarLint.CSharp/Rules/RedundantModifier.cs
+++ b/src/SonarLint.CSharp/Rules/RedundantModifier.cs
@@ -18,6 +18,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
  */
 
+using System;
 using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
@@ -27,6 +28,7 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using SonarLint.Common;
 using SonarLint.Common.Sqale;
 using SonarLint.Helpers;
+using System.Collections.Generic;
 
 namespace SonarLint.Rules.CSharp
 {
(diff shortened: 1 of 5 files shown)


src/SonarLint.CSharp/Rules/RedundantModifier.cs
```diff
@@ -109,6 +109,137 @@ public override void Initialize(AnalysisContext context)
...
+        private static readonly ISet<SyntaxKind> UnsafeConstructKinds = new HashSet<SyntaxKind>(new[]
+        {
+            SyntaxKind.AddressOfExpression,
+            SyntaxKind.PointerIndirectionExpression,
+            SyntaxKind.SizeOfExpression,
+            SyntaxKind.PointerType,
+            SyntaxKind.FixedStatement
+        });
+
+        private static void MarkAllUnsafeBlockInside(SyntaxNode container, SyntaxNodeAnalysisContext context)
```
dbolkensteyn: @tvsonar you have a false negative on doubly nested types:
```
    public unsafe class Foo
    {
        public unsafe class Bar
        {
            private int* p;

            public unsafe class Baz // False negative
            {
                private int* p2;
            }
        }
    }
```

my feeling is that this method can be simplified to report on all descendant `unsafe` keywords

src/SonarLint.CSharp/Rules/RedundantModifier.cs
```diff
@@ -109,6 +109,137 @@ public override void Initialize(AnalysisContext context)
...
+                c => CheckForUnnecessaryUnsafeBlocks(c),
+                SyntaxKind.ClassDeclaration,
+                SyntaxKind.StructDeclaration,
+                SyntaxKind.InterfaceDeclaration);
+        }
+
+        private static void CheckForUnnecessaryUnsafeBlocks(SyntaxNodeAnalysisContext c)
+        {
+            var typeDeclaration = (TypeDeclarationSyntax)c.Node;
+            if (typeDeclaration.Parent is TypeDeclarationSyntax)
```
dbolkensteyn: only the 1st and 2nd level of types are being checked, there is a false negative on this:
 ```
    public class Foo
    {
        public class Bar
        {
            public unsafe class Baz // False negative
            {
            }
        }
    }
```

dbolkensteyn:
LGTM

PR #817, opened 2016-02-23
SLVS-803 Rework S1244 (Floating point numbers should not be tested fo…

tvsonar:
…r equality)


dbolkensteyn:
@tvsonar LGTM but can you update SLVS-803's title? "Rework ..." type of tickets aren't supposed to make it in the release notes, they should be updated once we know exactly what we'd like to change (which in this case only seems to be the issue message)

dbolkensteyn:
also this PR seems to related to another comment I've made on the `TypeHelper` introduction: Does this rule also make sense on the `Single` data type? If so, the `TypeHelper` PR should probably be merged as is, and the functional change to the rule part of this one

PR #818, opened 2016-02-24
SLVS-808 Rework S2737 (throw instead of rethrow)

tvsonar:


dbolkensteyn:
@tvsonar can you update the ticket SLVS-808 to document what's changing?

src/SonarLint.CSharp/Rules/CatchRethrow.cs
```diff
@@ -67,18 +69,75 @@ public override void Initialize(AnalysisContext context)
...
+                            redundantCatches.Add(currentCatch);
+                            continue;
+                        }
+
+                        if (currentCatch.Filter != null)
+                        {
+                            continue;
+                        }
+
+                        ComputeExceptionTypesIfNeeded(catches, c.SemanticModel, ref caughtExceptionTypes);
```
dbolkensteyn: @tvsonar this isn't very elegant. Wouldn't using a `Lazy` be more appropriate here?

dbolkensteyn:
LGTM

PR #819, opened 2016-02-26
SLVS-809 S1944 - Inappropriate casts should not be made

tvsonar:


src/SonarLint/Helpers/SymbolHelper.cs
```diff
@@ -22,11 +22,45 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Linq.Expressions;
 
 namespace SonarLint.Helpers
 {
     public static class SymbolHelper
     {
+        public static IEnumerable<INamedTypeSymbol> GetAllNamedTypes(this INamespaceSymbol @namespace)
```
dbolkensteyn: these 2 new methods should be unit tested

src/SonarLint.CSharp/Rules/InvalidCastToInterface.cs
```diff
@@ -0,0 +1,173 @@
...
+                        c =>
+                        {
+                            var cast = (BinaryExpressionSyntax)c.Node;
+                            var interfaceType = c.SemanticModel.GetTypeInfo(cast.Right).Type as INamedTypeSymbol;
+                            var expressionType = c.SemanticModel.GetTypeInfo(cast.Left).Type as INamedTypeSymbol;
+
+                            CheckTypesForInvalidCast(interfaceType, expressionType, interfaceImplementerMappings,
+                                cast.Right.GetLocation(), c);
+                        },
+                        SyntaxKind.AsExpression);
```
dbolkensteyn: @tvsonar it should be straightforward to also apply this rule to `is` no?

dbolkensteyn: `as`, just like `is`, won't throw, unlike the cast

dbolkensteyn:
LGTM

PR #824, opened 2016-03-02
SL integration

tvsonar:
This is a work on top of https://github.com/SonarSource-VisualStudio/sonaranalyzer-csharp/pull/823 to inject the collision detection from the extension.

### DIFF
--- a/src/SonarLint.CSharp/Helpers/DiagnosticAnalyzerContextHelper.cs
+++ b/src/SonarLint.CSharp/Helpers/DiagnosticAnalyzerContextHelper.cs
@@ -29,7 +29,7 @@ namespace SonarLint.Helpers
         #region Register*ActionInNonGenerated
 
         public static void RegisterSyntaxNodeActionInNonGenerated<TLanguageKindEnum>(
-            this AnalysisContext context,
+            this SonarAnalysisContext context,
             Action<SyntaxNodeAnalysisContext> action,
             params TLanguageKindEnum[] syntaxKinds) where TLanguageKindEnum : struct
         {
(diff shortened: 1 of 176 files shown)


src/SonarLint.CSharp/Helpers/DiagnosticAnalyzerContextHelper.cs
```diff
@@ -29,7 +29,7 @@ public static class DiagnosticAnalyzerContextHelper
         #region Register*ActionInNonGenerated
 
         public static void RegisterSyntaxNodeActionInNonGenerated<TLanguageKindEnum>(
-            this AnalysisContext context,
+            this SonarAnalysisContext context,
```
dbolkensteyn: `SonarAnalyzerContext` would seem like a better name, no?

src/SonarLint/Helpers/SonarDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,36 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using Microsoft.CodeAnalysis.Diagnostics;
+using System;
+using System.Diagnostics;
+
+namespace SonarLint.Helpers
+{
+    public abstract class SonarDiagnosticAnalyzer : DiagnosticAnalyzer
```
dbolkensteyn: Why not just  `SonarAnalyzer` or `BaseSonarAnalyzer`? It'd be nice IMO to bring this closer to the product name

dbolkensteyn:
The code LGTM - I still have to make a manual test with SonarLint

dbolkensteyn:
LGTM I'll do the end-to-end test tomorrow when SonarLint will be updated

PR #826, opened 2016-03-02
AppVeyor: No longer publish VSIX snapshots on vsixgallery.com/

dbolkensteyn:


dbolkensteyn:
the diff isn't minimal

PR #828, opened 2016-03-05
SLVS-825 Improve S2757 to handle "=!"

Godin:
https://jira.sonarsource.com/browse/SLVS-825


Godin:
Note that nor title, nor description was not updated, while probably should.

PR #840, opened 2016-03-21
SLVS-766 Rework S101 (class naming) to be able to enable it by default

tvsonar:


tvsonar:
Classes named `_____` are not handled properly, the issue message is empty. we should not allow type names starting and ending with `_`.

tvsonar:
Note to self: the description has been changed.

PR #843, opened 2016-03-23
SLVS-831 Exclude XML comments from S125

tvsonar:


tvsonar:
Related to https://github.com/dotnet/roslyn/issues/10003

PR #852, opened 2016-04-07
SLVS-842 S2328 should not fail on invalid code

tvsonar:


src/Tests/SonarLint.UnitTest/TestCases/GetHashCodeMutable.Fixed.cs
```diff
@@ -38,4 +38,13 @@ public int SomeMethod(int value)
             return hash;
         }
     }
+
+    class
```
dbolkensteyn: @tvsonar shouldn't such an example be added to the invalid code collection? so that all rules are tested against it?

dbolkensteyn:
LGTM

src/Tests/SonarLint.UnitTest/TestCases/GetHashCodeMutable.Fixed.cs
tvsonar: Good point. Added it.

PR #853, opened 2016-04-11
SLVS-855 S3597 - "ServiceContract" and "OperationContract" attributes…

tvsonar:
… should be used together

PR #854, opened 2016-04-11
SLVS-854 S3598 - One-way "OperationContract" methods should have "void" return type

tvsonar:
This PR is based on top of #853

### DIFF
--- a/src/SonarLint.CSharp/Rules/WcfNonVoidOneWay.cs
+++ b/src/SonarLint.CSharp/Rules/WcfNonVoidOneWay.cs
@@ -1,0 +1,97 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015-2016 SonarSource SA
+ * mailto:contact@sonarsource.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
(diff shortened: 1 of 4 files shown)


src/SonarLint.CSharp/Rules/WcfNonVoidOneWay.cs
```diff
@@ -0,0 +1,97 @@
...
+            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category,
+                RuleSeverity.ToDiagnosticSeverity(), IsActivatedByDefault,
+                helpLinkUri: DiagnosticId.GetHelpLink(),
+                description: Description);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Rule); } }
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
```
dbolkensteyn: why not the code block action I think which will provide both the symbol and syntax in its context?

dbolkensteyn:
LGTM

src/SonarLint.CSharp/Rules/WcfNonVoidOneWay.cs
tvsonar: `CodeBlockAction` is executed for properties and field declarations with initializers too, which we don't need here.

PR #855, opened 2016-04-12
SLVS-866 Fixing regression testing

tvsonar:


its/NormalizeSarifReports.ps1
```diff
@@ -1,24 +1,15 @@
 ﻿$ErrorActionPreference = "Stop"
 
-function ConvertProperties
+function GetIssue
```
tvsonar: Probably a better name is `GetMinimalIssueData`

its/NormalizeSarifReports.ps1
```diff
@@ -1,24 +1,15 @@
 ﻿$ErrorActionPreference = "Stop"
```
dbolkensteyn: i'd expect this commit to only contain the changes to the expected files to keep the PR reviewable

dbolkensteyn:
in fact changing the expected file format is going to break sonarlint-vs-resharper

this means we should either update that one as well, or target the old SARIF format. WDYT @tvsonar ?

its/NormalizeSarifReports.ps1
```diff
@@ -1,24 +1,15 @@
...
-  $properties = New-Object –Type System.Object
-  $properties | Add-Member –Type NoteProperty –Name severity –Value $issue.properties.severity
-  $properties | Add-Member –Type NoteProperty –Name warningLevel –Value $issue.properties.warningLevel
-  $properties | Add-Member –Type NoteProperty –Name defaultSeverity –Value $issue.properties.defaultSeverity
-  $properties | Add-Member –Type NoteProperty –Name title –Value $issue.properties.title
-  $properties | Add-Member –Type NoteProperty –Name category –Value $issue.properties.category
-  $properties | Add-Member –Type NoteProperty –Name helpLink –Value $issue.properties.helpLink
-  $properties | Add-Member –Type NoteProperty –Name isEnabledByDefault –Value $issue.properties.isEnabledByDefault
-  $properties | Add-Member –Type NoteProperty –Name isSuppressedInSource –Value @{$true="True";$false="False"}[$issue.isSuppressedInSource] 
+  $issue = New-Object –Type System.Object
```
dbolkensteyn: I'd expect these changes to be in a separate commit from the one updating all the expected files, to keep things reviewable

tvsonar:
We won't be able to fully get back to the old SARIF format, because the previous version of the SARIF contained paths like `Nancy\\src\\Nancy.Testing.Tests\\..\\Nancy.Tests\\ShouldExtensions.cs`, whereas the new SARIF contains `Nancy\\src\\Nancy.Tests\\ShouldExtensions.cs`, so from the new one, it's not possible to create the old path. This would probably not be a problem for the sonarlint-vs-resharper, but it shows that we need to change the expected files at least a bit.

So I'm in favour of changing sonarlint-vs-resharper too. The new format has everything what that needs. I'll give that a go tomorrow.

dbolkensteyn:
In fact now that I am thinking about it... `sonarlint-vs-resharper` is relying on the "raw" reports (i.e. not the normailized ones) in the `output` folder.

So changes to the expected format won't impact it.

However it isn't compatible with Update 2 SARIF's format and will also indeed need at some point to be updated.

dbolkensteyn:
I'd prefer to see the `+1` on `startLine` already done in expected files (as in the new SARIF) as it seems more intuitive IMO

dbolkensteyn:
The indenting of `NormalizeSarifReports.ps1` contains some tabs making it look inconsist on text editors where tabs are set to 2 spaces

dbolkensteyn:
LGTM

PR #856, opened 2016-04-12
SLVS-858 S3600 - "params" should not be introduced on overrides

tvsonar:


appveyor.yml
```diff
@@ -1,5 +1,5 @@
 version: 1.0.{build}
-os: Visual Studio 2015
+os: Previous Visual Studio 2015
```
tvsonar: this should be removed

src/SonarLint.CSharp/Rules/MethodOverrideAddsParams.cs
```diff
@@ -0,0 +1,98 @@
...
+            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category,
+                RuleSeverity.ToDiagnosticSeverity(), IsActivatedByDefault,
+                helpLinkUri: DiagnosticId.GetHelpLink(),
+                description: Description);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Rule); } }
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
```
dbolkensteyn: why not code block action which provides both the syntax and symbol from its context?

dbolkensteyn:
LGTM

src/SonarLint.CSharp/Rules/MethodOverrideAddsParams.cs
tvsonar: same reason as on the other PR, we don't want to get called back on properties.

PR #857, opened 2016-04-12
SLVS-659 S3603 - Methods with "Pure" attribute should return a value

tvsonar:


src/SonarLint.CSharp/Rules/PureAttributeOnVoidMethod.cs
```diff
@@ -0,0 +1,83 @@
...
+    public class PureAttributeOnVoidMethod : SonarDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S3603";
+        internal const string Title = "Methods with \"Pure\" attribute should return a value";
+        internal const string Description =
+            "Marking a method with the \"[Pure]\" attribute specifies that the method doesn't " +
+            "make any visible changes; thus, the method should return a result, otherwise the " +
+            "call to the method should be equal to no-operation. So \"[Pure]\" on a \"void\" method " +
+            "is either a mistake, or the method doesn't do any meaningful task.";
+        internal const string MessageFormat = "Remove the \"Pure\" attribute or change the method to return a value.";
```
dbolkensteyn: @tvsonar I think it's fine to return `void` if you have `out` or `ref` parameters - those are different ways to return a value

dbolkensteyn:
LGTM

PR #861, opened 2016-04-15
SLVS-864 S1450 - Private fields only used as local variables in methods should become local variables

dbolkensteyn:


dbolkensteyn:
I've manually fixed all issues occurring in Akka.Net - and they all were real ones => the project still built successfully after all issues were refactored

src/SonarLint.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -0,0 +1,383 @@
...
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c =>
+                {
+                    var classNode = c.Node as TypeDeclarationSyntax;
+                    if (classNode != null && classNode.Modifiers.Any(m => m.IsKind(SyntaxKind.PartialKeyword)))
+                    {
+                        // Not supported
+                        return;
```
tvsonar: Is this not supported for performance reasons?

src/SonarLint.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -0,0 +1,383 @@
...
+                    }
+
+                    var privateFields = GetPrivateFields(classNode, c.SemanticModel);
+                    var referencesByEnclosingSymbol = GetReferencesByEnclosingSymbol(classNode, privateFields, c.SemanticModel);
+
+                    var classSymbol = c.SemanticModel.GetDeclaredSymbol(classNode);
+                    var classMethods = classSymbol.GetMembers().Where(m => m.Kind == SymbolKind.Method).ToImmutableHashSet();
```
tvsonar: `classSymbol` might be null on invalid input

src/SonarLint.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -0,0 +1,383 @@
...
+                },
+                SyntaxKind.ClassDeclaration, SyntaxKind.StructDeclaration);
+        }
+
+        private static IImmutableDictionary<ISymbol, PrivateField> GetPrivateFields(TypeDeclarationSyntax classNode, SemanticModel semanticModel)
+        {
+            return classNode.Members
+                        .Where(m => m.IsKind(SyntaxKind.FieldDeclaration))
+                        .Cast<FieldDeclarationSyntax>()
+                        .SelectMany(f => f.Declaration.Variables.Select(v => new PrivateField(v, semanticModel.GetDeclaredSymbol(v))))
```
tvsonar: we tend to check the return value of `GetDeclaredSymbol` for being null.

src/SonarLint.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -0,0 +1,383 @@
...
+            SemanticModel semanticModel)
+        {
+            var privateFieldNames = privateFields.Keys.Select(s => s.Name).ToImmutableHashSet();
+
+            var potentialReferences = node.DescendantNodes()
+                .Where(n => n.IsKind(SyntaxKind.IdentifierName))
+                .Cast<IdentifierNameSyntax>()
+                .Where(id => privateFieldNames.Contains(id.Identifier.ValueText));
+
+            var builder = new Dictionary<ISymbol, IDictionary<SyntaxNode, ISymbol>>();
```
tvsonar: this could be renamed to `referencesByEnclosingSymbol`

src/SonarLint.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -0,0 +1,383 @@
...
+        }
+
+        private static bool TryGetMemberBody(ISymbol memberSymbol, out BlockSyntax body)
+        {
+            body = null;
+
+            var syntax = memberSymbol.DeclaringSyntaxReferences.SingleOrDefault()?.GetSyntax();
+            if (syntax != null)
+            {
+                var methodSyntax = syntax as MethodDeclarationSyntax;
```
tvsonar: with `BaseMethodDeclarationSyntax` here, we could remove the `ConstructorDeclarationSyntax` below, and would not have false negatives with destructors

src/SonarLint.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -0,0 +1,383 @@
...
+            IImmutableDictionary<ISymbol, PrivateField> privateFields,
+            BlockSyntax body,
+            IImmutableDictionary<SyntaxNode, ISymbol> references,
+            out ISet<SyntaxNode> rewrittenNodes)
+        {
+            var symbolsToRewrite = references.Values.ToImmutableHashSet();
+
+            var localDeclaration = SyntaxFactory.LocalDeclarationStatement(
+                SyntaxFactory.VariableDeclaration(
+                    SyntaxFactory.PredefinedType(SyntaxFactory.Token(SyntaxKind.IntKeyword)).WithTrailingTrivia(SyntaxFactory.Space),
```
tvsonar: `int` is hardcoded here

src/SonarLint.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -0,0 +1,383 @@
...
+
+            var newSyntaxRoot = classNode.SyntaxTree.GetRoot().ReplaceSyntax(
+                replacements.Keys,
+                (original, partiallyRewritten) => replacements[original],
+                null,
+                null,
+                null,
+                null);
+            var newSyntaxTree = classNode.SyntaxTree.WithRootAndOptions(newSyntaxRoot, classNode.SyntaxTree.Options);
+            var newCompilation = CSharpCompilation.Create(nameof(PrivateFieldUsedAsLocalVariable), new[] { newSyntaxTree });
```
tvsonar: Is this faster then updating the current compilation with `ReplaceSyntaxTree`?

src/SonarLint.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
```diff
@@ -0,0 +1,383 @@
...
+
+            var diagnostics = newCompilation.GetDiagnostics();
+            foreach (var privateField in privateFields.Values.Where(f => !f.Excluded))
+            {
+                if (diagnostics.Any(d => d.Id == "CS0165" && d.GetMessage().Contains(privateField.UniqueName)))
```
tvsonar: Let's put this ID in `SonarLint.Common.WellKnownDiagnosticIds`

tvsonar:
LGTM, apart from the above comments.

src/SonarLint.Extra/Rules.Description/S1450.html
```diff
@@ -0,0 +1,29 @@
+<p>
+    When the value of a private field is always assigned to in a class' methods before being read, then it is not being used to store class information. Therefore, it should become a local variable in the relevant methods to prevent any misunderstanding.
+</p>
+
+<h2>Noncompliant Code Example</h2>
+<pre>
+{code}
```
tvsonar: {code} marking

tvsonar:
We should remove fields that have attributes on them. A struct can be marked `StructLayout(LayoutKind.Explicit)]` and its field `[FieldOffset(8)]`. (This is from Roslyn)

Also, a constructor that only sets a field, and then that field is never used should not be reported or better should be reported with a different message. (`ReadOnlyUnmanagedMemoryStream._memoryOwner` in Roslyn)

tvsonar:
LGTM, except for the above comments

src/SonarLint.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
dbolkensteyn: much faster, in the current compilation there are tons of other trees to be compiled as well.

tvsonar: I see. And then, this is the reason for not checking partial classes, isn't it?

src/SonarLint.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
dbolkensteyn: it's not really for performance reasons - I'm afraid this will add significant complexity to the code, and I'm not sure this is worth it, at least not for the time being because:
1. we're not 100% sure this rule will work in production - given how it's written, there is a real risk we'll have to drop it
2. we might replace all the implementation with one based on a CFG

src/SonarLint.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
dbolkensteyn: on which input? it would be really valuable to have reproducers that we can easily find back

src/SonarLint.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
dbolkensteyn: again, we need to know the input leading to this behavior so that we can update the unit test

src/SonarLint.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
tvsonar: I didn't find any input that leads to a null declared symbol, but [this method](http://source.roslyn.codeplex.com/#Microsoft.CodeAnalysis.CSharp/Compilation/SyntaxTreeSemanticModel.cs,1529) is producing the returned value, and it can return null.

src/SonarLint.CSharp/Rules/PrivateFieldUsedAsLocalVariable.cs
dbolkensteyn: indeed, this part will become more complex

PR #862, opened 2016-04-15
SLVS-853 S3459 - Unassigned members should be removed

tvsonar:


src/SonarLint.CSharp/Helpers/SyntaxHelper.cs
```diff
@@ -42,5 +43,17 @@ public static ExpressionSyntax RemoveParentheses(this ExpressionSyntax expressio
             }
             return currentExpression;
         }
+
+        public static SyntaxNode GetSelfOrTopParenthesizedExpression(this SyntaxNode node)
```
dbolkensteyn: why not take an `ExpressionSyntax` here, which IMO can be the only type which will have `ParenthesizedExpressionSyntax` as a parent? This way you get rid of the cast below

src/SonarLint.CSharp/Rules/AssignmentInsideSubExpression.cs
```diff
@@ -64,7 +64,7 @@ protected override void Initialize(SonarAnalysisContext context)
                 {
                     var assignment = (AssignmentExpressionSyntax)c.Node;
 
-                    var topParenthesizedExpression = GetTopParenthesizedExpression(assignment);
+                    var topParenthesizedExpression = (ExpressionSyntax)assignment.GetSelfOrTopParenthesizedExpression();
```
dbolkensteyn: this cast could be removed by changing `GetSelfOrTopParenthesizedExpression` to work on `ExpressionSyntax`

src/SonarLint.CSharp/Rules/AssignmentInsideSubExpression.cs
```diff
@@ -155,18 +155,6 @@ private static bool TryGetCoalesceExpression(ExpressionSyntax expression, out Bi
             return coalesceExpression != null && coalesceExpression.IsKind(SyntaxKind.CoalesceExpression);
         }
 
-        private static ExpressionSyntax GetTopParenthesizedExpression(ExpressionSyntax expression)
```
dbolkensteyn: ah in fact the previous version was already working on expression syntaxes

dbolkensteyn:
LGTM

src/SonarLint.CSharp/Helpers/SyntaxHelper.cs
tvsonar: good point.

PR #864, opened 2016-04-19
Added Immutable List types for https://github.com/louthy/language-ext…

dhalsim:
… and fixed for Generic types from Immutable Collections and language-ext by looking at ConstructedFrom (Lst<T> instead of Lst<int>)

### DIFF
--- a/src/SonarLint.CSharp/Rules/ReturnValueIgnored.cs
+++ b/src/SonarLint.CSharp/Rules/ReturnValueIgnored.cs
@@ -165,13 +165,7 @@ namespace SonarLint.Rules.CSharp
             KnownType.System_Collections_Immutable_ImmutableSortedSet,
             KnownType.System_Collections_Immutable_ImmutableSortedSet_T,
             KnownType.System_Collections_Immutable_ImmutableStack,
-            KnownType.System_Collections_Immutable_ImmutableStack_T,
-
-            KnownType.LanguageExt_Lst_T,
-            KnownType.LanguageExt_Map_TKey_TValue,
-            KnownType.LanguageExt_Set_T,
-            KnownType.LanguageExt_Que_T,
-            KnownType.LanguageExt_Stck_T
+            KnownType.System_Collections_Immutable_ImmutableStack_T
         });
 
         private static bool IsLinqMethod(IMethodSymbol methodSymbol)
--- a/src/SonarLint.CSharp/Rules/ReturnValueIgnored.cs
+++ b/src/SonarLint.CSharp/Rules/ReturnValueIgnored.cs
@@ -101,8 +101,11 @@ namespace SonarLint.Rules.CSharp
                 return;
             }
 
+            var constructedFrom = invokedMethodSymbol.ContainingType.ConstructedFrom;
+
             if (IsLinqMethod(invokedMethodSymbol) ||
                 HasOnlySideEffectFreeMethods(invokedMethodSymbol.ContainingType) ||
+                HasOnlySideEffectFreeMethods(constructedFrom) ||
                 IsPureMethod(invokedMethodSymbol))
             {
                 c.ReportDiagnostic(Diagnostic.Create(Rule, expression.GetLocation(), invokedMethodSymbol.Name));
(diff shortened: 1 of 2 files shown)


tvsonar:
@dhalsim Thanks for this contribution. Sorry, but I'm not going to merge this PR because we try not to add dependencies to third party libraries. We have no bandwidth to regularly check if a library still has the same functionality as before. For example, in this case this would mean checking if upcoming versions of `LanguageExt.Lst<T>` are still immutable or not...

Also, not reporting on calls on `Lst<T>` would mean false negatives from the rule, and not false positives, which is acceptable on our side. 

At the same time, you bugfix on `TryGetValue` is greatly appreciated. Could you send a separate PR with that?

dhalsim:
@tvsonar I don't think it would be changed from immutable to mutable though it is declared as immutable in the docs but it is your call so I'll remove `Lst<T>` references when I got the time. 

Maybe using `System.Diagnostics.Contracts.PureAttribute` in the LanguageExt types would be the middle way and will fix louthy/language-ext#86 also rule will work without referencing the related types hardcoded.

tvsonar:
@dhalsim Just to be sure, my comment was not just on `Lst<T>`, but all the 5 custom types. 

At the same time, adding `[Pure]` to the methods of these types seems to be a good way to go.

dhalsim:
@tvsonar I removed references to the LanguageExt types, in addition to that what do you think about checking `Pure` attribute also for class target?

https://msdn.microsoft.com/en-us/library/system.diagnostics.contracts.pureattribute(v=vs.110).aspx#Anchor_1

tvsonar:
I couldn't make any sense of that. "pure type" is not explained on that page, does that mean that all methods of the given type is pure?

Similarly to the other targets, what does that mean that a parameter is pure?

dhalsim:
Yeah I think Pure type could be used for all methods in that type are pure. So I don't need to mark every method with PureAttribute. Also I found [this document](https://msdn.microsoft.com/en-us/library/dd264808(v=vs.110).aspx#Anchor_3) explaining Pure type:

> Types that are marked with the PureAttribute (the attribute applies to all the type's methods).

I think should belong to another rule but a Pure parameter could prevent this usage:

```
public int MyMethod([Pure]MyType myType)
{
    …
    myType.MyProperty = "0"; // this line might create a warning
    …
}
```

Or does all the parameters should be treated Pure if its belonging method or class is already Pure?
There are so many combinations.

dhalsim:
Additional reference from [Code Contracts documentation](http://research.microsoft.com/en-us/projects/contracts/userdoc.pdf): 

> 3.3 Purity and Inheritance
Purity (Section 5.4) is also inherited. Once a method is pure, all of its overrides/implementations must be pure as well.

tvsonar:
Thanks for the additional links. I created the following JIRA ticket for this: https://jira.sonarsource.com/browse/SLVS-874

Also created a placeholder item for pure parameters: https://jira.sonarsource.com/browse/SLVS-875

tvsonar:
I squashed your commits, applied some minor code formatting changes and manually merged this PR with [this commit](https://github.com/SonarSource-VisualStudio/sonaranalyzer-csharp/commit/5af43aad8319e89da0b98f38d91283964b98ca10).

Thank you for finding and fixing this bug.

dhalsim:
@tvsonar I think we can dispose of `PureAttribute` for the **parameters**. That could be achieved by Immutable types: Types that have only readonly properties with private setters. This is even easier with the new c# 6.0 syntax:

```
public class Person
{
	public Person(string name)
	{
		Name = name;
	}
	
	public string Name { get; }	
}
```

See the full example in action: https://dotnetfiddle.net/diTDl9

I've found some more information about how immutable types should be implemented on [this site](http://joeduffyblog.com/2007/11/11/immutable-types-for-c/) -though it is very old-:

>
  1. Immutable types must have only initonly fields.
  1. **Immutable types must only contain fields of immutable types**
  1. **Immutable types should only subclass other immutable types**
  1. **Mutable types should not subclass immutable types**
  1. Immutable types must not leak the ‘this’ reference during construction

Also there is a debate on rosyln as a proposal about the immutable types: dotnet/roslyn#159
It is not clear but immutablility check on the compiler level seems more possible than implementing it  on the tool level.

PR #868, opened 2016-04-22
SLVS-869 S107 should not report on methods that can't be changed

tvsonar:


src/SonarLint.CSharp/Rules/TooManyParameters.cs
```diff
@@ -80,6 +81,13 @@ protected override void Initialize(ParameterLoadingAnalysisContext context)
                 SyntaxKind.ParameterList);
         }
 
+        private static bool CanBeChanged(SyntaxNode declaration, SemanticModel semanticModel)
+        {
+            var symbol = semanticModel.GetDeclaredSymbol(declaration) as IMethodSymbol;
+            return symbol == null ||
```
dbolkensteyn: @tvsonar it doesn't seem convervative to assume the method can be changed if we don't find its symbol

dbolkensteyn:
LGTM

src/SonarLint.CSharp/Rules/TooManyParameters.cs
tvsonar: good point

PR #872, opened 2016-04-26
SLVS-852 S3440 - Variables should not be checked against the values t…

tvsonar:
…hey're about to be assigned


dbolkensteyn:
LGTM

PR #873, opened 2016-04-26
SLVS-865 S3604 - Member initializer values should not be redundant

tvsonar:


dbolkensteyn:
There are false positives when the field is being assigned to, but only conditionally
```
private int foo = 42;

public Program(bool condition)
{
    if (condition)
    {
        foo = 0;
    }
}
```

tvsonar:
This is expected. One fix would be to only report on cases when the assignments are directly in the constructor. (There can be issues, with lambdas, try-catch blocks, conditionals, ...)

src/SonarLint.CSharp/Rules/MemberInitializerRedundant.cs
```diff
@@ -0,0 +1,443 @@
...
+        private static List<CtorDeclarationTuple> GetConstructorTuples(
+            SyntaxNodeAnalysisContext context, List<IMethodSymbol> constructorSymbols)
+        {
+            return constructorSymbols
+                .Select(ctor => new SyntaxNodeSymbolSemanticModelTuple<ConstructorDeclarationSyntax, IMethodSymbol>
+                {
+                    SyntaxNode = ctor.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax() as ConstructorDeclarationSyntax,
+                    Symbol = ctor
+                })
+                .Where(ctor => ctor.SyntaxNode != null)
```
m-g-sonar: As your collection only contains explicit constructors, by construction, why do you have to check again if the associated syntax node exists?

src/SonarLint.CSharp/Rules/MemberInitializerRedundant.cs
```diff
@@ -0,0 +1,443 @@
...
+                                var parentExpression = GetPossibleMemberAccessParent(identifier);
+
+                                bool isRead;
+                                if (TryGetReadWriteFromMemberAccess(parentExpression, out isRead))
+                                {
+                                    return !isRead;
+                                }
+                            }
+                            break;
+                        case SyntaxKind.SimpleMemberAccessExpression:
```
m-g-sonar: both cases are pretty similar. it's maybe simpler to unify these cases. I also wonder about the interest of having a switch, if you are only interested in 3 different nodes

src/SonarLint.CSharp/Rules/MemberInitializerRedundant.cs
```diff
@@ -0,0 +1,443 @@
...
+            private readonly SemanticModel semanticModel;
+
+            public MemberInitializerRedundancyChecker(IControlFlowGraph cfg, ISymbol memberToCheck, SemanticModel semanticModel)
+                : base(cfg)
+            {
+                this.memberToCheck = memberToCheck;
+                this.semanticModel = semanticModel;
+            }
+
+            protected override bool IsBlockValid(Block block)
```
m-g-sonar: bofies of IsBlockValid and IsBlockInvalid are extremely close too. I would try to unify them in a single method, testing if the field/prop/event is read, or written.

src/Tests/SonarLint.UnitTest/TestCases/MemberInitializerRedundant.Fixed.cs
```diff
@@ -0,0 +1,180 @@
...
+using System.Collections.Generic;
+
+namespace Tests.Diagnostics
+{
+    class Person
+    {
+        static int maxAge = 42;
+        int age; // Noncompliant
+        int Age { get; } // Noncompliant
+        int Age2 { get; } = 0; // we already report on this
```
m-g-sonar: where?

src/SonarLint.CSharp/Rules/MemberInitializerRedundant.cs
tvsonar: Good point. Probably I don't have to, but it's just safer to do so. I'm not 100% confident to remove this check.

src/SonarLint.CSharp/Rules/MemberInitializerRedundant.cs
tvsonar: The `switch` is used to have a similar structure here as in the LVA and dead store analysis.

I merged the two cases.

PR #881, opened 2016-05-04
SLVS-885 S2201 should not report on custom extension method

tvsonar:
Checked the Roslyn project, there are some cases of `ToArray` and `ToList` calls that force the loading of the `IEnumerable`.

Also, there's a `Single` call being reported. Single might have a side-effect, it can throw an exception.

I think we shouldn't filter out these cases.


dbolkensteyn:
LGTM

PR #888, opened 2016-05-09
SLVS-890 [Technical] Improve the performance of S2933 (field should b…

tvsonar:
…e readonly)

### DIFF
--- a/src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
+++ b/src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
@@ -27,9 +27,12 @@ using Microsoft.CodeAnalysis.Diagnostics;
 using SonarLint.Common;
 using SonarLint.Common.Sqale;
 using SonarLint.Helpers;
+using System.Collections.Generic;
 
 namespace SonarLint.Rules.CSharp
 {
+    using FieldTuple = SyntaxNodeSymbolSemanticModelTuple<VariableDeclaratorSyntax, IFieldSymbol>;
+
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     [SqaleConstantRemediation("2min")]
     [SqaleSubCharacteristic(SqaleSubCharacteristic.Understandability)]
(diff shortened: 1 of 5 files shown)


src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
```diff
@@ -58,97 +59,71 @@ public class FieldShouldBeReadonly : SonarDiagnosticAnalyzer
...
 
-                analysisContext.RegisterCompilationEndAction(c =>
-                {
-                    var fields = candidateFields.Except(nonCandidateFields);
-                    fields = fields.Intersect(assignedAsReadonly);
+                    var fields = assignedAsReadonly.Intersect(candidateFields).Except(nonCandidateFields);
+
                     foreach (var field in fields)
                     {
                         var declarationReference = field.DeclaringSyntaxReferences.FirstOrDefault();
```
tvsonar: the syntax could be collected beforehand, we are only reporting in the current file.

src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
```diff
@@ -158,16 +133,91 @@ protected override void Initialize(SonarAnalysisContext context)
...
-            });
+                },
+                SyntaxKind.ClassDeclaration,
+                SyntaxKind.InterfaceDeclaration);
+        }
+
+        private static void CollectFieldsFromPartial(SyntaxNodeSemanticModelTuple<TypeDeclarationSyntax> partialDeclaration,
+            HashSet<IFieldSymbol> assignedAsReadonly, HashSet<IFieldSymbol> nonCandidateFields)
+        {
+            var assignments = partialDeclaration.SyntaxNode.DescendantNodes()
```
tvsonar: split these into separate methods.

src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
```diff
@@ -158,16 +133,91 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                .OfType<AssignmentExpressionSyntax>()
+                .Where(a => assignmentKinds.Contains(a.Kind()));
+
+            foreach (var assignment in assignments)
+            {
+                ProcessExpressionChange(assignment.Left, partialDeclaration.SemanticModel, nonCandidateFields, assignedAsReadonly);
```
tvsonar: should be named `CollectFieldFrom...`

tvsonar:
There's false positive on the following:
```
class X
{
    private int x;
    public X()
    {
        new X().x = 12;
    }
}
```

This is not new, but should be fixed. (We need to check `this.x`, or simply `x`, any other reference can't be made `readonly`)

dbolkensteyn:
LGTM

PR #890, opened 2016-05-10
SLVS-896 [Technical] CodeFixes should show up only once in the SonarLint connected mode

tvsonar:


dbolkensteyn:
@tvsonar I don't expect a commit with a ticket # from SVS on this project

dbolkensteyn:
LGTM

tvsonar:
@dbolkensteyn I think SVS is fine here. This PR only implements functionality that is required for SonarLint. Do you think in this case we should create a purely technical ticket in SLVS and link that to the SVS ticket?

PR #897, opened 2016-06-01
SLVS-897 Add support for generic methods

tvsonar:


src/SonarLint.CSharp/Helpers/SyntaxHelper.cs
```diff
@@ -88,7 +88,7 @@ public static ExpressionSyntax GetSelfOrTopParenthesizedExpression(this Expressi
 
         public static bool IsExpressionOnThis(this ExpressionSyntax expression)
         {
-            if (expression.IsKind(SyntaxKind.IdentifierName))
+            if (expression is NameSyntax)
```
tvsonar: This fix is not relevant to the SLVS-897 rule, as the method is not called any more.

PR #901, opened 2016-06-07
SLVS-908 Support no-branch methods and simple true/false literal assi…

tvsonar:
…gnment

### DIFF
--- a/src/SonarLint.CSharp/Helpers/LiveVariableAnalysis/CSharp/LiveVariableAnalysis.cs
+++ b/src/SonarLint.CSharp/Helpers/LiveVariableAnalysis/CSharp/LiveVariableAnalysis.cs
@@ -19,13 +19,13 @@
  */
 
 using Microsoft.CodeAnalysis;
-using SonarLint.Helpers.Cfg.Common;
+using SonarLint.Helpers.FlowAnalysis.Common;
 using Microsoft.CodeAnalysis.CSharp;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 
-namespace SonarLint.Helpers.Cfg.CSharp
+namespace SonarLint.Helpers.FlowAnalysis.CSharp
 {
     public sealed class LiveVariableAnalysis : Common.LiveVariableAnalysis
     {
--- a/src/SonarLint.CSharp/Helpers/SymbolicExecution/CSharp/ExplodedGraph.cs
+++ b/src/SonarLint.CSharp/Helpers/SymbolicExecution/CSharp/ExplodedGraph.cs
@@ -1,0 +1,286 @@
+﻿/*
+ * SonarLint for Visual Studio
+ * Copyright (C) 2015-2016 SonarSource SA
+ * mailto:contact@sonarsource.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+
+using Microsoft.CodeAnalysis;
+using System.Collections.Generic;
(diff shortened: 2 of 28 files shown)


src/SonarLint.CSharp/Helpers/LiveVariableAnalysis/CSharp/LiveVariableAnalysis.cs
```diff
@@ -25,7 +25,7 @@
 using System.Linq;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 
-namespace SonarLint.Helpers.Cfg.CSharp
+namespace SonarLint.Helpers.Lva.CSharp
```
tvsonar: Namespaces are very fragmented. All of these (CFG, LVA, SE) should be moved to `SonarLint.Helper.FlowAnalysis`

src/SonarLint.CSharp/Helpers/SymbolicExecution/CSharp/ExplodedGraph.cs
```diff
@@ -0,0 +1,313 @@
...
+            public ProgramState ProgramState { get; }
+            public ProgramPoint ProgramPoint { get; }
+
+            public Node(ProgramPoint programPoint, ProgramState programState)
+            {
+                ProgramState = programState;
+                ProgramPoint = programPoint;
+            }
+
+            public override int GetHashCode()
```
tvsonar: All `GetHashCode` and `Equals` overrides are part of a following ticket. There is no program point revisiting here, so no collision is expected.

src/SonarLint.CSharp/Helpers/SymbolicExecution/Common/ProgramPoint.cs
```diff
@@ -0,0 +1,60 @@
...
+using SonarLint.Helpers.Cfg.Common;
+
+namespace SonarLint.Helpers.SymbolicExecution.Common
+{
+    public class ProgramPoint
+    {
+        public Block Block { get; }
+        public int Offset { get; }
+
+        internal int NumberOfTimesVisited { get; set; } = 1;
```
tvsonar: Not relevant to this ticket.

src/SonarLint.CSharp/Helpers/SymbolicExecution/CSharp/ExplodedGraph.cs
```diff
@@ -0,0 +1,286 @@
...
+        private readonly SemanticModel semanticModel;
+        private readonly ISymbol declaration;
+        private readonly IEnumerable<ISymbol> declarationParameters = new List<ISymbol>();
+        private readonly Common.LiveVariableAnalysis lva;
+
+        public event EventHandler ExplorationEnded;
+        public event EventHandler<InstructionProcessedEventArgs> InstructionProcessed;
+        public event EventHandler ExitBlockReached;
+
+        private readonly HashSet<Node> workList = new HashSet<Node>(); // todo: a queue would be better, but we don't want to insert the same node twice
```
benzonico: As discussed : this won't work (corrected in further commits).

src/SonarLint.CSharp/Helpers/SymbolicExecution/CSharp/ExplodedGraph.cs
```diff
@@ -0,0 +1,286 @@
...
+
+        public ExplodedGraph(IControlFlowGraph cfg, ISymbol declaration, SemanticModel semanticModel, Common.LiveVariableAnalysis lva)
+        {
+            this.cfg = cfg;
+            this.semanticModel = semanticModel;
+            this.declaration = declaration;
+            this.lva = lva;
```
benzonico: lva is a premature optimization at this stage. It forces you to deal with an extra problem not required right now.

src/SonarLint.CSharp/Helpers/SymbolicExecution/Common/ProgramState.cs
```diff
@@ -0,0 +1,121 @@
...
+        internal ProgramState AddSymbolicValueConstraint(SymbolicValue symbolicValue, SymbolicValueConstraint constraint)
+        {
+            var newProgramState = new ProgramState(
+                new Dictionary<ISymbol, SymbolicValue>(symbolValueAssociations),
+                new Dictionary<SymbolicValue, SymbolicValueConstraint>(constraints));
+            newProgramState.constraints[symbolicValue] = constraint;
+            return newProgramState;
+        }
+
+        public bool TryGetBoolValue(ISymbol symbol, out bool value)
```
benzonico: This seems a rather convoluted API : if I get things correctly the method returns if there is a constraint or not in the PS. 
As a caller, you don't really care. What you would like to know is : if I want to put a true (or false) constraint on this symbol for this PS, what is the new PS (if it is possible). So that function would return the same PS, a new one or null (or absence of value) if not achievable (setting true on false for instance). That will simplify code on caller site.

PR #902, opened 2016-06-07
SLVS-909 Support binary branching in exploded graph walk

tvsonar:


src/Tests/SonarLint.UnitTest/Helpers/ExplodedGraphTest.cs
```diff
@@ -169,5 +173,231 @@ public void ExplodedGraph_SequentialInput_Max()
...
+                        bool value;
+                        Assert.IsTrue(args.ProgramState.TryGetBoolValue(aSymbol, out value) && !value);
+                        numberOfLastInstructionVisits++;
+                    }
+                };
+
+            explodedGraph.Walk();
+
+            Assert.IsTrue(explorationEnded);
+            Assert.AreEqual(1, numberOfExitBlockReached); // currently the different SymbolicValue instances are causing the creation of different ProgramStates
```
tvsonar: This fails right now.

src/SonarLint.CSharp/Helpers/SymbolicExecution/CSharp/ExplodedGraph.cs
```diff
@@ -158,13 +160,71 @@ private void OnInstructionProcessed(SyntaxNode instruction, ProgramPoint program
...
+                        if (symbol != null && IsLocalScoped(symbol))
+                        {
+                            var symbolicValue = currentState.GetSymbolValue(symbol);
+                            if (symbolicValue == null)
+                            {
+                                throw new InvalidOperationException("Symbol without symbolic value");
+                            }
+
+                            bool boolValue;
+                            if (currentState.TryGetBoolValue(symbol, out boolValue))
```
benzonico: See comment about this method in previous commit : API is not great for this one.

src/SonarLint.CSharp/Helpers/SymbolicExecution/CSharp/ExplodedGraph.cs
```diff
@@ -158,13 +160,71 @@ private void OnInstructionProcessed(SyntaxNode instruction, ProgramPoint program
...
+                                    EnqueueNewNode(new ProgramPoint(binaryBranchBlock.TrueSuccessorBlock, 0), currentState);
+                                }
+                                else
+                                {
+                                    EnqueueNewNode(new ProgramPoint(binaryBranchBlock.FalseSuccessorBlock, 0), currentState);
+                                }
+                            }
+                            else
+                            {
+                                EnqueueNewNode(new ProgramPoint(binaryBranchBlock.TrueSuccessorBlock, 0), currentState.AddSymbolicValueConstraint(symbolicValue, BooleanLiteralConstraint.True));
```
benzonico: You should probably need only those two lines and handle the irrealisable program state either around those lines or in the EnqueueNewNode method.

src/SonarLint.CSharp/Helpers/SymbolicExecution/CSharp/ExplodedGraph.cs
```diff
@@ -158,13 +160,71 @@ private void OnInstructionProcessed(SyntaxNode instruction, ProgramPoint program
...
+                                EnqueueNewNode(new ProgramPoint(binaryBranchBlock.FalseSuccessorBlock, 0), currentState.AddSymbolicValueConstraint(symbolicValue, BooleanLiteralConstraint.False));
+                            }
+                        }
+                        else
+                        {
+                            EnqueueNewNode(new ProgramPoint(binaryBranchBlock.TrueSuccessorBlock, 0), currentState);
+                            EnqueueNewNode(new ProgramPoint(binaryBranchBlock.FalseSuccessorBlock, 0), currentState);
```
benzonico: I guess that you should probably overload the constructor to have a default zero offset instead of repeating it all over ?

src/SonarLint.CSharp/Helpers/SymbolicExecution/CSharp/ExplodedGraph.cs
```diff
@@ -197,21 +257,21 @@ private void VisitInstruction(Node node)
...
                     }
                     break;
                 default:
                     break;
             }
 
-            EnqueueNewNode(newProgramPoint, newProgramState);
-            OnInstructionProcessed(instruction, node.ProgramPoint, newProgramState);
+            EnqueueNewNode(newProgramPoint, currentState);
+            OnInstructionProcessed(instruction, node.ProgramPoint, currentState);
```
benzonico: If this is API for checks, it should probably have been added with commit of the rule (as well as all the event and event handlers related to this)

benzonico: ok this is somehow also used for tests. Nevermind.

src/SonarLint.CSharp/Helpers/SymbolicExecution/CSharp/ExplodedGraph.cs
```diff
@@ -328,7 +327,13 @@ private void EnqueueNewNode(ProgramPoint programPoint, ProgramState programState
             {
                 programPoints[pos] = pos;
             }
-            workList.Add(new Node(pos, programState));
+
+            var newNode = new Node(pos, programState);
+            if (!alreadyAddedNodes.Contains(newNode))
+            {
+                alreadyAddedNodes.Add(newNode);
```
benzonico: No need to use contains, you can rely on result of Add if i'm correct : https://msdn.microsoft.com/en-us/library/bb353005(v=vs.110).aspx

src/SonarLint.CSharp/Helpers/ControlFlowGraph/CSharp/ControlFlowGraphBuilder.cs
```diff
@@ -1049,10 +1049,10 @@ private void BuildVariableDeclaration(VariableDeclarationSyntax declaration)
 
         private void BuildVariableDeclarator(VariableDeclaratorSyntax variableDeclarator)
         {
-            currentBlock.ReversedInstructions.Add(variableDeclarator);
```
benzonico: There is a lack of why such a move here. Nothing is clear either from the commit title or the code about why you did that change. 
Can you make things clear ? I tend to think this is an improvement of the CFG based on the fact that in C# you are required to initialize local var but, please confirm this and most probably open a dedicated ticket for it ?

src/SonarLint.CSharp/Helpers/SymbolicExecution/CSharp/ExplodedGraph.cs
tvsonar: @benzonico you are right, that the set's `Add` doesn't need the `Contains`, but the `worklist` queue does. And then it's better to have the set's `Add` also in the `if`.

PR #903, opened 2016-06-08
SLVS-905 S2583 - Conditions should not unconditionally evaluate to "true" or to "false"

tvsonar:


src/Tests/SonarLint.UnitTest/TestCases/ConditionEvaluatesToConstant.cs
```diff
@@ -0,0 +1,139 @@
...
+        public void Method6(bool cond)
+        {
+            var i = 10;
+            while (i < 20)
+            {
+                i = i + 1;
+            }
+
+            var b = true;
+            while (b) // Non-compliant, the above while changes ProgramState every time we are inside the body
```
tvsonar: This is a false negative

src/Tests/SonarLint.UnitTest/PackagingTests/VsixTest.cs
```diff
@@ -33,10 +33,10 @@ public void Size_Check()
             const string vsixFileName = "SonarAnalyzer.vsix";
 #if DEBUG
             const string pathEnding = @"bin\Debug";
-            const int approxFileSize = 973 * 1024;
+            const int approxFileSize = 1068 * 1024;
 #else
             const string pathEnding = @"bin\Release";
-            const int approxFileSize = 303 * 1024;
+            const int approxFileSize = 335 * 1024;
```
benzonico: would be cool to precise why there is an increase.

src/SonarLint.CSharp/Rules/ConditionEvaluatesToConstant.cs
```diff
@@ -0,0 +1,173 @@
...
+        {
+            IControlFlowGraph cfg;
+            if (!ControlFlowGraph.TryGet(body, context.SemanticModel, out cfg))
+            {
+                return;
+            }
+
+            var lva = LiveVariableAnalysis.Analyze(cfg, declaration, context.SemanticModel);
+
+            var explodedGraph = new ExplodedGraph(cfg, declaration, context.SemanticModel, lva);
```
benzonico: This is ok for implementation of this rule now but you will need to share the walking of the graph with other checks.

src/SonarLint.CSharp/Rules/ConditionEvaluatesToConstant.cs
```diff
@@ -0,0 +1,173 @@
...
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using LiveVariableAnalysis = Helpers.FlowAnalysis.CSharp.LiveVariableAnalysis;
+
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [SqaleConstantRemediation("15min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.LogicReliability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags(Tag.Bug, Tag.Cert, Tag.Cwe, Tag.Misra)]
+    public class ConditionEvaluatesToConstant : SonarDiagnosticAnalyzer
```
benzonico: slightly debatable name : the word "constant" bothers me a bit.

src/SonarLint.CSharp/Rules/ConditionEvaluatesToConstant.cs
```diff
@@ -160,6 +169,12 @@ private void ProcessVisitedBlocks(HashSet<SyntaxNode> conditionTrue, HashSet<Syn
 
         private static void CollectConditions(ConditionEvaluatedEventArgs args, HashSet<SyntaxNode> conditionTrue, HashSet<SyntaxNode> conditionFalse)
         {
+            if (args.Condition.IsKind(SyntaxKind.TrueLiteralExpression) ||
```
benzonico: What does this have to do with the commit ? if this is to exclude while(true) I think this is a bit over aggressive (you still want to raise issue on if(true) no ? )

src/Tests/SonarLint.UnitTest/TestCases/ConditionEvaluatesToConstant.cs
benzonico: Why is that a false negative ?

src/SonarLint.CSharp/Helpers/ControlFlowGraph/CSharp/ControlFlowGraphBuilder.cs
```diff
@@ -1049,10 +1049,12 @@ private void BuildVariableDeclaration(VariableDeclarationSyntax declaration)
 
         private void BuildVariableDeclarator(VariableDeclaratorSyntax variableDeclarator)
         {
+            // There are variable declarations which implicitly get a value, such as foreach (var x in xs)
+            currentBlock.ReversedInstructions.Add(variableDeclarator);
+
```
benzonico: Well,  that explains my other comment. But I have an issue with the fact that your CFG gets modified depending on your SE. That should not be the case as those are related but separate concepts.

benzonico: Note moreover that there seems to be zero impact on CFG tests if you modify things. That means you only have indirect tests (better than nothing) rather than a unit test for this case of variable declarations in CFG.

src/SonarLint.CSharp/Helpers/SymbolicExecution/CSharp/ExplodedGraph.cs
```diff
@@ -204,7 +204,9 @@ private void VisitBinaryBranch(BinaryBranchBlock binaryBranchBlock, Node node)
                         var identifier = (IdentifierNameSyntax)instruction;
                         var symbol = semanticModel.GetSymbolInfo(identifier).Symbol;
 
-                        if (symbol != null && IsLocalScoped(symbol))
+                        if (symbol != null &&
+                            IsLocalScoped(symbol) &&
```
benzonico: Unrelated to this committ but IsLocalScoped should probably accept null symbol and return a result accordingly as it is always used with a null check.

PR #904, opened 2016-06-09
SLVS-912 Support switch branching in exploded graph walk

tvsonar:


src/SonarLint.CSharp/Helpers/SymbolicExecution/CSharp/ExplodedGraph.cs
```diff
@@ -192,6 +193,25 @@ private ProgramState GetCleanedProgramState(Node node)
...
+            if (branchBlock is BinaryBranchBlock)
+            {
+                throw new InvalidOperationException();
+            }
+            
+            var newProgramState = GetCleanedProgramState(node);
+            EnqueueAllSuccessors(branchBlock, newProgramState);
+        }
+
+        private void EnqueueAllSuccessors(BranchBlock branchBlock, ProgramState newProgramState)
```
benzonico: I tend to think that there should be no distinction between this kind of blocks and a SimpleJump block for this usage : you just want to enqueue all successors. As such, they should probably have a common API to get their successors (a simpleJump being the particular case to having only one).

PR #913, opened 2016-06-13
SLVS-910 Support loops in the exploded graph walk

tvsonar:


src/SonarLint.CSharp/Helpers/SymbolicExecution/CSharp/ExplodedGraph.cs
```diff
@@ -463,13 +477,21 @@ private void EnqueueNewNode(ProgramPoint programPoint, ProgramState programState
                 programPoints[pos] = pos;
             }
 
-            var newNode = new Node(pos, programState);
+            if (programState.GetVisitedCount(pos) >= MaxProgramPointExecutionCount)
```
benzonico: the >= seems dubious to me : shouldn't it be a >

src/Tests/SonarLint.UnitTest/TestCases/ConditionEvaluatesToConstant.cs
```diff
@@ -202,5 +202,62 @@ public bool Property1
...
+                        guard2 = false;
+                    }
+                    else
+                    {
+                        guard3 = false;
+                    }
+                }
+            }
+
+            if (guard3) // Noncompliant, false-positive
```
benzonico: Probably not required

tvsonar: Add comment why this is kept

PR #914, opened 2016-06-14
SLVS-903 Report on always assigned method parameters

tvsonar:


tvsonar:
Note to self: this rule should be based on the syntax, and not the symbol. Both this and the dead store rule performs LVA on basically all methods, and the performance of the dead store rule is a lot better. The reason is most probably because of the syntax vs symbol based approach. Refactor this rule to be syntax based too.

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,43 +64,90 @@ public class MethodParameterUnused : SonarDiagnosticAnalyzer
...
+                SyntaxKind.ConstructorDeclaration);
         }
 
-        private static void ReportUnusedParametersOnMethod(IMethodSymbol methodSymbol, SymbolAnalysisContext context)
+        private static void ReportUnusedParametersOnMethod(BaseMethodDeclarationSyntax declaration, IMethodSymbol methodSymbol,
+            SyntaxNodeAnalysisContext context)
         {
-            if (!MethodCanBeSafelyChanged(methodSymbol))
+            if (!MethodCanBeSafelyChanged(methodSymbol) ||
+                IsUsedAsEventHandlerFunctionOrAction(methodSymbol, context.SemanticModel.Compilation))
```
tvsonar: This check is very expensive.

src/SonarLint.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -58,43 +64,104 @@ public class MethodParameterUnused : SonarDiagnosticAnalyzer
...
-                SymbolKind.NamedType);
+                SyntaxKind.MethodDeclaration,
+                SyntaxKind.ConstructorDeclaration);
+        }
+
+        private static void ReportUnusedParametersOnMethod(BaseMethodDeclarationSyntax declaration, IMethodSymbol methodSymbol,
+            SyntaxNodeAnalysisContext context)
+        {
+            if (!MethodCanBeSafelyChanged(methodSymbol) ||
+                IsUsedAsEventHandlerFunctionOrAction(methodSymbol, context.SemanticModel.Compilation))
```
tvsonar: Note to self: This call is expensive, so we should minimize the number of times it's called. So that's the reason why this rule originally was a symbol based rule on `NamedType`s. Consider reverting this refactoring if a type declaration (syntax) based solution is not performing well enough.

PR #917, opened 2016-06-17
SLVS-923 Add support for nullable value types

tvsonar:


src/SonarLint.CSharp/Rules/NullPointerDereference.cs
```diff
@@ -0,0 +1,278 @@
...
+                }
+
+                var symbol = semanticModel.GetSymbolInfo(identifier).Symbol;
+                if (!explodedGraph.IsLocalScoped(symbol))
+                {
+                    return programState;
+                }
+
+                var sv = programState.GetSymbolValue(symbol);
+                if (sv == SymbolicValue.Null)
```
benzonico: I would simplify the call here... matter of style, so no big deal.

src/SonarLint.CSharp/Rules/NullPointerDereference.cs
```diff
@@ -0,0 +1,278 @@
...
+                        return programState;
+                }
+
+                var symbol = semanticModel.GetSymbolInfo(identifier).Symbol;
+                if (!explodedGraph.IsLocalScoped(symbol))
+                {
+                    return programState;
+                }
+
+                var sv = programState.GetSymbolValue(symbol);
```
benzonico: This part can probably be factorized with the code repeated above.

PR #920, opened 2016-06-20
SLVS-926 Dead store rule should not report on unused local variables

tvsonar:


src/SonarLint.Extra/Rules.Description/S1854.html
```diff
@@ -39,7 +39,12 @@
 
 <h2>Exceptions</h2>
 <p>
-    No issue is reported on statements inside <code>try</code> blocks.
+    No issue is reported when
+    <ul>
+        <li>the analyzed method body contains <code>try</code> blocks,</li>
+        <li>a lambda expressions capture local variables, or</li>
+        <li>Rule S1481 already reports on a variable declaration.</li>
```
benzonico: This most probably misses a link to the rule mentioned, and lacks a clear explanation of what this means, I would go for something like : Variable is unused (case covered by rule S1481).

src/SonarLint.CSharp/Rules/DeadStores.cs
```diff
@@ -282,13 +285,31 @@ private void ProcessVariableDeclarator(SyntaxNode instruction, HashSet<ISymbol>
...
                     context.ReportDiagnostic(Diagnostic.Create(Rule, declarator.Initializer.EqualsToken.GetLocation(), symbol.Name));
                 }
                 liveOut.Remove(symbol);
             }
 
+            private bool IsUnusedLocal(ISymbol declaredSymbol)
+            {
+                var identifiers = node.DescendantNodes().OfType<IdentifierNameSyntax>();
+
+                foreach (var identifier in identifiers)
```
benzonico: I tend to think that using LinQ/func programming style this could be a nice one liner.

src/SonarLint.CSharp/Rules/VariableUnused.cs
```diff
@@ -93,5 +93,19 @@ protected override void Initialize(SonarAnalysisContext context)
                 });
             });
         }
+
+        internal static IEnumerable<ISymbol> GetPossiblyUsedSymbols(SyntaxNode node, SemanticModel semanticModel)
```
benzonico: PossiblyUsed is a bit odd... sounds like the result of the method is uncertain.

src/SonarLint.Extra/Rules.Description/S1854.html
tvsonar: The link is automatically added by the doc generator when it sees Rule S...

I'll add the explanation.

PR #921, opened 2016-06-20
SLVS-928 Change issue location of "dead store" rule to not only report on the equal sign

tvsonar:


src/SonarLint.CSharp/Rules/DeadStores.cs
```diff
@@ -368,11 +377,24 @@ private void CollectAllCapturedLocalSymbols(SyntaxNode instruction, HashSet<ISym
                 if (LiveVariableAnalysis.IsLocalScoped(symbol, declaration) &&
                     !outState.Contains(symbol))
                 {
-                    context.ReportDiagnostic(Diagnostic.Create(Rule, assignment.OperatorToken.GetLocation(), symbol.Name));
+                    var line = GetLineOfToken(assignment.OperatorToken, assignment.SyntaxTree);
```
benzonico: I am pretty sure this can be factorized in a method taking a few parameters to avoid duplication with the process variable declarator modification.

src/Tests/SonarLint.UnitTest/TestCases/DeadStores.cs
```diff
@@ -30,7 +30,8 @@ void calculateRate(int a, int b)
             i += i + 2; // Noncompliant
             i = 5;
             j = i;
-            i = doSomething();  // Noncompliant; retrieved value overwritten in for loop
+            i
+                = doSomething(); // Noncompliant; retrieved value overwritten in for loop
```
benzonico: I guess from this that you have no way to test precise issue location. Might be worth to, at least, create a ticket about that ?

tvsonar: Yes, and the message is not checked either. Ticket created: https://jira.sonarsource.com/browse/SLVS-932

PR #925, opened 2016-06-24
SLVS-894 S3459 (unassigned members) should not report on structs whos…

tvsonar:
…e fields are being set

### DIFF
--- a/src/SonarLint.CSharp/Rules/NotAssignedPrivateMember.cs
+++ b/src/SonarLint.CSharp/Rules/NotAssignedPrivateMember.cs
@@ -28,6 +28,7 @@ using SonarLint.Helpers;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using System;
 
 namespace SonarLint.Rules.CSharp
 {
@@ -137,8 +138,7 @@ namespace SonarLint.Rules.CSharp
 
             var identifiers = removableDeclarationCollector.ClassDeclarations
                 .SelectMany(container => container.SyntaxNode.DescendantNodes()
-                    .Where(node =>
-                        node.IsKind(SyntaxKind.IdentifierName))
+                    .Where(node => node.IsKind(SyntaxKind.IdentifierName))
                     .Cast<IdentifierNameSyntax>()
                     .Where(node => symbolNames.Contains(node.Identifier.ValueText))
                     .Select(node =>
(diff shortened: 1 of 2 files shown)


src/Tests/SonarLint.UnitTest/TestCases/NotAssignedPrivateMember.cs
```diff
@@ -24,9 +39,24 @@ class Nested
...
+            private ValueType1 v3; // Noncompliant
+
+            private ValueType2 v4; // Compliant, a member is assigned
+            private ValueType2 v5; // Compliant, a member is assigned
+            private ValueType2 v6; // Noncompliant
+
             public Nested()
             {
                 Property2 = 42;
+                v1.field++;
```
tvsonar: should also test `this.v1.field`

src/SonarLint.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -137,8 +138,7 @@ private static bool IsAutoPropertyWithNoInitializer(PropertyDeclarationSyntax de
 
             var identifiers = removableDeclarationCollector.ClassDeclarations
                 .SelectMany(container => container.SyntaxNode.DescendantNodes()
-                    .Where(node =>
-                        node.IsKind(SyntaxKind.IdentifierName))
+                    .Where(node => node.IsKind(SyntaxKind.IdentifierName))
```
m-g-sonar: useless change.

src/SonarLint.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -177,17 +176,30 @@ private static ISet<ISymbol> GetAssignedMemberSymbols(IList<MemberUsage> memberU
...
                     simpleMemberAccess.Name == node)
                 {
                     node = simpleMemberAccess;
                 }
 
-                // handle "((this.FieldName))"
+                // Handle "((expr.FieldName))"
                 node = node.GetSelfOrTopParenthesizedExpression();
 
+                if (GetMemberType(memberSymbol).IsValueType)
```
m-g-sonar: If you are only using this method to check if it's a ValueType, then, why not simply call it `isValueType` and do the test inside?

src/SonarLint.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -208,20 +220,39 @@ private static ISet<ISymbol> GetAssignedMemberSymbols(IList<MemberUsage> memberU
...
+                return field.Type;
+            }
+
+            var property = symbol as IPropertySymbol;
+            if (property != null)
+            {
+                return property.Type;
+            }
+
+            throw new ArgumentException($"Symbol needs to either be IFieldSymbol or IPropertySymbol, but it's {symbol.GetType()}");
```
m-g-sonar: I would rename the method "isValueType()` and simply return false if the symbol is not a field, nor a property. No need to throw an exception which will certainly interrupt the analysis, even if not possible by construction (it will?). I also guess that you are not having coverage on this line.

PR #926, opened 2016-06-27
SLVS-895 S2933 should not suggest making struct readonly if members a…

tvsonar:
…re assigned

### DIFF
--- a/src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
+++ b/src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
@@ -32,6 +32,7 @@ using System.Collections.Generic;
 namespace SonarLint.Rules.CSharp
 {
     using FieldTuple = SyntaxNodeSymbolSemanticModelTuple<VariableDeclaratorSyntax, IFieldSymbol>;
+    using TypeDeclarationTuple = SyntaxNodeSemanticModelTuple<TypeDeclarationSyntax>;
 
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     [SqaleConstantRemediation("2min")]
@@ -100,9 +101,6 @@ namespace SonarLint.Rules.CSharp
                         return;
                     }
 
-                    var assignedAsReadonly = new HashSet<IFieldSymbol>();
-                    var nonCandidateFields = new HashSet<IFieldSymbol>();
-
                     var partialDeclarations = declaredSymbol.DeclaringSyntaxReferences
                         .Select(reference => reference.GetSyntax())
                         .OfType<TypeDeclarationSyntax>()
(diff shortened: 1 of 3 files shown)


src/Tests/SonarLint.UnitTest/TestCases/FieldShouldBeReadonly.Fixed.cs
```diff
@@ -84,10 +84,63 @@ public X()
...
+    }
+    struct Y1
+    {
+        public string z;
+    }
+
+    class MyClass
+    {
+        private X1Struct x; // Compliant
+        private readonly X1Struct y; // Noncompliant
```
benzonico: I don't understand why this should raise an issue as the check is designed to raise issue to suggest that a field should be read only.

tvsonar: @benzonico This is the `.Fixed.cs` file, which is the output of the code fix.

src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
```diff
@@ -144,141 +134,226 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            {
+                var prefixUnaries = partialTypeDeclaration.SyntaxNode.DescendantNodes()
+                    .OfType<PrefixUnaryExpressionSyntax>()
+                    .Where(a => prefixUnaryKinds.Contains(a.Kind()));
 
-        private static void ProcessArgument(ArgumentSyntax argument, SemanticModel semanticModel,
-            HashSet<IFieldSymbol> assignedAsReadonly, HashSet<IFieldSymbol> nonCandidateFields)
-        {
-            if (argument.RefOrOutKeyword.IsKind(SyntaxKind.None))
+                foreach (var prefixUnary in prefixUnaries)
```
benzonico: Couldn't this be somehow a oneliner by havign the foreach on the Where method ? http://stackoverflow.com/questions/200574/linq-equivalent-of-foreach-for-ienumerablet ??

tvsonar: Yes, it can be, but... `.ForEach()` is defined on `List<T>`, so we'd need to call a `.ToList()` on the result of the `Where`. Just for `.ForEach()`, I usually don't do this.

src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
```diff
@@ -144,141 +134,226 @@ protected override void Initialize(SonarAnalysisContext context)
...
-            CollectFieldsFromPostfixUnaryExpressions(partialDeclaration, assignedAsReadonly, nonCandidateFields);
-            CollectFieldsFromArguments(partialDeclaration, assignedAsReadonly, nonCandidateFields);
+            return fieldDeclaration.Declaration.Variables
+                .Select(variableDeclaratorSyntax => new FieldTuple
+                {
+                    SyntaxNode = variableDeclaratorSyntax,
+                    Symbol = semanticModel.GetDeclaredSymbol(variableDeclaratorSyntax) as IFieldSymbol,
+                    SemanticModel = semanticModel
+                })
+                .Where(f => f.Symbol != null)
```
benzonico: given the method FieldIsRelevant, this one is redundant

src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
```diff
@@ -144,141 +134,226 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                assignedAsReadonly.Intersect(candidateFields).Except(nonCandidateFields);
 
-            foreach (var postfixUnary in postfixUnaries)
+            public ReadonlyFieldCollector(IEnumerable<TypeDeclarationTuple> partialTypeDeclarations,
+                IDictionary<IFieldSymbol, VariableDeclaratorSyntax> fieldDeclarations)
             {
-                ProcessExpression(postfixUnary.Operand, partialDeclaration.SemanticModel, assignedAsReadonly, nonCandidateFields);
-            }
-        }
+                nonCandidateFields = new HashSet<IFieldSymbol>();
```
benzonico: I struggle to see why you need 3 sets.

src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
```diff
@@ -144,141 +134,226 @@ protected override void Initialize(SonarAnalysisContext context)
...
-        {
-            var assignments = partialDeclaration.SyntaxNode.DescendantNodes()
-                .OfType<AssignmentExpressionSyntax>()
-                .Where(a => assignmentKinds.Contains(a.Kind()));
 
-            foreach (var assignment in assignments)
+            private void CollectFieldsFromDeclaration(TypeDeclarationTuple partialTypeDeclaration)
             {
-                ProcessExpression(assignment.Left, partialDeclaration.SemanticModel, assignedAsReadonly, nonCandidateFields);
+                CollectFieldsFromAssignments(partialTypeDeclaration);
```
benzonico: This sounds like you should have an inner class dedicated to handle a unique partialTypeDeclaration. That would allow you to store some overused objects in fields of that inner class making things a bit more readable (ie: avoid passing the semantic model all over the methods).

src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
```diff
@@ -144,141 +134,226 @@ protected override void Initialize(SonarAnalysisContext context)
...
-                return;
+                bool isNestedMemberAccess;
+                var rootExpression = GetTopMemberAccess(expression, out isNestedMemberAccess);
+
+                if (rootExpression == null)
+                {
+                    return;
+                }
+
+                var fieldSymbol = semanticModel.GetSymbolInfo(rootExpression).Symbol as IFieldSymbol;
```
benzonico: This should most probably be part of the GetTopMemberAccess method, that would avoid passing out parameter.

src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
```diff
@@ -112,173 +110,253 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                }
+
+                private static ExpressionSyntax GetTopMemberAccessIfNested(ExpressionSyntax expression)
+                {
+                    bool isNestedMemberAccess;
+                    var top = GetTopMemberAccess(expression, out isNestedMemberAccess);
+                    return isNestedMemberAccess ? top : null;
+                }
+
+                private static ExpressionSyntax GetTopMemberAccess(ExpressionSyntax expression, out bool isNestedMemberAccess)
```
benzonico: I would try to remove the out param here : there's only a need to identify the first level of recursion and return null in this case if noParens if it is a NAmeSyntax at first level and same when memberAccess.Expression  is ThisExpression. 

That will allow you to get rid of the `GetTopMemberAccessIfNested` method.

src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
```diff
@@ -112,173 +110,253 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                public void CollectFields()
+                {
+                    CollectFieldsFromDeclarations();
+                    CollectFieldsFromAssignments();
+                    CollectFieldsFromPrefixUnaryExpressions();
+                    CollectFieldsFromPostfixUnaryExpressions();
+                    CollectFieldsFromArguments();
+                }
+
+                private void CollectFieldsFromDeclarations()
```
benzonico: I would change this to return a list/set and have it written as sometihng like :
```
return  allFields.Where(f =>
                         IsFieldRelevant(f.Symbol) &&
                         f.SyntaxNode.Initializer != null).Select(f -> f.Symbol);
```
Then you could do a UnionWith this result where it is called.

src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
tvsonar: @benzonico I don't get this.

src/SonarLint.CSharp/Rules/FieldShouldBeReadonly.cs
tvsonar: @benzonico I don't mind transforming this method to the suggested format. But I think in this case this `Collect*` pattern works fine here and wouldn't change this method to use a different pattern then the other four methods.

benzonico: IMO the 4 should be refactored to something I suggested to go for functional programming given what you're doing. 
As this is opiniated and I'm not the one who's going to maintain the code I'll let you rely on that but I tend to think this is a mistake you're going to pay in the maintenance of this check.

PR #929, opened 2016-06-28
SLVS-943 S2333 should handle missing constructs for unsafe redundancy…

tvsonar:
… check

### DIFF
--- a/src/SonarLint.CSharp/Rules/RedundantModifier.cs
+++ b/src/SonarLint.CSharp/Rules/RedundantModifier.cs
@@ -108,7 +108,7 @@ namespace SonarLint.Rules.CSharp
             if (TryGetUnsafeKeyword(typeDeclaration, out unsafeKeyword))
             {
                 MarkAllUnsafeBlockInside(typeDeclaration, context);
-                if (!HasUnsafeConstructInside(typeDeclaration))
+                if (!HasUnsafeConstructInside(typeDeclaration, context.SemanticModel))
                 {
                     ReportOnUnsafeBlock(context, unsafeKeyword.GetLocation());
                 }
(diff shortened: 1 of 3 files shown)


src/SonarLint.CSharp/Rules/RedundantModifier.cs
```diff
@@ -138,17 +138,85 @@ private static void CheckForUnnecessaryUnsafeBlocksBelow(TypeDeclarationSyntax t
...
                     }
                 }
             }
         }
 
-        private static bool HasUnsafeConstructInside(SyntaxNode container)
+        private static bool HasUnsafeConstructInside(SyntaxNode container, SemanticModel semanticModel)
         {
-            return container.DescendantNodes().Any(node => UnsafeConstructKinds.Contains(node.Kind()));
+            var isUnsafe = container.DescendantNodes().Any(node => UnsafeConstructKinds.Contains(node.Kind()));
```
m-g-sonar: I would definitely prefer to have a single `return` statement with calls to dedicated methods checking each cases, separated by OR operators.

src/SonarLint.CSharp/Rules/RedundantModifier.cs
```diff
@@ -138,17 +138,85 @@ private static void CheckForUnnecessaryUnsafeBlocksBelow(TypeDeclarationSyntax t
...
+
+            var currentType = type;
+            while (true)
+            {
+                switch (currentType.TypeKind)
+                {
+                    case TypeKind.Pointer:
+                        return true;
+                    case TypeKind.Array:
+                        currentType = ((IArrayTypeSymbol)currentType).ElementType;
```
m-g-sonar: I would prefer to have a recursive call to `IsUnsafe` using the element type rather to use a `while(true)`.

PR #932, opened 2016-06-29
SLVS-941 S3459 should not report on "fixed" arrays

tvsonar:


tvsonar:
Todo: https://github.com/SonarSource-VisualStudio/sonaranalyzer-csharp/pull/932/commits/aeeded56d9dba879bf6ec19f575e71ef19eb878c deserves a dedicated ticket in JIRA

tvsonar:
Todo: this PR should be cleaned up after https://github.com/SonarSource-VisualStudio/sonaranalyzer-csharp/pull/925 is merged.

src/SonarLint.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -124,6 +124,18 @@ protected override void Initialize(SonarAnalysisContext context)
                 SymbolKind.NamedType);
         }
 
+        private static bool IsNonInitialized(VariableDeclaratorSyntax declarator)
```
m-g-sonar: I would rename the method to `IsInitializedOrFixed`, and return the opposite. I don't negation in methods as it usually makes them hard to understand in case of double negations.

src/SonarLint.CSharp/Rules/NotAssignedPrivateMember.cs
```diff
@@ -124,6 +124,18 @@ protected override void Initialize(SonarAnalysisContext context)
...
+        {
+            var fieldDeclaration = declarator.Parent.Parent as BaseFieldDeclarationSyntax;
+            if (fieldDeclaration != null &&
+                fieldDeclaration.Modifiers.Any(m => m.IsKind(SyntaxKind.FixedKeyword)))
```
m-g-sonar: The JIRA ticket specifically targets arrays, but it seems to me that you are checking here for any field having the "fixed" modifier. Does this modifier only applies to arrays?

tvsonar: Yes, only arrays can be fixed.

PR #967, opened 2016-08-17
SLVS-971 S2328 should report only on the first occurrence of a field

tvsonar:


src/SonarLint.CSharp/Rules/GetHashCodeMutable.cs
```diff
@@ -93,22 +94,51 @@ protected override void Initialize(SonarAnalysisContext context)
...
-                },
-                SyntaxKind.MethodDeclaration);
+
+                    syntaxNodes.Add(identifierSymbol, new List<IdentifierNameSyntax>());
+                }
+
+                syntaxNodes[identifierSymbol].Add(identifier);
+            }
+
+            foreach (var identifierReferences in syntaxNodes.Values)
```
benzonico: is there a reason to loop a second time on this ? you can report directly in the first loop. You still need the Dictionary but no need for a 2nd loop.

PR #969, opened 2016-08-17
SLVS-976 S3249 should report only on the first occurrence of an inval…

tvsonar:
…id call


src/SonarLint.CSharp/Rules/GetHashCodeEqualsOverride.cs
```diff
@@ -108,7 +124,7 @@ protected override void Initialize(SonarAnalysisContext context)
             if (invokedMethod.IsInType(KnownType.System_Object) &&
                 !IsEqualsCallInGuardCondition(invocation, invokedMethod))
             {
-                context.ReportDiagnostic(Diagnostic.Create(Rule, invocation.GetLocation(), invokedMethod.Name));
+                issueLocations.Add(invocation.GetLocation());
```
benzonico: Why modify a parameter instead of returning a list ?

tvsonar: Good point.

PR #973, opened 2016-08-18
SLVS-991 S2259 should report on indexers

tvsonar:


its/expected/akka.net/Akka-{5DEDDF90-37F0-48D3-A0B0-A5CBD8A7E377}-S2259.json
```diff
@@ -0,0 +1,17 @@
+{
+"issues":  [
+{
+"id":  "S2259",
+"message":  "\"extraArgsValues\" is null on at least one execution path.",
```
tvsonar: False positive due to not having integer evaluation.

PR #983, opened 2016-08-23
SLVS-999 S2583 should not report on full logical && and || expressions

tvsonar:


src/SonarLint.CSharp/Rules/ConditionEvaluatesToConstant.cs
```diff
@@ -123,5 +122,11 @@ private static void CollectConditions(ConditionEvaluatedEventArgs args, HashSet<
                 conditionFalse.Add(args.Condition);
             }
         }
+
+        private static readonly ISet<SyntaxKind> OmittedSyntaxKinds = ImmutableHashSet.Create(
+            SyntaxKind.LogicalAndExpression,
+            SyntaxKind.LogicalOrExpression,
+            SyntaxKind.TrueLiteralExpression,
```
benzonico: True Literal and FalseLiteral are not tested as far as I can see.

src/SonarLint.CSharp/Rules/ConditionEvaluatesToConstant.cs
```diff
@@ -123,5 +122,11 @@ private static void CollectConditions(ConditionEvaluatedEventArgs args, HashSet<
                 conditionFalse.Add(args.Condition);
             }
         }
+
+        private static readonly ISet<SyntaxKind> OmittedSyntaxKinds = ImmutableHashSet.Create(
```
benzonico: I am a bit bothered by the approach here : I find it less explicit than what was done in java where the special case of if(a && b) is hardcoded. Here it is rather implicit as any condition that is a LogicalAnd/OR will be excluded without a way to grab the context in which this might appear.

src/SonarLint.CSharp/Rules/ConditionEvaluatesToConstant.cs
tvsonar: Thanks. Added the missing test cases.

src/SonarLint.CSharp/Rules/ConditionEvaluatesToConstant.cs
tvsonar: I think it's quite nice actually. Whenever we see a constant true or false logical && or ||, we know that either the left or the right operand was constant, and thus we already reported on it. (So there's no need to report on the complete logical && or || although its value is a constant)

I added a minor change to this PR in https://github.com/SonarSource-VisualStudio/sonaranalyzer-csharp/pull/983/commits/2d95d6644ebf819257fd42e4235eaf2b61b6c701, which makes a bit more obvious what happens. (A constant true or false SV is pushed onto the stack after we processed the right operand of && or ||)

PR #987, opened 2016-08-24
SLVS-1001 SE should pop values in the correct order for relational op…

tvsonar:
…erators


tvsonar:
Tests should be added, but it's not easy at this stage, because there's no transitivity, so relations are only calculated on operations of the same operands. But all these operations are binary operation and all binary operations were messed up before.

benzonico:
You should most probably introduce direct testing to solve this. 
If you are not able to write a test demonstrating the issue then this fix is useless. (yes, this is exagerating but : no issue-> no problem-> no need for a solution).

PR #989, opened 2016-08-26
SLVS-1003 Improve versioning to allow CI job change the AssemblyFileVersion

tvsonar:


tvsonar:
Note to self: this ticket requires a lot of manual testing, SonarLint for Visual Studio depends on the assembly version to deduplicate analyzers in the connected mode.

PR #990, opened 2016-08-29
SAVB-31 Regression testing should be added

tvsonar:


src/SonarLint.CSharp/Helpers/SymbolicExecution/ExplodedGraph.cs
```diff
@@ -374,7 +374,9 @@ protected override void VisitInstruction(ExplodedGraphNode node)
                     break;
 
                 case SyntaxKind.ImplicitElementAccess:
-                    newProgramState = newProgramState.PopValues(((ImplicitElementAccessSyntax)instruction).ArgumentList?.Arguments.Count ?? 0);
+                    newProgramState = newProgramState
+                        .PopValues(((ImplicitElementAccessSyntax)instruction).ArgumentList?.Arguments.Count ?? 0)
+                        .PushValue(new SymbolicValue());
```
m-g-sonar: To keep consistency with other cases from the switch, it would be better to separate these 2 chained method calls into 2 distinct statements (1 to pop, 1 to push).

tvsonar: I think new code should use this format, I was trying to use it lately, but not changing all usages at once.

m-g-sonar: Okay, then don't change anything now, but it would be nice to changing everything in another commit soon.

PR #1002, opened 2016-09-07
SAVB-35 Add self-assigned variable rule

tvsonar:


its/expected/Ember-MM/Ember Media Manager-{9B57D3AB-AF12-4012-B945-284C2448DC81}-S1656.json
```diff
@@ -0,0 +1,30 @@
+{
+"issues":  [
+{
+"id":  "S1656",
+"message":  "Remove or correct this useless self-assignment.",
```
tvsonar: This is a new rule, so there's no comparison basis for it. Both issues seem to be real problems.

src/SonarLint.VisualBasic/Rules/SelfAssignment.cs
```diff
@@ -0,0 +1,54 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+using SonarLint.Common;
+using SonarLint.Common.Sqale;
+using SonarLint.Helpers;
+using SonarLint.Helpers.VisualBasic;
+
+namespace SonarLint.Rules.VisualBasic
+{
+    [DiagnosticAnalyzer(LanguageNames.VisualBasic)]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.DataReliability)]
```
m-g-sonar: You can move this part to the parent class `SelfAssignmentBase`, as these annotastions are shared between the two implementation of the rule.

src/Tests/SonarLint.UnitTest/Rules/SelfAssignmentTest.cs
```diff
@@ -30,7 +29,8 @@ public class SelfAssignmentTest
         [TestCategory("Rule")]
         public void SelfAssignment()
         {
-            Verifier.VerifyAnalyzer(@"TestCases\SelfAssignment.cs", new SelfAssignment());
+            Verifier.VerifyAnalyzer(@"TestCases\SelfAssignment.cs", new SonarLint.Rules.CSharp.SelfAssignment());
+            Verifier.VerifyAnalyzer(@"TestCases\SelfAssignment.vb", new SonarLint.Rules.VisualBasic.SelfAssignment());
```
m-g-sonar: Use a different unit test, in order to test both analyzer independently.

src/SonarLint.VisualBasic/Rules/SelfAssignment.cs
tvsonar: Investigated this a bit, and you're right, I can easily move most of the attributes to the base class. Exceptions are: `DiagnosticAnalyzer` and `Rule`.

PR #1004, opened 2016-09-07
SAVB-29 Add comment at line end rule

tvsonar:


src/SonarLint.VisualBasic/Rules/CommentLineEnd.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        internal static readonly DiagnosticDescriptor Rule =
+            new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category,
+                RuleSeverity.ToDiagnosticSeverity(), false,
+                helpLinkUri: DiagnosticId.GetHelpLink(),
+                description: Description);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Rule); } }
+
+        private const string DefaultPattern = @"^'\s*\S+$";
```
alban-auzeill: Is case of trailing whitespaces, it's confusing to see an issue:
* RSPEC-139 Comments should not be located at the end of lines of code

I prefer this default pattern `^'\s*\S+\s*$` and let the following rule highlight the whitespaces:
* RSPEC-1131 Lines should not end with trailing whitespaces

alban-auzeill: Do we want to raise issues with empty or blank comment ?

tvsonar: Thanks for the trailing whitespace finding.
Also, I don't think we should report on empty comments in this rule.

PR #1007, opened 2016-09-07
SAVB-32 Add matching branches in if-else chain rule

tvsonar:


src/SonarLint.Extra/Rules.Description/S1871_vb.html
```diff
@@ -0,0 +1,73 @@
+<p>
+    Having two <code>Cases</code> in the same <code>Select</code> statement or branches in the same <code>If</code> structure
```
alban-auzeill: How this file is generated ? from which RSPEC? I can't find this content [here](https://jira.sonarsource.com/browse/RSPEC-1871)

src/Tests/SonarLint.UnitTest/TestCases/ConditionalStructureSameImplementation_If.vb
```diff
@@ -0,0 +1,41 @@
...
+            If someCondition Then DoSomething1() Else DoSomething2() : DoSomething1() End
+
+            If (someCondition1) Then
+                DoSomething1()
+                DoSomething1()
+            ElseIf (someCondition2) Then
+
+                DoSomething1()
+                DoSomething2()
+            Else
```
alban-auzeill: Could you add a test with more than one `ElseIf` ? My understanding let me think that the implementation does not support them.

src/SonarLint.Extra/Rules.Description/S1871_vb.html
tvsonar: It's manual, not generated for the time being. It will be automated the same way as you're automating it in other language plugins. 

I added the missing subtask.

src/Tests/SonarLint.UnitTest/TestCases/ConditionalStructureSameImplementation_If.vb
tvsonar: Added another `ElseIf`.

PR #1009, opened 2016-09-09
SLVS-1008 Recognizing "nameof()" should check if there's a "nameof" m…

tvsonar:
…ethod symbol

### DIFF
--- a/src/SonarLint.CSharp/Helpers/SyntaxHelper.cs
+++ b/src/SonarLint.CSharp/Helpers/SyntaxHelper.cs
@@ -110,13 +110,25 @@ namespace SonarLint.Helpers
             return false;
         }
 
-        public static bool IsInNameofCall(this ExpressionSyntax expression)
+        public static bool IsInNameofCall(this ExpressionSyntax expression, SemanticModel semanticModel)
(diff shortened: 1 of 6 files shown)


src/SonarLint.CSharp/Helpers/SyntaxHelper.cs
```diff
@@ -110,10 +110,22 @@ public static bool IsOnThis(this ExpressionSyntax expression)
...
+
+            var calledSymbol = semanticModel.GetSymbolInfo(nameofCall).Symbol as IMethodSymbol;
+            if (calledSymbol != null)
+            {
+                return false;
+            }
+
             var nameofIdentifier = (nameofCall?.Expression as IdentifierNameSyntax)?.Identifier;
             return nameofIdentifier.HasValue &&
                 (nameofIdentifier.Value.IsKind(SyntaxKind.NameOfKeyword) || (nameofIdentifier.Value.ToString() == SyntaxFacts.GetText(SyntaxKind.NameOfKeyword)));
```
m-g-sonar: The first part before the `||` operator should be removed.

PR #1013, opened 2016-09-12
SLVS-1016 Add deeply nested control flow statement rule

tvsonar:


src/SonarLint.CSharp/Rules/FunctionNestingDepth.cs
```diff
@@ -82,16 +73,13 @@ private void CheckFunctionNestingDepth(SyntaxNodeAnalysisContext context)
             walker.Visit(context.Node);
         }
 
-        private class NestingDepthWalker : CSharpSyntaxWalker
+        private new class NestingDepthWalker : CSharpSyntaxWalker
         {
-            private readonly int maximumNestingDepth;
-            private readonly Action<SyntaxToken> actionMaximumExceeded;
-            private int currentDepth = 0;
+            private readonly FunctionNestingDepthBase.NestingDepthWalker walker;
```
m-g-sonar: This sounds wrong to me. `FunctionNestingDepthBase.NestingDepthWalker` is not really a walker, but only a base containing call-back logic (if the called back action walks the tree, it does not know it). I would rename it.

PR #1016, opened 2016-09-12
Fixes for Cyclomatic Complexity calculation

erik-am:
SonarLint doesn't calculate the McCabe Cyclomatic Complexity correctly, because it's counting return statements as well.

Cyclomatic complexity should be equal to **the number of branch points in your code + 1**.

Let's look at some examples, to make clear why the current calculation doesn't make sense.

```
if (expression) {
	// statement 1
}
// statement 2
```
This piece of code has a complexity of 2. While statement 2 is always executed, the complexity is in the fact that statement 1 is sometimes executed and sometimes not.

```
if (expression) {
	// statement 1
} else {
	// statement 2
}
```
This piece of code also has a complexity of 2. It's either executing statement 1 or 2, so there's one decision point.

```
if (expression) {
	// statement 1
	return something;
} else {
	// statement 2
	return something else;
}
```

Now SonarLint will actually increase the complexity 4! But the number of decision points did not increase.
If you are already inside the else branch, you are _always_ going to execute the return, so there's no decision point.

The cyclomatic complexity is a very useful metric that relates to testability.
The higher the cyclomatic complexity, the more tests you will need to cover all the branches.

Currently, because SonarLint incorrectly counts return statements, you can decrease the complexity of the previous example to 2 by refactoring to:

```
var output;
if (expression) {
	// statement 1
	output = something;
} else {
	// statement 2
	output = something else;
}
return output;
```

Now SonarLint reports the correct complexity (2). However, this refactoring did not decrease the number of decision points and it also did not decrease the effort to unit test all the paths.

With the same reasoning you should also not raise cyclomatic complexity for a `throw` statement (which is currently counted for VB.NET), because it's an unescapable statement.
Only if you put an `if` branch around it, then the if will increase the number of paths, but the throw statement on itself won't.

(If you want to be really puristic on cyclomatic complexity, then return statements should actually decrease cyclomatic complexity, where cyclomatic complexity = number of branch points - number of exit points + 2, but I wouldn't go that far and stick to the simple definition, i.e., branch points + 1)

### Proposal

I removed counting the return statements from the C# implementation. For VB.NET there were even more changes to make.

I hope this PR will be accepted. The McCabe Cyclomatic Complexity is an important and well-known metric for code quality. It would really help if SonarLint follows the definition of this rule correctly, so that code quality experts can speak the same language.

### DIFF
--- a/src/SonarLint.CSharp/Metrics/Metrics.cs
+++ b/src/SonarLint.CSharp/Metrics/Metrics.cs
@@ -102,18 +102,7 @@ namespace SonarLint.Common.CSharp
                         n =>
                             // TODO What about the null coalescing operator?
                             ComplexityIncreasingKinds.Contains(n.Kind()) ||
-                            IsReturnButNotLast(n) ||
                             IsFunctionWithBody(n));
-        }
-        private bool IsReturnButNotLast(SyntaxNode node)
-        {
-            return node.IsKind(SyntaxKind.ReturnStatement) && !IsLastStatement(node);
-        }
-        private bool IsLastStatement(SyntaxNode node)
-        {
-            var nextToken = node.GetLastToken().GetNextToken();
-            return nextToken.Parent.IsKind(SyntaxKind.Block) &&
-                IsFunction(nextToken.Parent.Parent);
         }
 
         private static readonly SyntaxKind[] TriviaKinds =
--- a/src/SonarLint.VisualBasic/Metrics/Metrics.cs
+++ b/src/SonarLint.VisualBasic/Metrics/Metrics.cs
@@ -202,6 +202,7 @@ namespace SonarLint.Common.VisualBasic
             SyntaxKind.DeclareSubStatement,
             SyntaxKind.DeclareFunctionStatement,
             SyntaxKind.SubNewStatement,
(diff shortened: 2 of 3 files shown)


tvsonar:
@erik-am Thanks for this PR. This will take some time to review a merge. We are aware that the complexity calculation is (terribly) wrong. We have an MMF on it here: https://jira.sonarsource.com/browse/MMF-151.

There are some open questions regarding lambdas, autoproperties, ... see the comments on the JIRA ticket. Also, the current implementation returns different complexity values for methods with body and methods with expression bodies. Decision needs to be made on all these question. And ideally we want to have the same complexity calculation across languages (Java, C#)

Also, Visual Studio Enterprise already calculate complexity metrics, we haven't checked the reported values, but in an ideal world, SonarQube/SonarLint should report exactly the same values.

p3pijn:
For reference, also see the following discussion on Google Groups regarding the cyclomatic complexity calculation:
https://groups.google.com/forum/m/?utm_medium=email&utm_source=footer#!msg/sonarqube/FVoRAr4QlXU/FkuOC3_EAgAJ

Corniel:
What is the status of this change?

PR #1018, opened 2016-09-13
S3216: added support for Task<TResult>

overeemm:
Rule S3216 did not signal ConfigureAwait on Task<TResult>, causing many expressions to be missed.


overeemm:
Not sure what caused the failing of the build. Did I broke a test or is it about performance?

tvsonar:
@overeemm Thanks for this fix. The failing test says that there were new issues found by your modification but you haven't added any expected files for them. Expected files are [here](https://github.com/SonarSource-VisualStudio/sonaranalyzer-dotnet/tree/master/its/expected), and regression testing is run by regression-test.bat. Don't worry about it, I'll have a look.

src/SonarLint.CSharp/Rules/TaskConfigureAwait.cs
```diff
@@ -78,7 +78,7 @@ protected override void Initialize(SonarAnalysisContext context)
                     }
 
                     var type = c.SemanticModel.GetTypeInfo(expression).Type;
-                    if (type.Is(KnownType.System_Threading_Tasks_Task))
+                    if (type.Is(KnownType.System_Threading_Tasks_Task) || type.OriginalDefinition.Is(KnownType.System_Threading_Tasks_Task_TResult))
```
tvsonar: Note to self: `Task<T>` derives from `Task`, so we could use `type.DerivesFrom(KnownType.System_Threading_Tasks_Task)`, instead of the `.Is()`, and that would allow handling custom subtypes of `Task`.

tvsonar:
@overeemm I modified a bit this PR, and merged to master with this commit: https://github.com/SonarSource-VisualStudio/sonaranalyzer-dotnet/commit/18e82e8bafcca211d8dd82c4f9bc88c135ba0d4d

Thanks again for the fix.

PR #1019, opened 2016-09-13
SLVS-1004 Add missing non-null constraint to simple assignments and v…

tvsonar:
…ariable declarations


its/expected/akka.net/Akka.Remote.TestKit-{E5957C3E-2B1E-469F-A680-7953B4DEA31B}-S2583.json
```diff
@@ -0,0 +1,30 @@
+{
+"issues":  [
+{
+"id":  "S2583",
+"message":  "Change this condition so that it does not always evaluate to \"true\".",
```
tvsonar: real issue, the identifier references an enum value, which is then compared to `null`. But enum values are non-nullable, so the comparison is useless.

PR #1034, opened 2016-10-05
Allow implementations of the (de)serialization constructor not using the streaming context.

Corniel:
In most occasions, implementing this constructor will not take use the streaming context at all. However, the compiler forces you to have the signature for this constructor that is private in a lot of cases, including sealed classes en structs. Therefore is should not raise a warning imho.


Corniel:
Note that I forgot to update the CodeFix test in this pull request. I fixed that too.

tvsonar:
@Corniel Thanks for this fix. 
It seems to me that your implementation is only checking if the constructors signature is matching the expected one, but it doesn't check if the constructor belongs to a serializable class or not. To me it seems that this constructor is not tied to the `Serializable` attribute, but to the `ISerializable` interface (https://msdn.microsoft.com/en-us/library/system.runtime.serialization.iserializable.aspx#Anchor_2). I'm not that familiar with serializable constructors, so could you confirm this to me?

Corniel:
It does not check on that. As far as I know (I just tested it again), this ctor is only used for ISerializable objects, not for objects with a Serializable attribute. You can argue, that having a ctor with this arguments NOT beging ISerialzable always should lead to a warning. I can add that too if you want.

Corniel:
This one can be closed. I made a 'mistake' that leaded to a new request.

PR #1037, opened 2016-10-14
Avoid implicit conversions in VB.NET

Corniel:
Only use short-circuiting operators, as further analyzing the expression will not change the outcome. It is a typical VB.NET issue, as the none short-circuiting keywords are the default onces. Bitwise operators are taken into account. Types that can be implicitly casted to a boolean are not taken into account.


Corniel:
I have to admit that I'm still a newbie when it comes to GIT. So now this (unintended) contains two request: avoiding implicit conversions, and the use of the short-circuiting operetors. Sorry for that.

src/SonarAnalyzer.VisualBasic/Rules/AvoidImplicitConversions.cs
```diff
@@ -0,0 +1,98 @@
...
+                helpLinkUri: DiagnosticId.GetHelpLink(),
+                description: Description);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Rule); } }
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                DetactImplicitConversions,
+                SyntaxKind.EqualsValue,
```
tvsonar: I think a couple of cases are missing form here:
* assignment statements (simple assignment, `+=`, ...)
* arguments passed to methods, where the argument is `CInt(y)` and the parameter type is `Integer`
* Hidden return statements in lambdas: `Dim a As Func(Of Integer) = Function () CInt(5)`

See for C# in https://github.com/SonarSource-VisualStudio/sonaranalyzer-dotnet/blob/master/src/SonarAnalyzer.CSharp/Rules/LossOfFractionInDivision.cs#L76.

src/SonarAnalyzer.VisualBasic/Rules/AvoidImplicitConversions.cs
```diff
@@ -0,0 +1,98 @@
...
+                SyntaxKind.ReturnStatement);
+        }
+
+        private void DetactImplicitConversions(SyntaxNodeAnalysisContext context)
```
tvsonar: Spelling mistake on the method name. Also the method can probably be `static`.

src/SonarAnalyzer.VisualBasic/Rules/AvoidImplicitConversions.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            var keyword = syntax.Keyword.Kind();
+            string type;
+            if (Types.TryGetValue(keyword, out type))
+            {
+                context.ReportDiagnostic(Diagnostic.Create(Rule, syntax.GetLocation(), type));
+            }
+        }
+
+        private static readonly Dictionary<SyntaxKind, string> Types = new Dictionary<SyntaxKind, string>
```
tvsonar: `ImmutableDictionary` expresses the intent better.

src/SonarAnalyzer.VisualBasic/Rules/AvoidImplicitConversions.cs
```diff
@@ -0,0 +1,98 @@
...
+{
+    [DiagnosticAnalyzer(LanguageNames.VisualBasic)]
+    [SqaleConstantRemediation("2min")]
+    [SqaleSubCharacteristic(SqaleSubCharacteristic.Readability)]
+    [Rule(DiagnosticId, RuleSeverity, Title, IsActivatedByDefault)]
+    [Tags(Tag.Confusing)]
+    public class AvoidImplicitConversions : SonarDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S????";
+         internal const string Title = "Avoid implicit conversions";
```
tvsonar: I'm not a VB.NET developer, could you provide some links which state that this is the best practice?
The title and description will need to be improved. According to https://msdn.microsoft.com/en-us/library/kca3w8x6.aspx `CStr`, `CInt`, ... are "explicit conversions", so the title is already misleading.

Each of our rules have a proper rule description in our rule database: [See here](https://jira.sonarsource.com/browse/RSPEC-3760?jql=project%20%3D%20RSPEC%20AND%20issuetype%20in%20standardIssueTypes()%20AND%20status%20%3D%20Active). Before implementation there should be a finalized description. You won't be able to edit the spec, but I created the following rule-idea ticket, on which you can comment after registration: https://jira.sonarsource.com/browse/SLVS-1079

Corniel: It is true that official documentation states that it is a conversion, not a casting, but the whole naming CInt vs CType(expression ,Integer) suggests otherwise. I've seen a lot of code that uses this where casting was intended. The only reason this construct exists, is that it allowed easy porting from VB4..VB6, to VB.NET. If you look to what the compiler does, it uses functionality that is part of Microsoft.VisualBasic.dll.

That's why I choose the name implicit. But, again, it is up to debate. I'll try to find a source that backups my (strong) opinion at the matter.

src/SonarAnalyzer.VisualBasic/Rules/AvoidImplicitConversions.cs
Corniel: VB(.NET) does (unfortunately) not support +=, /=, -=, <<=, ect...

The second part will automaticly lead to => CType(Integer, Integer) => Integer. Unnecessary casts are already handles by other rules. So you can argue that is not needed?!

I'll will add a hidden return statement in lambda. Sounds like something I missed. Thanks for pointing out.

src/SonarAnalyzer.VisualBasic/Rules/AvoidImplicitConversions.cs
Corniel: I found a discussion that proves to me this conversion instead of casting is misleading: http://stackoverflow.com/questions/306937/vb6-cast-expression

And here you can read how those conversions work: https://msdn.microsoft.com/en-us/library/s2dy91zy.aspx

If you THINK that you are casting, this behaviour is bad, and in a lot of cases wrong.

Corniel:
I updated the implementation according to your suggestions, but is now not picked-up?

Corniel:
Any update @tvsonar ?

tvsonar:
@Corniel I'm going to work on this in the next couple of days. The " short-circuiting operators" rule should be part of version 1.20.0. And I'll have to see the other one in detail, and create a specification for that etc. So on that one, I'm not sure yet.

Corniel:
Perfectly fine. Most important (to me) is that I know what the status is. We all have more to do. :)

tvsonar:
@Corniel I merged the short-circuit operation rule though https://github.com/SonarSource-VisualStudio/sonaranalyzer-dotnet/commit/07cbb90c3e2acf03dc6192f463f8994bda251eb8 and https://github.com/SonarSource-VisualStudio/sonaranalyzer-dotnet/commit/dad183579d4827e91bce0925af5f89bd17b93a0c

Could you please clean up this PR, to only contain the changes for the second rule. Thanks.

Corniel:
I will do that, and create a new request for it. Thanks for merging.

Corniel:
Resubmitted.

PR #1044, opened 2016-10-25
SLVS-1080 S1144 should work with "dynamic" references

tvsonar:


src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
```diff
@@ -324,21 +320,33 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                symbol = methodSymbol.ReducedFrom;
             }
+
+            return symbol?.OriginalDefinition;
+        }
+
+        private static IEnumerable<ISymbol> GetAllCandidateSymbols(SymbolInfo symbolInfo)
+        {
+            return new[] { symbolInfo.Symbol }
+                .Concat(symbolInfo.CandidateSymbols)
```
Evangelink: Concat crashes if called with null. Should we check/assert "symbolInfo.CandidateSymbols" is never null?

src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
```diff
@@ -324,21 +320,33 @@ protected override void Initialize(SonarAnalysisContext context)
...
+            usedSymbols.UnionWith(candidateSymbols);
+        }
 
-                if (symbol != null)
-                {
-                    usedSymbols.Add(symbol.OriginalDefinition);
-                }
+        private static ISymbol GetOriginalSymbol(ISymbol candidateSymbol)
+        {
+            var symbol = candidateSymbol;
```
tvsonar: Because we don't like reassigned parameters, exceptions. https://jira.sonarsource.com/browse/RSPEC-2829

src/SonarAnalyzer.CSharp/Rules/UnusedPrivateMember.cs
tvsonar: I checked, and `CandidateSymbols` is never `null`: http://source.roslyn.codeplex.com/#Microsoft.CodeAnalysis/Compilation/SymbolInfo.cs,32

PR #1045, opened 2016-10-25
Slvs 1070

tvsonar:


src/SonarAnalyzer.Common/Rules/Utilities/TokenTypeAnalyzerBase.cs
```diff
@@ -194,6 +202,15 @@ private void ClassifyIdentifier(SyntaxToken token, ISymbol symbol)
                 }
             }
 
+            private int myVar;
+
+            public int MyProperty
```
SonarTech: MINOR Make this an auto-implemented property and remove its backing field. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=csharpsquid%3AS2292)

SonarTech:
SonarQube analysis reported 1 issue
* MINOR 1 minor

Watch the comments in this conversation to review them.

src/SonarAnalyzer.Common/Rules/Utilities/TokenTypeAnalyzerBase.cs
```diff
@@ -181,6 +181,14 @@ private void ClassifyIdentifier(SyntaxToken token, ISymbol symbol)
                     return;
                 }
 
+                if (token.ToString() == "value" &&
```
Evangelink: Is it worth putting a const here for "value"?

tvsonar: I don't think so. String literals will be interned, so they will automatically reference the same constant. See here
* https://msdn.microsoft.com/en-us/library/system.string.intern(v=vs.110).aspx#Anchor_2
* https://blogs.msdn.microsoft.com/ericlippert/2009/09/28/string-interning-and-string-empty/

PR #1058, opened 2016-11-30
MemberOverrideCallsBaseMember: Adding attributes to a derived class i…

Evangelink:
…s a difference

So don't warn there

### DIFF
--- a/src/SonarAnalyzer.CSharp/Rules/MemberOverrideCallsBaseMember.cs
+++ b/src/SonarAnalyzer.CSharp/Rules/MemberOverrideCallsBaseMember.cs
@@ -110,6 +110,11 @@ namespace SonarAnalyzer.Rules.CSharp
                 return false;
             }
 
+            if (SymbolHelper.IsAnyAttributeInOverridingChain(propertySymbol))
+            {
+                return false;
+            }
+
             return CheckGetAccessorIfAny(propertySyntax, propertySymbol, semanticModel) &&
                 CheckSetAccessorIfAny(propertySyntax, propertySymbol, semanticModel);
         }
@@ -190,13 +195,7 @@ namespace SonarAnalyzer.Rules.CSharp
 
             var methodSymbol = semanticModel.GetDeclaredSymbol(methodSyntax);
 
-            if (methodSymbol == null ||
-                !methodSymbol.IsOverride ||
-                methodSymbol.IsSealed ||
-                methodSymbol.OverriddenMethod == null ||
-                IgnoredMethodNames.Contains(methodSymbol.Name) ||
-                methodSymbol.Parameters.Any(p => p.HasExplicitDefaultValue) ||
-                methodSymbol.OverriddenMethod.Parameters.Any(p => p.HasExplicitDefaultValue))
+            if (IsMethodSymbolExcluded(methodSymbol))
             {
                 return false;
             }
@@ -219,6 +218,19 @@ namespace SonarAnalyzer.Rules.CSharp
             }
 
             return AreArgumentsMatchParameters(methodSymbol, semanticModel, expressionToCheck, invokedMethod);
+        }
+
+        private static bool IsMethodSymbolExcluded(IMethodSymbol methodSymbol)
+        {
+            return
+                methodSymbol == null ||
+                !methodSymbol.IsOverride ||
+                methodSymbol.IsSealed ||
+                methodSymbol.OverriddenMethod == null ||
+                IgnoredMethodNames.Contains(methodSymbol.Name) ||
+                methodSymbol.Parameters.Any(p => p.HasExplicitDefaultValue) ||
+                methodSymbol.OverriddenMethod.Parameters.Any(p => p.HasExplicitDefaultValue) ||
+                SymbolHelper.IsAnyAttributeInOverridingChain(methodSymbol);
         }
 
         private static bool HasDocumentationComment(SyntaxNode node)
--- a/src/SonarAnalyzer.Common/Helpers/SymbolHelper.cs
+++ b/src/SonarAnalyzer.Common/Helpers/SymbolHelper.cs
@@ -200,5 +200,37 @@ namespace SonarAnalyzer.Helpers
 
             return Enumerable.Empty<IParameterSymbol>();
         }
+
+        public static bool IsAnyAttributeInOverridingChain(IPropertySymbol propertySymbol)
+        {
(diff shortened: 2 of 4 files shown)


src/SonarAnalyzer.Common/Helpers/SymbolHelper.cs
```diff
@@ -200,5 +200,37 @@ public static IEnumerable<IParameterSymbol> GetParameters(this ISymbol symbol)
 
             return Enumerable.Empty<IParameterSymbol>();
         }
+
+        public static bool IsAnyAttributeInOverridingChain(IPropertySymbol propertySymbol)
+        {
+            var localPropertySymbol = propertySymbol;
```
tvsonar: A better name would be `currentPropertySymbol`.

src/SonarAnalyzer.Common/Helpers/SymbolHelper.cs
```diff
@@ -200,5 +200,37 @@ public static IEnumerable<IParameterSymbol> GetParameters(this ISymbol symbol)
 
             return Enumerable.Empty<IParameterSymbol>();
         }
+
+        public static bool IsAnyAttributeInOverridingChain(IPropertySymbol propertySymbol)
```
tvsonar: The body of this method almost matches the same method with `IMethodSymbol`. Can't we merge some logic to avoid copy-pasted code?

src/Tests/SonarAnalyzer.UnitTest/TestCases/MemberOverrideCallsBaseMember.Fixed.cs
```diff
@@ -49,9 +69,19 @@ public override int GetHashCode()
     class Derived : Base
     {
         public override int MyProperty3 => 42;
+        [Description]
```
tvsonar: We only test direct parent-child relationships. Let's add longer chains too.

src/SonarAnalyzer.CSharp/Rules/MemberOverrideCallsBaseMember.cs
```diff
@@ -196,7 +201,9 @@ private static bool IsMethodCandidate(MethodDeclarationSyntax methodSyntax, Sema
                 methodSymbol.OverriddenMethod == null ||
                 IgnoredMethodNames.Contains(methodSymbol.Name) ||
                 methodSymbol.Parameters.Any(p => p.HasExplicitDefaultValue) ||
-                methodSymbol.OverriddenMethod.Parameters.Any(p => p.HasExplicitDefaultValue))
+                methodSymbol.OverriddenMethod.Parameters.Any(p => p.HasExplicitDefaultValue) ||
+                SymbolHelper.IsAnyAttributeInOverridingChain(methodSymbol)
+                )
```
tvsonar: So far, we didn't put closing parenthesis on new line. We can, but then let's create a formatting guideline too, and enforce that on all .NET code bases.

src/SonarAnalyzer.Common/Helpers/SymbolHelper.cs
```diff
@@ -200,5 +200,32 @@ public static IEnumerable<IParameterSymbol> GetParameters(this ISymbol symbol)
...
+        {
+            return IsAnyAttributeInOverridingChain(propertySymbol, x => x.OverriddenProperty);
```
tvsonar: `x` should be named `property` or `propertySymbol`.

src/SonarAnalyzer.Common/Helpers/SymbolHelper.cs
```diff
@@ -200,5 +200,32 @@ public static IEnumerable<IParameterSymbol> GetParameters(this ISymbol symbol)
...
+        }
+
+        public static bool IsAnyAttributeInOverridingChain(IMethodSymbol methodSymbol)
+        {
+            return IsAnyAttributeInOverridingChain(methodSymbol, x => x.OverriddenMethod);
```
tvsonar: `x` should be named `method` or `methodSymbol`.

src/SonarAnalyzer.Common/Helpers/SymbolHelper.cs
```diff
@@ -200,5 +200,32 @@ public static IEnumerable<IParameterSymbol> GetParameters(this ISymbol symbol)
...
+        {
+            var currentSymbol = symbol;
+            while (currentSymbol != null)
+            {
+                if (currentSymbol.GetAttributes().Any())
+                {
+                    return true;
+                }
+
+                currentSymbol = getOverridenMember(currentSymbol);
```
tvsonar: We should only call the `Func` if the `currentSymbol.IsOverride` is `true`.

src/Tests/SonarAnalyzer.UnitTest/TestCases/MemberOverrideCallsBaseMember.Fixed.cs
```diff
@@ -129,4 +130,185 @@ public override int MyProperty2
             }
         }
     }
+
+    class MyAttributesTestCase
+    {
+        [InheritedAttribute]
```
tvsonar: Are you sure that these attributes properly resolve here? Test files are compiled on-by-one.

tvsonar: Also, there's no need to have `Inherited` attribute usages. We don't do anything based on that in the cod.e

src/SonarAnalyzer.CSharp/Rules/MemberOverrideCallsBaseMember.cs
Evangelink: Isn't better to use a method or variable instead of that big if?

src/Tests/SonarAnalyzer.UnitTest/TestCases/MemberOverrideCallsBaseMember.Fixed.cs
```diff
@@ -129,4 +129,114 @@ public override int MyProperty2
             }
         }
     }
+
+    [AttributeUsage(AttributeTargets.All)]
```
tvsonar: These `AttributeUsage` attributes are most probably useless here. We can remove them, and the behavior will be the same.

Evangelink:
Merged on branch v1.20 before future merge in master.

PR #1059, opened 2016-12-01
Added testcase that causes a NullReferenceException in rules

overeemm:
The added testcase causes a NullReference, because the expression in the OrderBy does not have a type. This of course causes exceptions during compilation. People can work around this issue by rewriting the expression-bodied method to a standard method.


tvsonar:
Nice catch @overeemm. Thanks for reporting and for the repro.

Evangelink:
We took your test case and fixed the associated code. The PR will be part of the next release (v1.21).
You can see the code [here](https://github.com/SonarSource-VisualStudio/sonaranalyzer-dotnet/commit/06ee3efaa017fd94cb6c51d02dc7895cb6b81ca4)

PR #1060, opened 2016-12-02
SLVS-1098: Do not throw exception on invalid input in "EmptyNullableValueAccess" rule

Evangelink:


Evangelink:
Merged on branch v1.20 before future merge in master.

PR #1066, opened 2016-12-08
SLVS-1097 - Fix crossreference pattern matching

valhristov:


src/SonarAnalyzer.RuleDocGenerator.Web/RuleImplementationMeta.cs
```diff
@@ -31,7 +31,7 @@ namespace SonarAnalyzer.RuleDocGenerator
 {
     public class RuleImplementationMeta
     {
-        internal const string CrosslinkPattern = "(Rule )(S[0-9]+)";
+        internal const string CrosslinkPattern = "\\{rule:(?:csharpsquid|vbnet):(S[0-9]+)\\}";
```
tvsonar: Let's use `([rR]ule\W+)?\{rule:(?:csharpsquid|vbnet):(?<ruleId>S[0-9]+)\}` instead. This allows an optional "Rule" prefix.

Evangelink:
LGTM

PR #1068, opened 2016-12-13
Rule S112 - System reserved exceptions should not be thrown

valhristov:


src/SonarAnalyzer.Utilities/Rules.Description/S112.html
```diff
@@ -0,0 +1,33 @@
+<p>Throwing such generic exceptions as <code>IndexOutOfRangeException</code>, <code>NullReferenceException</code>, <code>OutOfMemoryException</code>
```
tvsonar: There are missing exceptions here, for example "System.SystemException" is not listed

src/SonarAnalyzer.CSharp/Rules/ThrowReservedExceptions.cs
```diff
@@ -0,0 +1,60 @@
...
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(DiagnosticId)]
+    public class ThrowReservedExceptions : SonarDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S112";
+        internal const string MessageFormat = "{0} should not be thrown by user code.";
```
tvsonar: As discussed, the exception name should be between quotes.

src/SonarAnalyzer.CSharp/Rules/ThrowReservedExceptions.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        private static readonly ISet<string> ReservedExceptionTypeNames =
+            ImmutableHashSet.Create(new[]
```
tvsonar: `Create` takes `params`, you can use that too.

valhristov:
Added a VB.NET version of the rule, fixed all suggestions

PR #1070, opened 2016-12-14
SLVS-762 S1172 (unused parameters) should also run on public methods …

tvsonar:
…contained within private classes


src/SonarAnalyzer.Common/Helpers/SymbolHelper.cs
```diff
@@ -232,5 +232,25 @@ private static bool IsAnyAttributeInOverridingChain<TSymbol>(TSymbol symbol, Fun
 
             return false;
         }
+
+        public static Accessibility EffectiveAccessibility(this ISymbol symbol)
```
valhristov: This could be GetEffectiveAccessibility (or Calculate?), otherwise it sounds like a property name.

PR #1072, opened 2016-12-15
Rule S3256 - "String.IsNullOrEmpty" should be used.

valhristov:


src/SonarAnalyzer.CSharp/Rules/UseStringIsNullOrEmpty.cs
```diff
@@ -0,0 +1,122 @@
...
+                    var memberAccessExpression = invocationExpression.Expression as MemberAccessExpressionSyntax;
+                    if (memberAccessExpression == null)
+                    {
+                        return;
+                    }
+
+                    // .Equals
+                    var methodName = c.SemanticModel.GetSymbolInfo(memberAccessExpression.Name);
+
+                    if (methodName.Symbol.IsInType(KnownType.System_String))
```
tvsonar: I don't like deep nesting. Let's invert the condition and return early.

src/SonarAnalyzer.CSharp/Rules/UseStringIsNullOrEmpty.cs
```diff
@@ -0,0 +1,122 @@
...
+                    {
+                        var firstArgument = invocationExpression.ArgumentList.Arguments.FirstOrDefault();
```
tvsonar: Can `ArgumentList` be `null` on some invalid code? (Probably not, but it would be easy to be safe: `?.Arguments`)

src/SonarAnalyzer.CSharp/Rules/UseStringIsNullOrEmpty.cs
```diff
@@ -0,0 +1,122 @@
...
+                        if (firstArgument == null)
+                        {
+                            return;
+                        }
+
+                        // x.Equals(value), where x is string.Empty, "" or const ""
+                        if (IsStringIdentifier(firstArgument.Expression, c.SemanticModel) &&
+                            IsConstantEmptyString(memberAccessExpression.Expression, c.SemanticModel))
+                        {
+                            c.ReportDiagnostic(Diagnostic.Create(rule, invocationExpression.GetLocation(), MessageFormat));
```
tvsonar: We should `return` after the reporting, the next condition can never be true, if we got here.

src/SonarAnalyzer.CSharp/Rules/UseStringIsNullOrEmpty.cs
```diff
@@ -0,0 +1,122 @@
...
+        private static bool IsConstantEmptyString(ExpressionSyntax expression, SemanticModel semanticModel)
+        {
+            var literalExpression = expression as LiteralExpressionSyntax;
+
+            if (literalExpression?.Token.ValueText == string.Empty)
+            {
+                return true;
+            }
+
+            var constValue = semanticModel.GetConstantValue(expression);
```
tvsonar: `expression` might be `null` (on invalid code) and then this call would fail.

PR #1079, opened 2016-12-21
SLVS-1111 IComparable should be implemented together with other compa…

valhristov:
…rison methods

NOTE: There are no tests with single operator missing, because such code cannot be compiled anyway.


src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
```diff
@@ -0,0 +1,140 @@
...
+                                   .Any(IsSystemIComparable);
+        }
+
+        private static IList<string> GetMembersToOverride(ClassDeclarationSyntax classDeclaration, SemanticModel semanticModel)
+        {
+            var overridenOperators = classDeclaration.Members
+                                                     .OfType<OperatorDeclarationSyntax>()
+                                                     .Select(o => o.OperatorToken.ValueText);
+
+            var operatorsToOverride = RequiredOperators.Except(overridenOperators)
```
tvsonar: It's not a big deal, but `Except` is a generic extension method, which is most probably N^2 in complexity. This could be improved if `RequiredOperators` was an `ISet`.

src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        private static bool ImplementsComparable(ClassDeclarationSyntax classDeclaration, SemanticModel semanticModel)
+        {
+            if (classDeclaration.BaseList == null)
+            {
+                return false;
+            }
+
+            return classDeclaration.BaseList
```
tvsonar: You should get the symbol for the class declaration, and fully go through the semantic model. A class declaration can be one part of a partial class, so the BaseList might not contain all the base classes and interfaces.
Instead you should use `(semanticModel.GetDeclaredSymbol(classDeclaration) as INamedTypeSymbol).AllInterfaces` to access all implemented interfaces.

Also note that there's already a helper method for this: `TypeHelper.ImplementsAny(...)`

src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
```diff
@@ -0,0 +1,140 @@
...
+                    if (!ImplementsComparable(classDeclaration, c.SemanticModel))
+                    {
+                        return;
+                    }
+
+                    var membersToOverride = GetMembersToOverride(classDeclaration, c.SemanticModel);
+
+                    if (membersToOverride.Any())
+                    {
+                        c.ReportDiagnostic(Diagnostic.Create(Rule, classDeclaration.Identifier.GetLocation(), string.Join(", ", membersToOverride)));
```
tvsonar: with a very strange invalid input `class : IComparable` the Identifier might be `null`.

src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
```diff
@@ -0,0 +1,140 @@
...
+                                   .Types
+                                   .Select(t => t.Type)
+                                   .Select(t => semanticModel.GetTypeInfo(t))
+                                   .Any(IsSystemIComparable);
+        }
+
+        private static IList<string> GetMembersToOverride(ClassDeclarationSyntax classDeclaration, SemanticModel semanticModel)
+        {
+            var overridenOperators = classDeclaration.Members
+                                                     .OfType<OperatorDeclarationSyntax>()
```
tvsonar: You should use the semantic model, because in a partial class you'd only get the operators that are in the current syntax tree. You can get the members like this: `namedType.GetMembers().OfType<IMethodSymbol>().Where(m => m.MethodKind == MethodKinds.Operator)`, or something very similar.

src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        private static bool IsSystemIComparable(TypeInfo typeInfo)
+        {
+            if (typeInfo.Type == null)
+            {
+                return false;
+            }
+
+            return typeInfo.Type.ContainingNamespace?.Name == nameof(System) &&
```
tvsonar: I tend to prefer going through the `KnownType` abstraction. There are extension methods that help you with type checking, so it would look something like `typeInfo.Type.Is(KnownType.System_IComparable)`

src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            var operatorsToOverride = RequiredOperators.Except(overridenOperators)
+                                                       .ToList();
+
+            var overridesEquals = classDeclaration.Members
+                                                  .OfType<MethodDeclarationSyntax>()
+                                                  .Any(m => IsObjectEquals(m, semanticModel));
+            if (!overridesEquals)
+            {
+                operatorsToOverride.Insert(0, ObjectEquals);
```
tvsonar: Good idea to put the `.Equals` in the first place, but `.Insert` has a O(N) complexity on `List<T>`. You can avoid this by simply doing this check first and adding `.Equals` as the first item in the list. (Again, not a big deal considering that the list will be short, but why not)

src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                return false;
+            }
+
+            if (methodDeclaration.ParameterList?.Parameters.Count != 1)
+            {
+                return false;
+            }
+
+            var parameterType = semanticModel.GetTypeInfo(methodDeclaration.ParameterList.Parameters[0].Type).Type;
```
tvsonar: You could get the `IMethodSymbol` for the `methodDeclaration`, and then you have direct access to the type of the parameters and later to the return type.

src/SonarAnalyzer.Common/Helpers/KnownType.cs
```diff
@@ -177,6 +177,8 @@ internal sealed class KnownType
         public static readonly KnownType System_NullReferenceException = new KnownType("System.NullReferenceException");
         public static readonly KnownType System_OutOfMemoryException = new KnownType("System.OutOfMemoryException");
 
+        public static readonly KnownType System_IComparable = new KnownType($"{nameof(System)}.{nameof(System.IComparable)}");
```
tvsonar: Currently this is not used anywhere. By the way, these `nameof`s look a bit weird. (I see the value of them, because this way there's not gonna be any typos though)

src/Tests/SonarAnalyzer.UnitTest/TestCases/ComparableInterfaceImplementation.cs
```diff
@@ -0,0 +1,542 @@
...
+        {
+            return Compare(left, right) < 0;
+        }
+
+        public static bool operator >(MissingNotEqual left, MissingNotEqual right)
+        {
+            return Compare(left, right) > 0;
+        }
+    }
+}
```
tvsonar: What should happen in the following case?
```C#
class Base : IComparable // Compliant
{
 ... implement IComparable, and all the required operators
}
class Derived : Base, IComparable
{
  // empty
}
```
I think in this case `Derived` should also be compliant, because it already has all the required methods. What does FxCop do in this case?

its/expected/Ember-MM/Ember.Plugins-{9496C697-5AFD-4813-AEDC-AF33FACEADF0}-S1210.json
```diff
@@ -0,0 +1,17 @@
+{
+"issues":  [
+{
+"id":  "S1210",
+"message":  "When implementing IComparable or IComparable\u003cT\u003e, you should also override Equals, \u003c, \u003e, ==, !=.",
```
valhristov: We don't actually know... We test for both and if any is implemented, we check for Equals and the operators. I can refactor the code to return the implemented interfaces, though.

src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
valhristov: Required operators is a IList because ISet does not preserve the original order of its items... And the tests fail when the diagnostic messages do not match exactly.

src/Tests/SonarAnalyzer.UnitTest/TestCases/ComparableInterfaceImplementation.cs
valhristov: We complain, FxCop - doesn't :)

valhristov: And this is a problem, because classSymbol.GetMembers() returns methods declared on this class only, not base classes

valhristov: And btw, isn't the IComparable on Derived redundant? May be we should complain about that too?

src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
```diff
@@ -0,0 +1,142 @@
...
+                            string.Join(" or ", implementedComparableInterfaces),
+                            string.Join(", ", membersToOverride)));
+                    }
+                },
+                SyntaxKind.ClassDeclaration);
+        }
+
+        private static IEnumerable<string> GetImplementedComparableInterfaces(INamedTypeSymbol classSymbol)
+        {
+            return classSymbol.Interfaces
```
tvsonar: There's an `.AllInterfaces` property, which returns the interfaces that are not directly implemented but through base classes or through other interfaces.

src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
```diff
@@ -0,0 +1,142 @@
...
+                              .Where(i => i.OriginalDefinition.IsAny(ComparableInterfaces))
+                              .Select(GetClassNameOnly)
+                              .ToList();
+        }
+
+        private static string GetClassNameOnly(INamedTypeSymbol typeSymbol)
+        {
+            var fullName = typeSymbol.OriginalDefinition.ToString();
```
tvsonar: you can use `typeSymbol.OriginalDefinition.Name`. Or there's a `.ToDisplayParts()`, and then you'd have to get the last item from it. (In both cases generics might need to be special cased)

src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
```diff
@@ -0,0 +1,142 @@
...
+                        return;
+                    }
+
+                    var implementedComparableInterfaces = GetImplementedComparableInterfaces(classSymbol);
+                    if (!implementedComparableInterfaces.Any())
+                    {
+                        return;
+                    }
+
+                    var classMembers = classSymbol.GetMembers().OfType<IMethodSymbol>();
```
tvsonar: As you mentioned, `.GetMembers()` return only the directly contained methods. You can get all the methods by calling `.GetMembers()` on the `.BaseType`, and for the base type of that one, ...

There's already a helper method for it, so: `classSymbol.GetSelfAndBaseTypes().SelectMany(type => type.GetMembers())`. This will however always return `object.Equals` too. One way of copying with that problem is excluding `object` from the type chain:
`classSymbol.GetSelfAndBaseTypes().Where(type => !type.Is(KnownType.System_Object)).SelectMany(type => type.GetMembers())`

src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
valhristov: The problem here is that if I use typeSymbol.OriginalDefinition.Name, the result for IComparable<T> is IComparable (i.e. without T). Using ToDisplayParts() should work, but seems like overkill, just to remove the namespace part... What do you think?

src/Tests/SonarAnalyzer.UnitTest/TestCases/ComparableInterfaceImplementation.cs
tvsonar: S1939 should report on the redundant `IComparable`

src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
tvsonar: I see. `.ToString()` calls `.ToDisplayString()` internally. So let's use `.ToDisplayString()` which shows the intent that we want the name, and not just calling a base `.ToString()` method.

src/SonarAnalyzer.Common/Helpers/KnownType.cs
```diff
@@ -174,6 +174,9 @@ internal sealed class KnownType
         public static readonly KnownType System_NullReferenceException = new KnownType("System.NullReferenceException");
         public static readonly KnownType System_OutOfMemoryException = new KnownType("System.OutOfMemoryException");
 
+        public static readonly KnownType System_IComparable = new KnownType("System.IComparable");
+        public static readonly KnownType System_IComparable_T = new KnownType($"System.IComparable<T>");
```
tvsonar: the `$` is redundant here.

src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
```diff
@@ -0,0 +1,155 @@
...
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(DiagnosticId)]
+    public class ComparableInterfaceImplementation : SonarDiagnosticAnalyzer
+    {
+        internal const string DiagnosticId = "S1210";
+        internal const string MessageFormat = "When implementing {0}, you should also override {1}.";
```
tvsonar: This is a minor detail, but we don't override operators but overload them. So the message is slightly wrong, but I think we shouldn't care.

valhristov: True, but we suggest to override Equals() in addition to the operators, hence I decided to use override... Otherwise it would have been more difficult to construct semantically correct message.

src/SonarAnalyzer.CSharp/Rules/ComparableInterfaceImplementation.cs
```diff
@@ -0,0 +1,155 @@
...
+                    if (membersToOverride.Any())
+                    {
+                        c.ReportDiagnostic(Diagnostic.Create(
+                            Rule, 
+                            classDeclaration.Identifier.GetLocation(),
+                            string.Join(" or ", implementedComparableInterfaces),
+                            string.Join(", ", membersToOverride)));
+                    }
+                },
+                SyntaxKind.ClassDeclaration);
```
valhristov: Yes, I am fixing that too.

PR #1080, opened 2016-12-21
SLVS-1108: Rule S1163 - Exceptions should not be thrown in finally bl…

Evangelink:
…ocks


src/SonarAnalyzer.Common/Helpers/TypeHelper.cs
```diff
@@ -158,6 +158,19 @@ public static bool DerivesFromAny(this ITypeSymbol typeSymbol, ISet<KnownType> b
             return false;
         }
 
+        public static bool DerivesFromAny(this ITypeSymbol typeSymbol, IEnumerable<ITypeSymbol> baseTypes)
```
tvsonar: This method is never used.

src/SonarAnalyzer.CSharp/Rules/NoExceptionsInFinally.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        private class ThrowInFinallyWalker : CSharpSyntaxWalker
+        {
+            private readonly SyntaxNodeAnalysisContext context;
+
+            public ThrowInFinallyWalker(SyntaxNodeAnalysisContext context)
+            {
+                this.context = context;
+
+                if (!(context.Node is FinallyClauseSyntax))
```
tvsonar: There's a cast (`is`) in this line, and also one in line 48. You could have only a single cast before the `new ThrowInFinallyWalker()` call.

src/SonarAnalyzer.CSharp/Rules/NoExceptionsInFinally.cs
```diff
@@ -45,19 +45,23 @@ public class NoExceptionsInFinally : SonarDiagnosticAnalyzer
...
         {
             private readonly SyntaxNodeAnalysisContext context;
 
-            public ThrowInFinallyWalker(SyntaxNodeAnalysisContext context)
+            public ThrowInFinallyWalker(SyntaxNodeAnalysisContext context, FinallyClauseSyntax finallyClause)
             {
                 this.context = context;
 
-                if (!(context.Node is FinallyClauseSyntax))
+                if (finallyClause == null)
```
tvsonar: @Evangelink I would remove this check. That way the constructor is simpler, there's one less parameter. And I don't think it's a valid fear that you'll call this walker with anything else then a finally clause.

PR #1083, opened 2016-12-30
Visual Studio 2017 RC compatibility

sylvainvivenot:
Version range is changed to allow installation on Visual Studio 2017 RC


tvsonar:
Thanks @sylvainvivenot. Are you using directly the SonarAnalyzer extension? Why aren't you using SonarLint, which is in the VS marketplace? (We'll be releasing SonarLint for VS2017 shortly)

PR #1085, opened 2017-01-05
SLVS-1117: Rule S1451 - Track lack of copyright and license headers

Evangelink:


src/SonarAnalyzer.CSharp/Rules/CheckFileLicense.cs
```diff
@@ -0,0 +1,132 @@
...
+                    stac.ReportDiagnostic(Diagnostic.Create(Rule, Location.Create(stac.Tree, TextSpan.FromBounds(0, 0)), properties));
+                }
+            });
+        }
+
+        private static bool IsRegexPatternValid(string pattern)
+        {
+            try
+            {
+                new Regex(pattern);
```
SonarTech: CRITICAL Either remove this useless object instantiation of class "Regex" or use it. [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=csharpsquid%3AS1848)

SonarTech:
SonarQube analysis reported 1 issue
* CRITICAL 1 critical

Watch the comments in this conversation to review them.

src/SonarAnalyzer.CSharp/Rules/CheckFileLicense.cs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        private static string GetHeaderOrDefault(SyntaxTrivia trivia)
+        {
+            var isComment = trivia.IsKind(SyntaxKind.SingleLineCommentTrivia) || trivia.IsKind(SyntaxKind.MultiLineCommentTrivia);
+
+            return isComment ? trivia.ToString() : null;
+        }
+
+        private static bool AreHeadersEqual(string header, string expectedHeader, bool isRegex)
```
tvsonar: It's strange to pass the `isRegex` here, when that's coming from an instance field. Let's make the method non-static, and then we can use the field directly.

The same is true for the `expectedHeader`.

src/SonarAnalyzer.CSharp/Rules/CheckFileLicense.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                return Regex.IsMatch(header, expectedHeader, RegexOptions.Compiled);
+            }
+            else
+            {
+                return header == expectedHeader;
+            }
+        }
+
+        private ImmutableDictionary<string, string> CreateProperties()
```
tvsonar: Let's rename this to `CreateDiagnosticProperties` to be more explicit. (Making it a property is also an option)

src/SonarAnalyzer.CSharp/Rules/CheckFileLicenseCodeFixProvider.cs
```diff
@@ -0,0 +1,108 @@
...
+        public sealed override FixAllProvider GetFixAllProvider() => DocumentBasedFixAllProvider.Instance;
+
+        protected sealed override Task RegisterCodeFixesAsync(SyntaxNode root, CodeFixContext context)
+        {
+            var diagnostic = context.Diagnostics.First();
+            if (!diagnostic.Properties.Any() ||
+                !diagnostic.Properties.ContainsKey(CheckFileLicense.IsRegularExpressionPropertyKey) ||
+                !diagnostic.Properties.ContainsKey(CheckFileLicense.HeaderFormatPropertyKey))
+            {
+                return Task.Delay(0);
```
tvsonar: We discussed it in person, but to me this is not easy to read. I would `return;` here and make the method async. You could also return `Task.CompletedTask` which shows the intent a bit better.

src/SonarAnalyzer.CSharp/Rules/CheckFileLicenseCodeFixProvider.cs
```diff
@@ -0,0 +1,108 @@
...
+                    Title,
+                    c =>
+                    {
+                        var fixedLeadingTrivias = FixNodeLeadingTrivias(syntaxNode.GetLeadingTrivia(), diagnostic.Properties[CheckFileLicense.HeaderFormatPropertyKey]);
+                        var newRoot = root.ReplaceNode(syntaxNode, syntaxNode.WithLeadingTrivia(fixedLeadingTrivias));
+                        return Task.FromResult(context.Document.WithSyntaxRoot(newRoot));
+                    }),
+                context.Diagnostics);
+
+            return Task.Delay(0);
```
tvsonar: if the method is async, then you can drop this line.

src/SonarAnalyzer.CSharp/Rules/CheckFileLicenseCodeFixProvider.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        private IEnumerable<SyntaxTrivia> FixNodeLeadingTrivias(SyntaxTriviaList currentTriviaList, string comment)
```
tvsonar: this method can probably be `static`.

src/Tests/SonarAnalyzer.UnitTest/Common/RuleTest.cs
```diff
@@ -28,6 +28,7 @@
 using SonarAnalyzer.Common;
 using Microsoft.CodeAnalysis.CodeFixes;
 using SonarAnalyzer.Helpers;
+using SonarAnalyzer.Rules.CSharp;
```
tvsonar: This seems to be redundant.

src/SonarAnalyzer.ruleset
```diff
@@ -17,6 +17,7 @@
     <Rule Id="S131" Action="Warning" />
     <Rule Id="S1313" Action="Warning" />
     <Rule Id="S134" Action="Warning" />
+    <Rule Id="S1451" Action="Warning" />
```
tvsonar: This should also be added to the SonarAnalyzer.Test.ruleset.

src/SonarAnalyzer.CSharp/Rules/CheckFileLicenseCodeFixProvider.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Collections.Generic;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    public class CheckFileLicenseCodeFixProvider : SonarCodeFixProvider
+    {
+        internal const string Title = "Add or update license header";
+        public sealed override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray.Create(CheckFileLicense.DiagnosticId);
+        public sealed override FixAllProvider GetFixAllProvider() => DocumentBasedFixAllProvider.Instance;
```
tvsonar: You can use the `WellKnownFixAllProvider` here, maybe that's faster.

PR #1089, opened 2017-01-09
SLVS-1124 The symbolic value representing the result of nameof(xxx) s…

valhristov:
…hould always have a NotNull constraint

### DIFF
--- a/src/SonarAnalyzer.CSharp/Helpers/SyntaxHelper.cs
+++ b/src/SonarAnalyzer.CSharp/Helpers/SyntaxHelper.cs
@@ -30,6 +30,7 @@ namespace SonarAnalyzer.Helpers
         public static readonly ExpressionSyntax NullLiteralExpression = SyntaxFactory.LiteralExpression(SyntaxKind.NullLiteralExpression);
         public static readonly ExpressionSyntax FalseLiteralExpression = SyntaxFactory.LiteralExpression(SyntaxKind.FalseLiteralExpression);
         public static readonly ExpressionSyntax TrueLiteralExpression = SyntaxFactory.LiteralExpression(SyntaxKind.TrueLiteralExpression);
+        public static readonly string NameOfKeywordText = SyntaxFacts.GetText(SyntaxKind.NameOfKeyword);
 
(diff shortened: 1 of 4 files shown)


src/Tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.cs
```diff
@@ -7,6 +7,14 @@ namespace Tests.Diagnostics
 {
     public class ConditionEvaluatesToConstant
     {
+        void Nameof(string s)
+        {
+            if (null == nameof(Method1)) // Noncompliant
+//              ^^^^^^^^^^^^^^^^^^^^^^ {{Change this condition so that it does not always evaluate to "false".}}
```
tvsonar: The `{{...}}` should probably be on line 12.

src/SonarAnalyzer.CSharp/Helpers/SyntaxHelper.cs
```diff
@@ -120,13 +120,19 @@ public static bool IsInNameofCall(this ExpressionSyntax expression, SemanticMode
...
             if (calledSymbol != null)
             {
                 return false;
             }
 
-            var nameofIdentifier = (nameofCall?.Expression as IdentifierNameSyntax)?.Identifier;
+            var nameofIdentifier = (expression?.Expression as IdentifierNameSyntax)?.Identifier;
+
             return nameofIdentifier.HasValue &&
                 (nameofIdentifier.Value.ToString() == SyntaxFacts.GetText(SyntaxKind.NameOfKeyword));
```
tvsonar: Could you extract `SyntaxFacts.GetText(SyntaxKind.NameOfKeyword)` to a `static readonly` field? (Similarly to `SyntaxHelper.NullLiteralExpression`)

src/Tests/SonarAnalyzer.UnitTest/TestCases/ConditionEvaluatesToConstant.cs
valhristov: you are so right

PR #1095, opened 2017-01-23
SLVS-1139: Analyzer should learn from any const field

Evangelink:


src/SonarAnalyzer.CSharp/Rules/NullPointerDereference.cs
```diff
@@ -157,47 +157,49 @@ private ProgramState ProcessElementAccess(ProgramState programState, ElementAcce
...
 
                 var symbol = semanticModel.GetSymbolInfo(identifier).Symbol;
-                if (symbol == null)
-                {
-                    return programState;
-                }
-                if ((IsNullableValueType(symbol) && !IsGetTypeCall(memberAccess)) ||
+                var fieldSymbol = symbol as IFieldSymbol;
+
+                if (symbol == null ||
```
valhristov: This if has became very hard to understand; I think separating the OR conditions in different if statements would help (I am not sure, however, if SQ will complain about that).

src/SonarAnalyzer.CSharp/Rules/NullPointerDereference.cs
```diff
@@ -157,47 +157,49 @@ private ProgramState ProcessElementAccess(ProgramState programState, ElementAcce
                 return ProcessIdentifier(programState, identifier, symbol);
             }
 
-            private IdentifierNameSyntax GetIdentifierFromMemberAccess(MemberAccessExpressionSyntax memberAccess)
+            private static Tuple<IdentifierNameSyntax, bool> GetIdentifierFromMemberAccess(MemberAccessExpressionSyntax memberAccess)
```
tvsonar: Let's make this `Tuple` a dedicated type instead. `Item1` and `Item2` doesn't mean anything.

src/SonarAnalyzer.CSharp/Helpers/SymbolicExecution/ExplodedGraph.cs
```diff
@@ -526,13 +526,33 @@ private void VisitBinaryBranch(BinaryBranchBlock binaryBranchBlock, ExplodedGrap
...
+        #region VisitExpression
 
         private ProgramState VisitMemberAccess(MemberAccessExpressionSyntax memberAccess, ProgramState programState)
         {
             SymbolicValue memberExpression;
             var newProgramState = programState.PopValue(out memberExpression);
-            var sv = new MemberAccessSymbolicValue(memberExpression, memberAccess.Name.Identifier.ValueText);
+            SymbolicValue sv = null;
+            var identifier = memberAccess.Name as IdentifierNameSyntax;
+            if (identifier != null)
```
tvsonar: I don't like this deep `if` nesting here.

PR #1098, opened 2017-01-25
SLVS-1143 Do not throw exceptions for new C# constructs in CFG and SE

valhristov:
Seems to be not throwing exceptions with this class:
```
    public class CSharp7
    {
        public void OutVariables(string s)
        {
            if (int.TryParse(s, out int n))
            {
                Console.WriteLine(n);
            }
        }

        public void Patterns(object o)
        {
            if (o is null)
            {
                return;
            }

            if (o is int i)
            {
                Console.WriteLine(i);
            }

            switch (o)
            {
                case int n:
                    Console.Write(n);
                    break;
                case string s when s.Length > 0:
                    Console.WriteLine(s);
                    break;
                case null:
                    Console.WriteLine("null");
                    break;
                default:
                    break;
            }
        }

        public (int, int) ReturnTuples(string one, string two)
        {
            return (int.Parse(one), int.Parse(two));
        }

        public (int One, int Two) ReturnNamedTuples(string one, string two)
        {
            return (One: int.Parse(one), Two: int.Parse(two));
        }

        public void ConsumeTuples()
        {
            var t1 = ReturnTuples("one", "two");
            Console.WriteLine($"{t1.Item1}, {t1.Item2}");

            var t2 = ReturnNamedTuples("one", "two");
            Console.WriteLine($"{t2.One}, {t2.Two}");

            (var one, var two) = ReturnTuples("three", "four");
            Console.WriteLine($"{one}, {two}");
        }
    }
```


tvsonar:
And does it throw without your modification?

valhristov:
Yes, it throws.

PR #1099, opened 2017-01-25
VB.NET redundant ByVal keyword

Corniel:
The ByVal keyword for decorrating function and constructor parameters is redundant, and should not be used.


Corniel:
Any views on this?

tvsonar:
@Corniel Thanks for implementing this rule. We're working on the additional ceremony around the implementation of rules. Namely, we're creating the rule specification in our rule database, adding a JIRA ticket for the rule so that it shows up in the release notes... When all this is done, we'll have a closer look at the PR. And in the end it will be merged to master.

Please give us a bit of time, here's the first step: https://jira.sonarsource.com/browse/RSPEC-3860

cc @valhristov

Corniel:
You stated that ByVal is ignored for ParamArray args(). That is as far as I know not correct (aslo for ParamArray is redundant), and also not what I implemented. What the reason for this statement?

tvsonar:
@Corniel We say in the RSPEC that the rule is ignoring `ByVal` on `ParamArray`. This means that no issue will be reported on `ByVal ParamArray`. The reason for this can be found here: https://msdn.microsoft.com/en-us/library/538f81ec.aspx#Anchor_0. The second list item says that the parameter array must be passed by value and it's therefore a good practice to explicitly include the `ByVal` keyword.

Corniel:
I never, ever saw a ParamArray that explicitly was marked. Not as ByVal nor ByRef (what would be strange). Not in code of my colleagues, not in code examples on-line,

The whole Idea of redundancy is (imho) that you should not type/specify what is default behaviour. If you want to follow the MSDN, I can understand that, but I would not be mine choice.

valhristov:
I was searching yesterday why MSDN considers adding ByVal on ParamArray a good practice, but I could not find anything. I think we should keep the current implementation and if someone gives a good reason for excluding the ParamArray parameters we could revisit the rule.

Corniel:
I agree with @valhristov. :)

Corniel:
What is the status of this one @tvsonar / @valhristov ?

valhristov:
Hi @Corniel,

We moved the code to another repository a couple of months ago and unfortunately I can no longer merge your pull requests.

I created tickets for each of the new rules:
https://jira.sonarsource.com/browse/SONARVB-214
https://jira.sonarsource.com/browse/SONARVB-229
https://jira.sonarsource.com/browse/SONARVB-231

There are two options for how to proceed from now on:

- I can copy the changes to [Sonar C# (yes, the VB.NET code is still coupled with the C#)](https://github.com/SonarSource/sonar-csharp), but this way you will not be credited for the rules, e.g. the commit will be made with my name.
- You could copy the changes to the new repository, and you will receive the credit for the new rules. In addition, you will have to sign a document that transfers the rights to your code to SonarSource. I can send you the document if you choose this path.

We cannot commit to any timeframe for the next release of Sonar VB.NET, but a good guess would be around the end of the summer.

Corniel:
I'm fine with you copying it. It is not about getting credits, but about getting a better experience developing vb.NET. :)

valhristov:
Hi Corniel,

I created three tickets:
[https://github.com/SonarSource/sonar-csharp/issues/286]
[https://github.com/SonarSource/sonar-csharp/issues/285]
[https://github.com/SonarSource/sonar-csharp/issues/284]

I cannot say when we will start working on them, but we want these rules included in the next release of Sonar VB.NET in the end of the summer.

Regards,
Valeri

PR #1101, opened 2017-01-28
VB.NET. Use the If operator instead of the IIf function

Corniel:
Visual Basic .NET offers a non-short-cutting function (<code>IIf()</code>) that returns one of the two objects based on the outcome of an expression. This undesirable, as it is slower (because of the non-short-cutting), and potentially leading to runtime exceptions because it always evaluate always evaluates all three of its arguments.


Corniel:
Can somebody tell me what the status is?

PR #1102, opened 2017-01-28
VB.NET: Use CType for casting or explicit convertions

Corniel:
As promised, a recommit of this changes.

Visual Basic .NET offers (for backward compatibility) set of functions that convert from object to different primitive types. However, the naming is misleading, as it suggest a cast. It is better to choose explicitly cast using CType() or use Convert.To().


Corniel:
Remark, one of the cases is not detected. At least, that is the claim, but the debugger hits the point and adds the issue.

valhristov:
Hi Corniel, can you cite a resource which states that the CBool, CByte, etc. functions should not be used?

I am not a VB expert, but as far as I can tell (by exploring the compiled code with ILSpy) C<type> functions do basically the same as the CType. Also, [the Remarks section in this MSDN article](https://msdn.microsoft.com/en-us/library/s2dy91zy.aspx#Anchor_3) states that they should be the preferred choice when converting types.

I agree that they do more than I would expect and I would prefer to use DirectCast/TryCast or Convert.To<type> functions, instead, because they clearly demonstrate my intents.

Corniel:
See:
* http://stackoverflow.com/questions/2703585/casting-datatypes-with-directcast-ctype-trycast
* http://stackoverflow.com/questions/3056514/difference-between-directcast-and-ctype-in-vb-net

The CType, and CBool, ect, exist because VisualBasic was loosely typed. In that context they had meaning. You still can argue that there implementations are limited (The CStr(Guid.NewGuid() will fail), or wrong (it uses rounding when converting double to int). However, in strongly typed code, you should just use System.Convert for casting (ore XmlConvert or JsonConvert when that is more suitable), and use DirectCast (TryCast is only supports reference types, so that will only work for casting to string).

I was also wrong in that sense that I forgot to check if CType did the same as DirectCast. So that should be change.

PR #1108, opened 2017-02-10
SONARCS-666 Improve issue highlighting in S1066

valhristov:


src/SonarAnalyzer.CSharp/Rules/IfCollapsible.cs
```diff
@@ -33,6 +35,7 @@ public class IfCollapsible : SonarDiagnosticAnalyzer
     {
         internal const string DiagnosticId = "S1066";
         internal const string MessageFormat = "Merge this if statement with the enclosing one.";
+        internal const string AdditionalMessage = "Merge this if statement with the inner one.";
```
Evangelink: I am fine with this message but maybe using "enclosed" for the mirroring naming would be better.

PR #1127, opened 2017-02-23
SONARCS-817 Add secondary locations for Cyclomatic complexity metrics

valhristov:


src/SonarAnalyzer.CSharp/Helpers/CyclomaticComplexityWalker.cs
```diff
@@ -0,0 +1,145 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarAnalyzer.Common;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace SonarAnalyzer
+{
+    public class CyclomaticComplexityWalker : CSharpSyntaxWalker
+    {
+        public IList<SecondaryLocation> IncrementLocations { get; } = new List<SecondaryLocation>();
```
Evangelink: You can leave it as is but it gets back to what we discussed and I don't think `public` with a mutable collection is nice.

src/SonarAnalyzer.CSharp/Helpers/CyclomaticComplexityWalker.cs
```diff
@@ -0,0 +1,145 @@
...
+            base.VisitBinaryExpression(node);
+        }
+
+        public override void VisitCaseSwitchLabel(CaseSwitchLabelSyntax node)
+        {
+            CollectLocation(node.GetLocation());
+            base.VisitCaseSwitchLabel(node);
+        }
+
+        private void CollectLocation(Location location)
```
Evangelink: I would take the node and not the location here as every calling method is always `node.GetLocation()`.

Evangelink: I would also take (maybe optional) the call to the base method to reduce every calling methods.

src/SonarAnalyzer.CSharp/Rules/FunctionComplexity.cs
```diff
@@ -41,37 +52,64 @@ public class FunctionComplexity : FunctionComplexityBase
...
+                c => CheckComplexity<AccessorDeclarationSyntax>(c, "accessor", a => a.Keyword.GetLocation()),
                 SyntaxKind.GetAccessorDeclaration,
                 SyntaxKind.SetAccessorDeclaration,
                 SyntaxKind.AddAccessorDeclaration,
                 SyntaxKind.RemoveAccessorDeclaration);
         }
 
-        protected override int GetComplexity(SyntaxNode node) =>
-            new Metrics(node.SyntaxTree).GetComplexity(node);
+        protected void CheckComplexity<TSyntax>(SyntaxNodeAnalysisContext context, string declarationType, Func<TSyntax, Location> getLocation)
```
Evangelink: May I suggest to reorder the string because we haven't made any real change on the callers.

src/SonarAnalyzer.CSharp/SonarAnalyzer.CSharp.csproj
```diff
@@ -88,6 +88,7 @@
     </Compile>
     <Compile Include="Helpers\ControlFlowGraph\Blocks\ForeachCollectionProducerBlock.cs" />
     <Compile Include="Helpers\ControlFlowGraph\Blocks\ForInitializerBlock.cs" />
+    <Compile Include="Helpers\CyclomaticComplexityWalker.cs" />
```
Evangelink: I would put it into `Common` and not `Helpers`

src/SonarAnalyzer.Common/SonarAnalyzer.Common.csproj
```diff
@@ -100,7 +100,23 @@
     <Compile Include="Helpers\SymbolicExecution\SymbolicValues\Unary\*.cs" />
     <Compile Include="Helpers\SymbolicExecution\SymbolicValues\*.cs" />
     <Compile Include="Helpers\SymbolicExecution\*.cs" />
-    <Compile Include="Helpers\*.cs" />
```
Evangelink: Let's undo this to get back to `*.cs`

src/Tests/SonarAnalyzer.UnitTest/TestCases/FunctionComplexity.cs
```diff
@@ -4,19 +4,20 @@ namespace Tests.Diagnostics
...
     {
-        public FunctionComplexity() // Noncompliant {{The Cyclomatic Complexity of this constructor is 4 which is greater than 3 authorized.}}
-//             ^^^^^^^^^^^^^^^^^^
+        public FunctionComplexity() // Noncompliant [0] {{The Cyclomatic Complexity of this constructor is 4 which is greater than 3 authorized.}}
+                                    // Secondary@-1 [0] {{+1}}
         {
-            if (false) { }
-            if (false) { }
-            if (false) { }
+            if (false) { } // Secondary [0] {{+1}}
```
Evangelink: Do we have some test confirming the location of the secondary messages? If not I think we should.

src/SonarAnalyzer.CSharp/SonarAnalyzer.CSharp.csproj
valhristov: There is no Common in this project :)

Evangelink: My bad xD

PR #1135, opened 2017-02-28
Improve tests for should not be called

Evangelink:


Evangelink:
@valhristov No ticket for this one, could you review it? Don't look at the SONARCS-821 commit I just needed the code that was done in it.

PR #1137, opened 2017-03-01
SONARCS-838: All rules should have a consistent message structure

Evangelink:


Evangelink:
You can review, I will just try to fix this encoding of the `'` so it is easier for us to read.

PR #1138, opened 2017-03-01
SONARCS-831: Do not overload 'operator ==' on reference types

Evangelink:


src/SonarAnalyzer.CSharp/Rules/DotNotOverloadOperatorEqual.cs
```diff
@@ -0,0 +1,97 @@
...
+            var classDeclaration = declaration.Parent as ClassDeclarationSyntax;
+            if (classDeclaration == null)
+            {
+                return;
+            }
+
+            var hasAdditionOrSubstractionOverload =
+                classDeclaration.ChildNodes()
+                                .OfType<OperatorDeclarationSyntax>()
+                                .Any(op => op.OperatorToken.IsKind(SyntaxKind.PlusToken) ||
```
valhristov: Don't we have something like op.IsKind(SyntaxKind.PlusOperator), e.g. on the operator, not the token? The operator check seems more robust than the token.

src/Tests/SonarAnalyzer.UnitTest/TestCases/DotNotOverloadOperatorEqual.cs
```diff
@@ -0,0 +1,86 @@
+using System;
+using System.Collections.Generic;
+
+namespace Tests.Diagnostics
+{
+    class MyType
+    {
+        public static MyType operator ==(MyType x, MyType y) // Noncompliant
```
valhristov: I cannot find tests for the diagnostic message.

src/SonarAnalyzer.CSharp/Rules/DotNotOverloadOperatorEqual.cs
```diff
@@ -0,0 +1,97 @@
...
+        private void CheckForIssue(SyntaxNodeAnalysisContext analysisContext)
+        {
+            var declaration = (OperatorDeclarationSyntax)analysisContext.Node;
+
+            if (!declaration.OperatorToken.IsKind(SyntaxKind.EqualsEqualsToken))
+            {
+                return;
+            }
+
+            var structDeclaration = declaration.Parent as StructDeclarationSyntax;
```
valhristov: this is minor, but you could replace this with:
if (declaration.Parent is StructDeclarationSyntax)
{
    return;
}

src/SonarAnalyzer.CSharp/RspecStrings.resx
```diff
@@ -1,17 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!--
```
valhristov: Why is the header changed?

Evangelink: No idea I just updated the rule as the description changed.

src/SonarAnalyzer.CSharp/Rules/DotNotOverloadOperatorEqual.cs
Evangelink: I haven't found anything better.

src/Tests/SonarAnalyzer.UnitTest/TestCases/DotNotOverloadOperatorEqual.cs
Evangelink: Done

PR #1145, opened 2017-03-07
SONARCS-816 False positive in S2583 when analyzing a Singleton patter…

valhristov:
…n implementation with a lock


src/SonarAnalyzer.CSharp/Helpers/ControlFlowGraph/Blocks/LockBlock.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+                throw new ArgumentNullException(nameof(lockNode));
+            }
+
+            LockNode = lockNode;
+        }
+
+        public LockStatementSyntax LockNode { get; }
+
+        internal override Block GetPossibleNonEmptySuccessorBlock()
```
Evangelink: If you don't expect potential sub-class to change this I would seal the override.

src/SonarAnalyzer.CSharp/Helpers/ControlFlowGraph/Blocks/LockBlock.cs
```diff
@@ -0,0 +1,48 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using SonarAnalyzer.Helpers.FlowAnalysis.Common;
+using System;
+
+namespace SonarAnalyzer.Helpers.FlowAnalysis.CSharp
+{
+    public class LockBlock : SimpleBlock
```
Evangelink: Is it worth sealing this class or do you expect some sub-class to come?

valhristov: This is just a blatant copy-paste of the JumpBlock class. It was not sealed, I did not check the other block classes, but I would guess it is consistent...

PR #1146, opened 2017-03-07
SONARCS-839: No unused parameters should be spotted on empty, unsuppo…

Evangelink:
…rted or 'Main(string[] args)' methods


src/SonarAnalyzer.CSharp/Rules/MethodParameterUnused.cs
```diff
@@ -57,9 +57,18 @@ protected override void Initialize(SonarAnalysisContext context)
...
+                    {
+                        return;
+                    }
+
                     var symbol = c.SemanticModel.GetDeclaredSymbol(declaration);
                     if (symbol == null ||
-                        !symbol.ContainingType.IsClassOrStruct())
+                        !symbol.ContainingType.IsClassOrStruct() ||
+                        symbol.IsMainMethod() ||
+                        IsBodyOnlyThrowNotImplementedException(declaration, c.SemanticModel))
```
valhristov: Do we handle methods with a single comment inside? Should we?
```
public void Foo(int arg1)
{
  // Empty on purpose
}
```

Evangelink: This is handled and considered as an empty method so we don't report on it.

PR #1147, opened 2017-03-07
SONARCS-834 Exception types should be "public"

valhristov:


valhristov:
TODO: We should merge this with the original commit

PR #1149, opened 2017-03-08
SONARCS-815: Mutable fields should not be 'public static'

Evangelink:


valhristov:
I would put Exceptions section [here](https://jira.sonarsource.com/browse/RSPEC-3888)

PR #1150, opened 2017-03-08
SONARCS-832 'ref' and 'out' parameters should not be used.

valhristov:


src/SonarAnalyzer.CSharp/Rules/DoNotUseOutRefParameters.cs
```diff
@@ -0,0 +1,95 @@
...
+                c =>
+                {
+                    var parameter = (ParameterSyntax)c.Node;
+
+                    if (!parameter.Modifiers.Any(IsRefOrOut))
+                    {
+                        return;
+                    }
+
+                    var modifier = parameter.Modifiers.First(IsRefOrOut);
```
Evangelink: I would create some helper and merge this statement with the previous `if`.

src/SonarAnalyzer.CSharp/Rules/DoNotUseOutRefParameters.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            return method.Name.StartsWith("Try", StringComparison.Ordinal) &&
+                method.ReturnType.Is(KnownType.System_Boolean) && 
+                modifier.IsKind(SyntaxKind.OutKeyword);
+        }
+
+        private IMethodSymbol GetContainingMethod(ParameterSyntax parameter, SemanticModel semanticModel)
+        {
+            var method = parameter.GetParents().OfType<MethodDeclarationSyntax>().FirstOrDefault();
+            if (method != null)
```
Evangelink: Merge all the following using the ternary operator.

src/SonarAnalyzer.CSharp/Rules/DoNotUseOutRefParameters.cs
```diff
@@ -0,0 +1,95 @@
...
+        private bool IsTryPattern(IMethodSymbol method, SyntaxToken modifier)
+        {
+            return method.Name.StartsWith("Try", StringComparison.Ordinal) &&
+                method.ReturnType.Is(KnownType.System_Boolean) && 
+                modifier.IsKind(SyntaxKind.OutKeyword);
+        }
+
+        private IMethodSymbol GetContainingMethod(ParameterSyntax parameter, SemanticModel semanticModel)
+        {
+            var method = parameter.GetParents().OfType<MethodDeclarationSyntax>().FirstOrDefault();
```
Evangelink: Could you make a test with dumb parenthesis around some of the parameters? I am not sure your method is going to handle the case. Also I don't think you need the method to retrieve the declaring method (have a look at LiveVariableAnalysis.cs on line 188), Tamas was retrieving the parent without your helper method.

src/SonarAnalyzer.CSharp/Rules/DoNotUseOutRefParameters.cs
valhristov: The code is simple and readable, adding a helper to return Nullable<SyntaxToken> will not improve the situation much and I don't particularly like testing for default(SyntaxToken) too...

PR #1156, opened 2017-03-13
SONARCS-850: Rule S110: Inheritance tree of classes should not be too…

Evangelink:
… deep


src/SonarAnalyzer.CSharp/Rules/AvoidExcessiveInheritance.cs
```diff
@@ -0,0 +1,103 @@
...
+            defaultValue: MaximumDepthDefaultValue)]
+        public int MaximumDepth { get; set; } = MaximumDepthDefaultValue;
+
+        private const string FilteredClassesDefaultValue = "";
+        [RuleParameter(
+            key: "filteredClasses",
+            type: PropertyType.String,
+            description: "Classes to be filtered out of the count of inheritance (comma separated).",
+            defaultValue: FilteredClassesDefaultValue)]
+        public string FilteredClasses { get; set; } = FilteredClassesDefaultValue;
```
valhristov: I would recommend renaming to IgnoredClassNames or something... FilteredClasses does not sound nice to me (personal opinion, you don't have to adhere).

PR #1158, opened 2017-03-13
SONARCS-856 Rule S3904: Assemblies should have version information

valhristov:


src/SonarAnalyzer.CSharp/Rules/RequireAssemblyVersionAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(DiagnosticId)]
+    public class RequireAssemblyVersionAttribute : RequireAssemblyAttributeBase
+    {
+        internal const string DiagnosticId = "S3904";
+        internal const string MessageFormat = "Provide an AssemblyVersion attribute for this assembly.";
```
Evangelink: I would add single quotes around AssemblyVersion.

src/SonarAnalyzer.CSharp/Rules/RequireAssemblyVersionAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+using SonarAnalyzer.Common;
+using SonarAnalyzer.Helpers;
+
+namespace SonarAnalyzer.Rules.CSharp
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [Rule(DiagnosticId)]
+    public class RequireAssemblyVersionAttribute : RequireAssemblyAttributeBase
+    {
+        internal const string DiagnosticId = "S3904";
```
Evangelink: You can make it private as we don't use it anywhere else.

src/SonarAnalyzer.CSharp/Rules/RequireAssemblyVersionAttribute.cs
Evangelink: You can make it private as we don't use it anywhere else.

src/SonarAnalyzer.CSharp/Rules/RequireAssemblyVersionAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+
+        protected sealed override DiagnosticDescriptor Rule => rule;
+
+        protected override bool IsRequiredAttribute(AttributeSyntax attribute, SemanticModel semanticModel)
+        {
+            var attributeConstructor = semanticModel.GetSymbolInfo(attribute).Symbol as IMethodSymbol;
+
+            return attributeConstructor != null &&
+                attributeConstructor.ContainingType.Is(KnownType.System_Reflection_AssemblyVersionAttribute);
```
Evangelink: The extension method `Is` handles null so you could simplify with `attributeConstructor?.ContainingType.Is`

PR #1160, opened 2017-03-13
SONARCS-843: Rule S1206: 'Equals(object)' and 'GetHashCode()' should …

Evangelink:
…be overridden in pairs


src/SonarAnalyzer.CSharp/Rules/OverrideGetHashCodeOnOverridingEquals.cs
```diff
@@ -0,0 +1,83 @@
...
+            DiagnosticDescriptorBuilder.GetDescriptor(DiagnosticId, MessageFormat, RspecStrings.ResourceManager);
+        protected sealed override DiagnosticDescriptor Rule => rule;
+
+        protected override void Initialize(SonarAnalysisContext context)
+        {
+            context.RegisterSyntaxNodeActionInNonGenerated(
+                c =>
+                {
+                    var declaration = (TypeDeclarationSyntax)c.Node;
+                    if (declaration.Keyword.IsKind(SyntaxKind.InterfaceKeyword))
```
valhristov: Can you write a test which will fail this if? I mean, isn't registering on Class/StructDeclaration enough (e.g. if you don't register on InterfaceDeclaration, you will not get InterfaceDeclarationSyntax)?

PR #1164, opened 2017-04-11
InvalidCastToInterface: do not raise S1944 when casting to Nullable<T>

fsateler:


fsateler:
Moved to https://github.com/SonarSource/sonar-csharp/pull/219

PR #1166, opened 2017-06-08
Added test and attempt to fix edge case in S3897

overeemm:
We've found an edge case in rule S3897 when using inheritance. My test case illustrates the edge case. My fix however is not complete, but a step towards the right direction. Hopefully you can review and finish the fix. The rule results in a lot of false warnings right now.


overeemm:
Found out that this repo is old.

